PR #8, opened 2022-03-22
sigstore: Create API stubs

### LINKED ISSUE #7: Create API stubs
Create basic package structure and empty modules so that we can begin collaborating.

tetsuo-cpp:
Closes #7


sigstore/rekor/_client.py
```diff
@@ -0,0 +1,7 @@
+"""
+Client implementation for interacting with Rekor.
+"""
+
+
+class RekorClient:
```
tetsuo-cpp, line 6: Rekor's API seems to be a lot more complicated. The Rekor client in Go is generated via Swagger [here](https://github.com/sigstore/rekor/tree/main/pkg/generated/client).

We might have to figure out the main things we want/need to support for the beta.

tetsuo-cpp:
@di @woodruffw
What do you think of this package structure? At the moment I have:

- `fulcio`: Client API for interacting with Fulcio
- `rekor`: Client API for interacting with Rekor
- `pysign`: API for signing Python packages (builds on top of `fulcio` and `rekor`). The code in `_cli.py` should be a thin wrapper that contains argument parsing and calls into this module.

I'm thinking of adding `oidc` for the auth stuff too. But we can leave that for later.

di:
> Hmm, what do you want the code signing CLI tool to be called? I don't think we can just call it sigstore as that will potentially clash with other things.

I think the importable, public API should probably be `from sigstore import sign, verify`

tetsuo-cpp:
> I think the importable, public API should probably be `from sigstore import sign, verify`

Sure, but what about the tool itself? It's ok to use it like `python -m sigstore sign <blah>`. But if we're going to install something in the user `PATH`, I think it has to be more specific than `sigstore`.

PR #11, opened 2022-03-23
Add a simple Rekor client

di:
Draft PR: this is a rough Rekor client I created while starting to implement the signing mechanism, figured I'd throw this up early so we can discuss how we'd like to write clients for Fulcio and Rekor.


di:
Made this from the wrong branch, see https://github.com/trailofbits/sigstore-python/pull/12 instead.

PR #12, opened 2022-03-23
Add a simple Rekor client

di:
Draft PR: this is a rough Rekor client I created while starting to implement the signing mechanism, figured I'd throw this up early so we can discuss how we'd like to write clients for Fulcio and Rekor.


sigstore/_internal/rekor/_client.py
```diff
@@ -2,8 +2,83 @@
...
+
+DEFAULT_REKOR_URL = "https://rekor.sigstore.dev/api/v1/"
+
+
+class Endpoint:
+    def __init__(self, url, session):
+        self.url = url
+        self.session = session
+
+    def get(self):
```
tetsuo-cpp, line 18: Maybe inherit from `abc.ABC` and mark these as `@abstractmethod`? I can push a commit if you're happy for me to do that.

sigstore/_internal/rekor/_client.py
```diff
@@ -2,8 +2,83 @@
...
 
-    pass
+    def __init__(self, url=DEFAULT_REKOR_URL):
+        self.url = url
+        self.session = requests.Session()
+        self.session.headers.update(
+            {"Content-Type": "application/json", "Accept": "application/json"}
+        )
+
+    @property
```
tetsuo-cpp, line 35: Nice idea with this `Endpoint` pattern. Fulcio has fewer endpoints but I think I'll use that pattern there too.

sigstore/_internal/rekor/_client.py
di: Please do!

tetsuo-cpp: I don't think I have write permissions to your repo, could you add me? I have some commits for this + type hints and a few runtime checks.

tetsuo-cpp:
By the way, during development have you been running this against the public Sigstore instances? I've been trying to figure out the best way to test my Fulcio client.

di:
> By the way, during development have you been running this against the public Sigstore instances? I've been trying to figure out the best way to test my Fulcio client.

I have been. I think that's the best we can do right now w/ regards to testing.

tetsuo-cpp:
@mbestavros

sigstore/_internal/rekor/_client.py
di: Added!

sigstore/_internal/rekor/_client.py
```diff
@@ -2,8 +2,90 @@
...
+
+
+class RekorIndex(Endpoint):
+    @property
+    def retrieve(self) -> Endpoint:
+        return RekorRetrieve(urljoin(self.url, "retrieve/"), session=self.session)
+
+
+class RekorRetrieve(Endpoint):
+    def post(self, sha256_hash: Optional[str] = None) -> requests.Response:
```
tetsuo-cpp, line 47: We should figure out what the return type of these should be.

It should either work like this where we return a `requests.Response` which the caller is responsible for checking the status and then extracting the contents. Or we should do the checking within these functions and return something more specific to the endpoint (like the one below that returns `Tuple[str, str]`).

sigstore/_internal/rekor/_client.py
```diff
@@ -2,8 +2,95 @@
...
+from abc import ABC
+from typing import Any, Optional, Tuple
+from urllib.parse import urljoin
+
+import requests  # type: ignore
+
+DEFAULT_REKOR_URL = "https://rekor.sigstore.dev/api/v1/"
+
+
+class Endpoint(ABC):
```
tetsuo-cpp, line 15: I removed the `get` and `post` methods that returned `NotImplementedError`. MyPy didn't seem to like them since the subclasses of `Endpoint` had the methods taking different args.

sigstore/_internal/rekor/_client.py
di: I like what you're doing in https://github.com/trailofbits/sigstore-python/pull/13/ where these return dataclasses, we can do that here as well.

sigstore/_internal/rekor/_client.py
```diff
@@ -2,8 +2,95 @@
...
+
+class RekorIndex(Endpoint):
+    @property
+    def retrieve(self) -> Endpoint:
+        return RekorRetrieve(urljoin(self.url, "retrieve/"), session=self.session)
+
+
+class RekorRetrieve(Endpoint):
+    def post(self, sha256_hash: Optional[str] = None) -> requests.Response:
+        data = {"hash": f"sha256:{sha256_hash}"}
```
mbestavros, line 48: The `sha256:` prefix shouldn't be necessary according to the OpenAPI spec; I opened an issue when I ran into problems without it [here](https://github.com/sigstore/rekor/issues/699). Since it won't work without the prefix as of now, this is fine. But just a note for the future.

mbestavros (commented):
This is great! Thanks for tagging me 😄 I think the code looks good, and is a great foundation to build from.

My only question is around the `RekorEntries` class - there are multiple ways to get an entry, but the `get` method assumes you have the entry's UUID, which isn't necessarily a guarantee. You can also retrieve an entry using the signing materials used to create it (e.g. the artifact, signature, and key). I've got a rough implementation of it [here](https://github.com/mbestavros/sigstore-python-playground/blob/main/sigstore.py#L64).

How do you think that'd best fit in? Maybe a separate `get` with different inputs accepted? Or maybe it makes more sense to frame it as "search" (though that's also separate functionality on its own too).

di:
> My only question is around the RekorEntries class - there are multiple ways to get an entry, but the get method assumes you have the entry's UUID, which isn't necessarily a guarantee. You can also retrieve an entry using the signing materials used to create it (e.g. the artifact, signature, and key). I've got a rough implementation of it here.

I think you're mixing up the `GET /api/v1/log/entries/{entryUUID}`, which only takes a UUID, and `POST /api/v1/log/entries/retrieve`, which can take a hash, public key, email, etc.

> How do you think that'd best fit in? Maybe a separate get with different inputs accepted? Or maybe it makes more sense to frame it as "search" (though that's also separate functionality on its own too).

I think we should stick as close to the existing API structure as possible, so this would just require adding optional arguments to the method.

sigstore/_internal/rekor/_client.py
di: Yeah, unfortunately I'm not sure how much we can trust the OpenAPI spec to be correct and/or exist. It was removed for Fulcio entirely (https://github.com/sigstore/fulcio/pull/262). It probably makes the most sense for us to hand-craft our clients using the spec as a reference, but check them for accuracy against the live API.

di:
@tetsuo-cpp I think this is everything I need for keyless signing. I'd like to get this merged as-is so I can move forward with that, but we can definitely iterate on it more in the future (and add additional endpoints/functionality that doesn't currently exist).

Makefile
```diff
@@ -44,9 +44,7 @@ lint:
 		black $(ALL_PY_SRCS) && \
 		isort $(ALL_PY_SRCS) && \
 		flake8 $(ALL_PY_SRCS) && \
-		mypy $(PY_MODULE) test/ && \
```
tetsuo-cpp, line 47: Can we please just comment this out? I understand we don't want to block this work but I think we should still strive to have the API documented.

sigstore/_internal/rekor/_client.py
tetsuo-cpp: Sure, I'll do this before merging.

tetsuo-cpp:
> @tetsuo-cpp I think this is everything I need for keyless signing. I'd like to get this merged as-is so I can move forward with that, but we can definitely iterate on it more in the future (and add additional endpoints/functionality that doesn't currently exist).

Good call. I'll just push a few commits to address the stuff I mentioned. Then let's pull the trigger on this change.

tetsuo-cpp:
@di When you have a moment, could you also push the branch where you're implementing the `sign` command? I'm about to do some work to implement the `verify` command so it'd be good to see what your work looks like.

Makefile
tetsuo-cpp: [b1eb376](https://github.com/trailofbits/sigstore-python/pull/12/commits/b1eb3760859b1d1a8fe4ffb87867eeb0d6734a91)

di: Oh, this was a mistake, I meant to just remove `interrogate` and `git diff`. Thanks for catching!

tetsuo-cpp (approved):
I've pushed f10ed2c. I was just referring to the `openapi.yaml` in the Rekor repo to figure out what the return type should be, but I'm not 100% sure. Please take a look before merging. Otherwise LGTM as a first pass.

Makefile
tetsuo-cpp: Yep, that's what I mean though. Shouldn't we still try to aim to have a docstring for everything?

I've temporarily commented it out and will make an issue to reenable interrogate and get it passing.

PR #13, opened 2022-03-24
Implement Fulcio client API

### LINKED ISSUE #1: Implement Fulcio client
In order to implement signing, we need to write a Python client to interact with Fulcio. We need to write this in a generic way so that other programs can use `pysign`'s Fulcio client for their own purposes.

tetsuo-cpp:
Closes #1


tetsuo-cpp:
CC: @mbestavros

tetsuo-cpp:
@di I've converted this client to use the Endpoint pattern that you're using in your Rekor client. I still need to look into running Fulcio locally to test this.

di (approved):

PR #15, opened 2022-04-06
Update Fulcio client

di:
This is everything I needed to change to get the certificate request working for signing, plus some slight refactoring.


sigstore/_internal/fulcio/_client.py
```diff
@@ -80,40 +87,61 @@ def root_cert(self) -> Endpoint:
...
+        pem_blocks = resp.text.split(CERT_START.decode())
+        if not pem_blocks or pem_blocks[0]:
             raise FulcioClientError(f"Unexpected number of PEM blocks in Fulcio response: {resp}")
-        pem_blocks = pem_blocks[1:]
-        cert_list: List[Certificate] = []
-        for pem_block in pem_blocks:
-            cert: Certificate = load_pem_x509_certificate(pem_block)
-            cert_list.append(cert)
-        return CertificateResponse(cert_list, sct)
+        cert_list: List[Certificate] = [
```
tetsuo-cpp, line 131: Can we keep the check for 2 certificates or can there be an arbitrary number? The Go client only expects two: https://github.com/sigstore/fulcio/blob/main/pkg/api/client.go#L120-L125

di: I don't think that's what's happening there, [`pem.Decode`](https://pkg.go.dev/encoding/pem#Decode) says:

> Decode will find the next PEM formatted block (certificate, private key etc) in the input. It returns that block and the remainder of the input.

Which means the chain block might have multiple certificates. We probably need a similar decoder here.

di: Sorry, cb69b8a.

tetsuo-cpp: That's true. I should rephrase that. It's grabbing the first two certificates and ignoring the rest, whereas ours is loading them exhaustively into `chain_pems`.

Not sure that's a problem, but it's doing something different. The Java client does what ours does though.

tetsuo-cpp: Looks good.

PR #21, opened 2022-04-06
Implement `sign` command

### LINKED ISSUE #9: Implement `sign` command

di:
Fixes #9.


sigstore/_cli.py
```diff
@@ -9,10 +9,21 @@ def main():
...
+@click.option("identity_token", "--identity-token", type=click.STRING)
+@click.argument("file_", metavar="FILE", type=click.File("r"), required=True)
+def _sign(file_, identity_token):
+    click.echo(sign(file_=file_, identity_token=identity_token, output=click.echo))
 
 
 @main.command("verify")
-def _verify():
-    click.echo(verify())
+@click.option("certificate_path", "--cert", type=click.Path())
```
tetsuo-cpp, line 19: Just confirming. Is the idea here that we can supply a certificate and signature to verify "offline". But if we don't supply both of these, we check against Rekor?

sigstore/_sign.py
```diff
@@ -0,0 +1,114 @@
...
+
+    output("Retrieving signed certificate...")
+    # fulcio = FulcioClient("http://localhost:5555")
+    fulcio = FulcioClient()
+
+    oidc_identity = Identity(identity_token)
+
+    # Build an X.509 Certificiate Signing Request - not currently supported
+    """
+    builder = (
```
tetsuo-cpp, line 48: Is this an API that is proposed for `cryptography`?

sigstore/_sign.py
```diff
@@ -0,0 +1,114 @@
...
+    output("Generating ephemeral keys...")
+    # Generate the ephemeral private key
+    # private_key = Ed25519PrivateKey.generate()
+    # public_key = private_key.public_key()
+    private_key = ec.generate_private_key(ec.SECP384R1())
+    public_key = private_key.public_key()
+
+    output("Retrieving signed certificate...")
+    # fulcio = FulcioClient("http://localhost:5555")
+    fulcio = FulcioClient()
```
tetsuo-cpp, line 42: I guess we should eventually pipe through some CLI options to control the Fulcio and Rekor instances.

tetsuo-cpp (commented):
I think this is looking good. Are you planning to add anything to this?

sigstore/_sign.py
di: This is supported for `cryptography` (https://cryptography.io/en/latest/x509/reference/#cryptography.x509.CertificateSigningRequestBuilder) but sending a CSR to Fulcio is what's currently unsupported: https://github.com/sigstore/fulcio/issues/503

sigstore/_sign.py
di: Yes, let's make sure to do these as close as possible to how `cosign` does it today.

sigstore/_cli.py
di: We discussed this offline!

di:
@tetsuo-cpp I'll let you decide if we should lower or disable the coverage bar here, or try and add some very early tests. I think the former is fine as long as we come back later and test what we've got.

sigstore/_sign.py
tetsuo-cpp: Sure. Let's not block this PR on it.

I've made #27 to track this. We can revisit once we have basic functionality working for signing/verifying since we'll be in a better place to add the CLI flags.

sigstore/_sign.py
```diff
@@ -0,0 +1,102 @@
...
+)
+from sigstore._internal.oidc import Identity
+from sigstore._internal.rekor import RekorClient
+
+
+def _no_output(*a, **kw):
+    pass
+
+
+def sign(file_, identity_token, output=_no_output):
```
tetsuo-cpp, line 19: I think we should figure out what the return type should be soon. I'm thinking something like a `SigningResult` which allows you to tell between success/failure and whether there is a reason.

We could also just not have a return and consider any non-throwing call to `sign` to be a success.

I'll go ahead and merge this anyway since we have too many pieces of work branched off this, but we should definitely revisit it.

PR #23, opened 2022-04-08
Support multiple issuers and alternative proofs

### LINKED ISSUE #17: Support alternative proof for GitHub Actions OIDC tokens
Fulcio has a separate proof flow for GitHub Actions ODIC tokens which don't have an `email` claim: https://github.com/sigstore/fulcio/blob/main/config/config.jsn

We should provide the signed `sub` claim as proof for GHA tokens instead of attempting to get the `email` claim from the token.

di:
This PR provides an `Identity` class which supports all the OIDC issuers Fulcio currently supports, as well as determine alternative proofs (e.g. using the `subject` claim instead of `email` for GitHub Actions)

Fixes #17.

### DIFF
--- a/sigstore/_internal/oidc.py
+++ b/sigstore/_internal/oidc.py
@@ -1,0 +1,48 @@
+import jwt
+
+# From https://github.com/sigstore/fulcio/blob/b2186c01da1ddf807bde3ea8c450226d8e001d88/pkg/config/config.go#L182-L201  # noqa
+OIDC_ISSUERS = {
+    "https://accounts.google.com": {
+        "proof_claim": "email",
+    },
+    "https://oauth2.sigstore.dev/auth": {
+        "proof_claim": "email",
+    },
+    "https://token.actions.githubusercontent.com": {
+        "proof_claim": "sub",
+    },
+}
+AUDIENCE = "sigstore"
+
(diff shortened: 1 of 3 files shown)


sigstore/_internal/oidc.py
```diff
@@ -0,0 +1,48 @@
+import jwt
+
+# From https://github.com/sigstore/fulcio/blob/b2186c01da1ddf807bde3ea8c450226d8e001d88/pkg/config/config.go#L182-L201  # noqa
+OIDC_ISSUERS = {
+    "https://accounts.google.com": {
+        "proof_claim": "email",
```
tetsuo-cpp, line 6: Do we need to nest this dict in here? Can't it just map directly from the issuer to the proof?

di: I was imagining we might need more per-issuer details eventually, but I agree that it's not necessary right now.

tetsuo-cpp: Makes sense. I think we can cross that bridge when we come to it. If we need more issuer details, it might make more sense to have an `OIDCIssuer` abstract base class or something like that.

tetsuo-cpp (approved):
LGTM

PR #24, opened 2022-04-10
Implement SCT verification

### LINKED ISSUE #22: Implement SCT verification
This doesn't seem to exist in pyca/cryptography. We should check whether there's another library that includes this functionality. If not, we might need to hand-roll it ourselves.

tetsuo-cpp:
Closes #22


tetsuo-cpp:
This PR is based off the branch implementing the sign command. Unfortunately I can't point this PR at the sign branch and viewing the changes in isolation because the sign branch belongs to a fork.

At the moment, I can't seem to get the verification to work and am trying to figure out why (it's not very debuggable since it just throws a signature mismatch). I assume I'm somehow misconstructing the digitally signed part.

I've actually tried hacking https://github.com/pierky/sct-verify with the SCT, certificate, etc but I can't even get that to verify correctly. So whatever mistake I'm making, I'm reproducing it there too.

I'll continue investigating this...

haydentherapper:
Have you taken a look at https://pypi.org/project/ctutlz/ or https://github.com/theno/ctutlz?

Digging through the code a bit, this should provide the utilities that you need to verify SCTs. See for example https://github.com/theno/ctutlz/blob/db8efcf785efb4dcc7324ddba629d49e743890be/ctutlz/scripts/verify_scts.py#L94-L109, which handles verification of embedded SCTs.

Something to note: Fulcio is currently returning SCTs detached, not embedded in the certificate. I'm actively working on changing Fulcio to generate certificates with embedded SCTs. If you want to test against the live instance of Fulcio, you might have to hack around `ctutlz` a bit.

haydentherapper:
If you're seeing signature validation errors, it's likely that the bits that you're verifying are not the same bits that the CT log signed. For example, you need to consider that CT entry signature is generated over the TBS (to-be-signed) portion of the certificate (no cert signature), and it excludes the CT log extension. Additionally when reconstructing the merkle leaf to verify the signature, you will need to include the hash of the public key of the issuing certificate.

`ctutlz` handles these currently.

haydentherapper:
If you're seeing signature validation errors, it's likely that the bits that you're verifying are not the same bits that the CT log signed. For example, you need to consider that CT entry signature is generated over the TBS (to-be-signed) portion of the certificate (no cert signature), and it excludes the CT log extension. Additionally when reconstructing the merkle leaf to verify the signature, you will need to include the hash of the public key of the issuing certificate.

`ctutlz` handles these currently.

haydentherapper:
> Regarding the point about "excluding the CT log extension", does that mean that the extension length should not go at the end of the data to be signed? The structure looks like [this](https://datatracker.ietf.org/doc/html/rfc6962#section-3.2), but ctutlz and sct-verify both just have the extension length at the end (which is always 0).

There's two ways to verify an SCT. One is by verifying a detached SCT, where the SCT is presented alongside the certificate - The signature input is generated [here](https://github.com/theno/ctutlz/blob/db8efcf785efb4dcc7324ddba629d49e743890be/ctutlz/sct/signature_input.py#L7), what you linked above.

The other way is by verifying an SCT that's embedded in the certificate, meaning you don't have to store the SCT alongside the certificate since the SCT is embedded in an X.509 extension. That's constructed [here](https://github.com/theno/ctutlz/blob/db8efcf785efb4dcc7324ddba629d49e743890be/ctutlz/sct/signature_input.py#L38). When I mentioned "excluding the CT log extension", this is referring to the construction of the signature input for embedded SCT verification. The lengths are still included. There are certain extensions ([here](https://github.com/theno/ctutlz/blob/db8efcf785efb4dcc7324ddba629d49e743890be/ctutlz/sct/ee_cert.py#L188)] that are excluded when building the input.

Looking again at the first case, a detached SCT, I'm a bit confused why the ctutlz library doesn't use the TBS portion of the signature [here](https://github.com/theno/ctutlz/blob/db8efcf785efb4dcc7324ddba629d49e743890be/ctutlz/sct/signature_input.py#L31) when constructing the input. The DER encoding is not the same as the TBS portion. This could be the issue?

tetsuo-cpp:
Thanks for explaining.

> Note that Fulcio currently only supports detached SCTs, but in the coming weeks, will support both. I'd recommend preferring the latter, embedded SCTs, because it allows you to verify the SCT any time you use the certificate.

I've talked to @di about this recently and our plan at the moment is to begin by supporting detached SCTs and then move to supporting embedded SCTs later. The reason is that we're intending to demo this functionality in the coming weeks.

> Looking again at the first case, a detached SCT, I'm a bit confused why the ctutlz library doesn't use the TBS portion of the signature [here](https://github.com/theno/ctutlz/blob/db8efcf785efb4dcc7324ddba629d49e743890be/ctutlz/sct/signature_input.py#L31) when constructing the input. The DER encoding is not the same as the TBS portion. This could be the issue?

From looking at the RFC, it only specifies to use the TBS portion for the PreCert case. It isn't clear to me whether that's how it works for the ASN.1 case but I just assumed because `ctutlz` and `sct-verify` do it like this and it also seems to be how `cosign` does it according to [this snippet](https://github.com/google/certificate-transparency-go/blob/master/serialization.go#L119-L122).

sigstore/_internal/sct.py
```diff
@@ -70,7 +70,7 @@ def verify_sct(
     digitally_signed = _pack_digitally_signed(sct, cert)
     try:
         ctfe_key.verify(
-            signature=sct.signature.encode(),
+            signature=sct.signature,
```
di, line 73: Here's what I get when I run this with this change:

```
  File "/home/di/.pyenv/versions/3.11.0a6/bin/sigstore", line 33, in <module>
    sys.exit(load_entry_point('sigstore', 'console_scripts', 'sigstore')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/di/.pyenv/versions/3.11.0a6/lib/python3.11/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/di/.pyenv/versions/3.11.0a6/lib/python3.11/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/di/.pyenv/versions/3.11.0a6/lib/python3.11/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/di/.pyenv/versions/3.11.0a6/lib/python3.11/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/di/.pyenv/versions/3.11.0a6/lib/python3.11/contextlib.py", line 155, in __exit__
    self.gen.throw(typ, value, traceback)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/di/.pyenv/versions/3.11.0a6/lib/python3.11/site-packages/click/core.py", line 102, in augment_usage_errors
    yield
    ^^^^^
  File "/home/di/.pyenv/versions/3.11.0a6/lib/python3.11/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/di/git/trailofbits/sigstore-python/sigstore/_cli.py", line 26, in _sign
    sign(
    ^^^^^
  File "/home/di/git/trailofbits/sigstore-python/sigstore/_sign.py", line 73, in sign
    verify_sct(sct, cert, ctfe_key)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/di/git/trailofbits/sigstore-python/sigstore/_internal/sct.py", line 72, in verify_sct
    ctfe_key.verify(
    ^^^^^^^^^^^^^^^^
  File "/home/di/.pyenv/versions/3.11.0a6/lib/python3.11/site-packages/cryptography/hazmat/backends/openssl/ec.py", line 367, in verify
    _ecdsa_sig_verify(self._backend, self, signature, data)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/di/.pyenv/versions/3.11.0a6/lib/python3.11/site-packages/cryptography/hazmat/backends/openssl/ec.py", line 102, in _ecdsa_sig_verify
    res = backend._lib.ECDSA_verify(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: initializer for ctype 'unsigned char *' must be a bytes or list or tuple, not str
```

Why are you getting bytes for `sct.signature`?

di: Additionally, `sct.signature` is Base64 encoded here.

woodruffw: @tetsuo-cpp can confirm, but I think he was doing that to get the raw bytes for the signature. That wasn't 100% correct though, since the `signature` attribute was still Base64 encoded. [b79bc1b](https://github.com/trailofbits/sigstore-python/pull/24/commits/b79bc1bd0cdaa6d22e52b14864d29965eb7ba599) should get us a little closer, but I'm still debugging.

woodruffw: I might be missing something very obvious, but the `signature` field in the SCT doesn't appear to be a valid DER-encoded DSS signature.

So far I've base64 decoded the SCT, JSON-decoded that, then grabbed the `signature` and base64-decoded *that*, leaving me with a 74-byte long binary string. When I try to throw that string into `openssl asn1parse` it complains about the header being too long.

woodruffw: (ECDSA + SHA256 should be 72 bytes, not 74, but removing the first two bytes also doesn't produce a valid DER encoding)

tetsuo-cpp: Yeah sorry, I didn't push up that change but the diffs I posted in Slack were base 64 decoding the signature. So I have been testing with that change.

tetsuo-cpp:
It looks like there are a few flags preceding the signature that are added [here](https://github.com/google/certificate-transparency-go/blob/master/trillian/ctfe/serialize.go#L106-L125). The code comment refers an RFC describing TLS encoding [here](https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.1.4.1).

If I inspect an SCT that I recorded like this, I can see the flag values:
```
(env) tetsuo@Alexs-MacBook-Pro sigstore-python % python
Python 3.9.7 (default, Oct 12 2021, 22:38:23)
[Clang 13.0.0 (clang-1300.0.29.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import base64
>>> signature = base64.b64decode("BAMARzBFAiAUekMyQRyeS83yoxcdsHcv+u3qQPDADXiyYAZSlJrM7QIhAI5ZM1GdkVTH5gXyj6IM2JEP+sMc1/o2srrTxCS7xdKl")
>>> signature
b'\x04\x03\x00G0E\x02 \x14zC2A\x1c\x9eK\xcd\xf2\xa3\x17\x1d\xb0w/\xfa\xed\xea@\xf0\xc0\rx\xb2`\x06R\x94\x9a\xcc\xed\x02!\x00\x8eY3Q\x9d\x91T\xc7\xe6\x05\xf2\x8f\xa2\x0c\xd8\x91\x0f\xfa\xc3\x1c\xd7\xfa6\xb2\xba\xd3\xc4$\xbb\xc5\xd2\xa5'
>>> signature[0]
4 // NOTE(alex): This is the hash algorithm enum. 4 maps to sha256.
>>> signature[1]
3 // NOTE(alex): This is the signature algorithm enum. 3 maps to ecdsa.
>>> signature[4:]
b'0E\x02 \x14zC2A\x1c\x9eK\xcd\xf2\xa3\x17\x1d\xb0w/\xfa\xed\xea@\xf0\xc0\rx\xb2`\x06R\x94\x9a\xcc\xed\x02!\x00\x8eY3Q\x9d\x91T\xc7\xe6\x05\xf2\x8f\xa2\x0c\xd8\x91\x0f\xfa\xc3\x1c\xd7\xfa6\xb2\xba\xd3\xc4$\xbb\xc5\xd2\xa5'
```
The first two bytes contain sensible values for the type of signature we're getting so I believe we're interpreting that correctly.

The only question remaining for me is why we have to discard 4 bytes instead of 2. I can't see Fulcio putting anything else into the signature and the RFC describing TLS encoding doesn't say anything about padding between values.

haydentherapper:
I think you may have just found this, but we are actually sending a slightly different struct back to the client - https://github.com/sigstore/fulcio/blob/main/pkg/api/grpc_server.go#L135

This was not intentional, but when I found this, I didn’t want to break existing clients, so we left the struct as is. The embedded SCT will be in the correct format.

haydentherapper:
That's correct, the embedded [SCT](https://github.com/sigstore/fulcio/blob/be77c26f5ed420005ab2250c8051caecec90b1be/pkg/ca/intermediateca/intermediateca.go#L118) will be the SignedCertificateTimestampList as defined in https://datatracker.ietf.org/doc/html/rfc6962#section-3.3.

Given this confusion, I'm considering just changing the detached SCT to the proper value, a `SignedCertificateTimestamp` struct. (We're having to make a breaking change to the Cosign client anyways to support embedded SCTs).

woodruffw (approved):

Issue #4, opened 2022-03-07
Finalize importable `sigstore` API


tetsuo-cpp:
Once we've written the Fulcio and Rekor clients, we should be fill out the importable API. The API should roughly line up with the subcommands that `cosign` supports.

The CLI itself (which we'll implement in #3) will be a thin wrapper that wires the `argparse` flags to the API. So the meat of the CLI logic will be covered in this task.

I expect that we'll separate this into multiple issues when we begin working on it.

tetsuo-cpp:
This is a great article that explains what `cosign` is doing under the hood: https://martinheinz.dev/blog/56

tetsuo-cpp:
One thing I'd like to do here is to use `logging` throughout the API for more fine-grained reporting so that we can ask users to turn logging to `debug` before filing issues.

Then for actual feedback that we're using `output` for, we can follow a similar model to `pip-audit`'s `AuditState` where we can let users capture and use the output in whatever way they want.

di:
Changed this to 'finalize' instead of implement and moved it to the 1.0 milestone.

woodruffw:
xref #250 for some changes that will happen to these APIs before stabilization.

woodruffw:
For visibility: #299 is going to refactor the verification API per #250, which will get us much closer to a stabilize-able public API.

mayaCostantini:
Thanks a lot @woodruffw !

tetsuo-cpp:
I'm going to make a separate issue to begin generating API docs like we do for `pip-audit`.

woodruffw:
@mayaCostantini we've just cut 0.10.0, which has the first official iteration of the stable API 🙂 

The docs are here: https://sigstore.github.io/sigstore-python/

mayaCostantini:
Amazing, thanks @woodruffw !

PR #26, opened 2022-04-12
Implement `verify` command

### LINKED ISSUE #10: Implement `verify` command

tetsuo-cpp:
Closes #10


sigstore/_verify.py
```diff
@@ -0,0 +1,136 @@
...
+    # 1) Verify that the signing certificate is signed by the root certificate and that the signing
+    #    certificate was valid at the time of signing.
+    root_cert_path = Path(__file__).parent / FULCIO_ROOT_CERT
+    if not root_cert_path.is_file():
+        # Error
+        return
+    pem_bytes = open(root_cert_path, "rb").read()
+    root = load_pem_x509_certificate(pem_bytes)
+
+    sign_date = cert.not_valid_before
```
tetsuo-cpp, line 74: This chunk leading up to the Rekor interaction is based off the example code in the linked `securesystemslib` issue.

fulcio.crt.pem
```diff
@@ -0,0 +1,13 @@
+-----BEGIN CERTIFICATE-----
```
woodruffw, line 1: Accidental check-in, or is this a test asset? If so, we should probably put it under `test/assets`

woodruffw: Oh, sorry - I see that it's a root cert now! In that case, it should probably under `sigstore` and also recorded in `setup.py` or a `MANIFEST.in` file so that it correctly shows up in source and binary distributions.

sigstore/_internal/rekor/_client.py
```diff
@@ -43,6 +43,23 @@ def from_response(cls, dict_) -> RekorEntry:
         )
 
 
+@dataclass(frozen=True)
```
woodruffw, line 46: Nitpick: We should consider using Pydantic's `BaseModel` + `Field` here, so that we can do simplify to this:

```python
class RekorInclusionProof(BaseModel):
    log_index: int = Field(..., alias="logIndex")
    # other fields
```

That removes the need for the custom `from_dict` constructor + eliminates a potential source of errors via discrepancies.

jku:
Are there any plans to provide the same functionality as a module -- or is that future work? 

For TUF use case I imagine I would want to to call various sigstore-python methods with bytes as arguments (instead of file paths). So even if the module is not public yet, maybe it would make sense for the _verify module methods to take bytes as input and for the _cli module to read the files?

tetsuo-cpp:
@jku Yeah, I have #4 to expose some of this functionality as APIs. At the moment we're just focusing on getting things working.

sigstore/_internal/rekor/_client.py
woodruffw: This is tiny, so let's postpone.

tetsuo-cpp:
Still debugging this so I've moved back to draft. The `verify_certificate()` from PyOpenSSL is failing so I need to figure out why.

I also need to modify the call to Rekor to search with the hash + signing key.

tetsuo-cpp:
Ok, the Rekor search is fixed. Now just to debug the failing PyOpenSSL call...

tetsuo-cpp:
Ok this should be ready to go now.

Turns out I had the wrong Fulcio root certificate. The public instance is using the one at `fulcio_v1.crt.pem` and not `fulcio.crt.pem`.

sigstore/_internal/rekor/_client.py
tetsuo-cpp: No worries. Did this in [ac03fd1](https://github.com/trailofbits/sigstore-python/pull/26/commits/ac03fd197b6324304c42713eb09d9e27192016a2).

woodruffw (approved):
LGTM!

PR #36, opened 2022-04-24
Implement Signed Entry Timestamp (SET) verification

### LINKED ISSUE #33: Implement Signed Entry Timestamp (SET) verification
As part of the `verify` command, we should verify the SET that we get from Rekor.

tetsuo-cpp:
Closes  #33


sigstore/_internal/set.py
```diff
@@ -2,8 +2,52 @@
...
+
+import cryptography.hazmat.primitives.asymmetric.ec as ec
+import jcs  # type: ignore
+from cryptography.exceptions import InvalidSignature
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.serialization import load_pem_public_key
+
 from sigstore._internal.rekor import RekorEntry
 
+REKOR_ROOT_PUBKEY = resources.read_binary("sigstore._store", "rekor.pub")
```
tetsuo-cpp, line 17: I will allow the user to provide these (any default to using the ones in the package) in a separate PR since we're doing this in a few places. Let's do it like this for now.

setup.py
```diff
@@ -32,6 +32,7 @@
     install_requires=[
         "click>=8",
         "cryptography",
+        "jcs",
```
tetsuo-cpp, line 35: How do we feel about bringing this in as a dependency? The actual PyPI distribution is maintained by someone else but the code is developed here: https://github.com/cyberphone/json-canonicalization

Even if we decide to hand roll this in the future, I think we should use this in the short term to keep things moving as it's probably a decent amount of work to get this right.

woodruffw: I'm personally okay with it in terms of keeping things moving, but I agree that we should try to replace it in the medium term.

(We should look into what TUF's codebase does, since I believe they also have to canonicalize JSON for various tasks.)

woodruffw (approved):
LGTM overall; @tetsuo-cpp do you have a sense of how difficult it would be to add a minimal test for this functionality?

setup.py
di: Looks like TUF rolls it's own canonicalizer: https://github.com/theupdateframework/python-tuf/blob/31ca674aa0923e54d7486e5b4a068efb82a34641/tuf/api/serialization/json.py#L90-L104

I'd rather pull in `tuf` as a dependency since we're going to need it eventually anyways.

di (approved):
LGTM aside from the `jcs` dependency.

setup.py
tetsuo-cpp: I've pulled in `securesystemslib` since that's what's doing the canonicalisation under the hood. The API that `tuf` exposes with the `Signed` type isn't useful for us so I haven't brought in `tuf`.

tetsuo-cpp:
> LGTM overall; @tetsuo-cpp do you have a sense of how difficult it would be to add a minimal test for this functionality?

Shouldn't be difficult I don't think. But let's get everything in for the PyCon goals and I can spend some time unit testing each of these components afterwards.

PR #38, opened 2022-04-25
Implement verification of Merkle inclusion proof

### LINKED ISSUE #34: Implement verification of Merkle inclusion proof
As part of verification, we get an inclusion proof from Rekor. At the moment we have a `verify_merkle_inclusion` function which is currently a no-op but we should fill it in.

tetsuo-cpp:
Closes #34

### DIFF
--- a/sigstore/_internal/merkle.py
+++ b/sigstore/_internal/merkle.py
@@ -44,6 +44,7 @@ def _chain_inner(seed: bytes, hashes: List[str], log_index: int) -> bytes:
     are ordered from lower levels to upper, and |seed| is the initial subtree/leaf hash on the path
     located at the specified |index| on its level.
     """
+
     for i in range(len(hashes)):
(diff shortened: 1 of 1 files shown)


sigstore/_internal/merkle.py
```diff
@@ -1,9 +1,115 @@
...
+
+class InvalidInclusionProofError(Exception):
     pass
+
+
+LEAF_HASH_PREFIX = 0
+NODE_HASH_PREFIX = 1
+
+
+def _decomp_inclusion_proof(index: int, size: int) -> Tuple[int, int]:
```
tetsuo-cpp, line 26: Some of this is pretty cryptic. I've basically translated directly from the Trillian Go source code. Not sure if I should try to simplify this or rewrite it.

woodruffw (approved):
LGTM

PR #42, opened 2022-04-26
Verify that timestamp is before the signing certificate expiry

### LINKED ISSUE #35: Verify that timestamp is before the signing certificate expiry
We should check the signing timestamp that Rekor gives us and ensure that it is before the expiry of the signing certificate.

tetsuo-cpp:
Closes #35

### DIFF
--- a/sigstore/_verify.py
+++ b/sigstore/_verify.py
@@ -3,6 +3,7 @@ API for verifying artifact signatures.
 """
 
 import base64
+import datetime
 import hashlib
 from importlib import resources
 from typing import Optional, TextIO, cast
@@ -71,6 +72,8 @@ def verify(
     #
     # 4) Verify the inclusion proof supplied by Rekor for this artifact
     # 5) Verify the Signed Entry Timestamp (SET) supplied by Rekor for this artifact
+    # 6) Verify that the signing certificate was valid at the time of signing by comparing the
+    #    expiry against the integrated timestamp
 
     # 1) Verify that the signing certificate is signed by the root certificate and that the signing
     #    certificate was valid at the time of signing.
@@ -146,6 +149,18 @@ def verify(
             verify_set(entry)
         except InvalidSetError as inval_set:
             output(f"Failed to validate Rekor entry's SET: {inval_set}")
+            continue
+
+        # 6) Verify that the signing certificate was valid at the time of signing
+        integrated_time = datetime.datetime.fromtimestamp(entry.integrated_time)
+        if (
+            integrated_time < cert.not_valid_before
+            or integrated_time > cert.not_valid_after
+        ):
+            # No need to log anything here.
+            #
+            # If an artifact has been signed multiple times, this will happen so it's not really an
+            # error case.
             continue
 
         valid_sig_exists = True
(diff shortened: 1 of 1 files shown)


sigstore/_verify.py
```diff
@@ -148,6 +151,18 @@ def verify(
...
+        if (
+            integrated_time < cert.not_valid_before
+            or integrated_time > cert.not_valid_after
+        ):
+            # No need to log anything here.
+            #
+            # If an artifact has been signed multiple times, this will happen so it's not really an
+            # error case.
+            continue
+
```
woodruffw, line 165: I might be misunderstanding the behavior here -- does this actually verify the timestamp? It seems as though it just `continue`s if the timestamp is invalid.

di: IIUC, this is looking at every matching entry, and checking if any of the entries match the window. If none of them match, this loop will never get to the `valid_sig_exists = True` line and the validation will fail.

sigstore/_verify.py
```diff
@@ -148,6 +151,18 @@ def verify(
             output(f"Failed to validate Rekor entry's SET: {inval_set}")
             continue
 
+        # 6) Verify that the signing certificate was valid at the time of signing
+        integrated_time = datetime.datetime.utcfromtimestamp(entry.integrated_time)
+        if (
+            integrated_time < cert.not_valid_before
+            or integrated_time > cert.not_valid_after
```
di, line 158: Just want to confirm that neither of these should be inclusive? Specifically what should happen when `integrated_time == cert.not_valid_before`?

woodruffw: RFC 5280 isn't completely explicit about this, but I believe `notBefore` is inclusive and `notAfter` is exclusive. 

Quote https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.5:

> The field is represented as a SEQUENCE of two dates:
   the date on which the certificate validity period begins (notBefore)
   and the date on which the certificate validity period ends
   (notAfter).  Both notBefore and notAfter may be encoded as UTCTime or
   GeneralizedTime.

My reading of that is: "the certificate validity begins with `notBefore` (meaning inclusive) and ends with `notAfter` (meaning exclusive)".

woodruffw: Whoops, my language there is confusing: `time == notBefore` is valid and `time == notAfter` is invalid, based on my interpretation.

woodruffw: [d574759](https://github.com/trailofbits/sigstore-python/pull/42/commits/d574759123981512c7ab2f72172943b74c09931e)

woodruffw (approved):

sigstore/_verify.py
tetsuo-cpp: Good eye. Thanks.

di: Looks like d574759 does the opposite of what I'd expect?

di: Oh, this is the failure case. 👍

woodruffw:
Yep, that’s also what threw me off when describing it 🙂

PR #47, opened 2022-04-28
github/workflows: add release workflow

woodruffw:
This adds a release workflow for automating the creation of GitHub and PyPI releases whenever we release a new version of `sigstore-python`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,7 @@ jobs:
           curl -H \
             "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" \
             "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=sigstore" \
-          | jq .value \
+          | jq -r .value \
         )
 
         # sign all package distributions using the OIDC identity
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,8 +32,10 @@ jobs:
 
         # retrieve the OIDC identity
         identity_token=$( \
-          curl -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" "$ACTIONS_ID_TOKEN_REQUEST_URL" \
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+  pypi:
+    name: upload release to PyPI
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-python@v2
+
+    - name: deps
+      run: python -m pip install -U setuptools build wheel
```
di, line 22:
```suggestion
      run: python -m pip install -U build
```

These two should be unnecessary in the local environment if we're doing PEP 517 isolated builds with `build`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
+on:
+  release:
+    types:
+      - published
+
+name: release
+
+jobs:
+  pypi:
+    name: upload release to PyPI
```
di, line 14:
```suggestion
    name: Build, sign and publish release to PyPI
```

woodruffw:
This should be about ready to go. I'll do some local sanity checking, do another RC now that #49 is in, and then finalize this.

.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
...
+      run: python -m pip install -U build
+
+    - name: build
+      run: python -m build
+
+    - name: sign
+      run: |
+        # use the latest development changes for sigstore, at least
+        # until things are stabilized further
+        make dev && source env/bin/activate
```
di, line 31: This could also be:
```suggestion
        python -m pip install .
```
Not sure we need to invoke the entire development environment and it's dependencies here.

woodruffw: Good point, force of habit 🙂

woodruffw:
Blocked on #54.

di (approved):

Issue #51, opened 2022-04-28
Bandit scanning


lukehinds:
If you folks would like we could bring Bandit into CI. I am maintainer over there if we needed anything in particular. We should not feel obliged though, whatever you all see as fitting.

di:
I'm in favor!

PR #53, opened 2022-04-28
Implement Bandit CI Tests

lukehinds:
Resolves: #51

Signed-off-by: Luke Hinds <lhinds@redhat.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ from setuptools import find_packages, setup
 
 version = {}
 with open("./sigstore/_version.py") as f:
-    exec(f.read(), version) # nosec
+    exec(f.read(), version) # nosec B102
 
 with open("./README.md") as f:
     long_description = f.read()
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ from setuptools import find_packages, setup
 
 version = {}
 with open("./sigstore/_version.py") as f:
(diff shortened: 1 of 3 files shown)


lukehinds:
So we can see a failure on this, there are two ways of dealing with false postives.

One is to append nosec e.g. ` some_code # nosec`

Or if you prefer to not run a particular test, we can add the test number (which in the failed case would be `B102`: https://bandit.readthedocs.io/en/1.7.4/plugins/b102_exec_used.html

woodruffw:
> Or if you prefer to not run a particular test, we can add the test number (which in the failed case would be `B102`: https://bandit.readthedocs.io/en/1.7.4/plugins/b102_exec_used.html

Yeah, let's exclude this specific test on that specific result: it's an intentional use of `exec`.

setup.py
```diff
@@ -18,7 +18,7 @@
 
 version = {}
 with open("./sigstore/_version.py") as f:
-    exec(f.read(), version)
+    exec(f.read(), version) # nosec
```
woodruffw, line 21:
```suggestion
    exec(f.read(), version) # nosec B102
```

woodruffw:
LGTM modulo restricting the `nosec`!

woodruffw (approved):

Issue #31, opened 2022-04-22
Support ambient credential detection


di:
Cosign supports 'ambient credential detection' for a number of environments where OIDC identities are available by default. We should also similarly support:

- [x] GitHub Actions (#59)
- [x] Google Cloud:  VMs, GKE clusters, Cloud Build, etc. (#88, #91)
- [x] GitLab (SaaS): https://docs.gitlab.com/ee/integration/openid_connect_provider.html
  - blocked on https://github.com/sigstore/fulcio/issues/243 (https://github.com/sigstore/fulcio/pull/983)
    - blocked on https://gitlab.com/groups/gitlab-org/-/epics/7335
- [x] CircleCI: https://circleci.com/docs/2.0/openid-connect-tokens/
  - blocked on https://github.com/sigstore/fulcio/issues/591
    - blocked on https://circleci.canny.io/cloud-feature-requests/p/customizable-audience-claim-in-oidc-tokens
  - implementation: #72
- [x] BuildKite (#499)
  - Fulcio: https://github.com/sigstore/fulcio/pull/890

See also https://dlorenc.medium.com/a-bit-of-ambiance-comes-to-sigstore-f80d1d6b1c30

This issue is tracking support for SaaS products and not self-hosted instances, e.g. GitLab's hosted product and not their on-premise or self-hosted services. Self-hosted services are out-of-scope, pending further discussion with Fulcio.

woodruffw:
Specifically, for GitHub action, we'll need:

1. To confirm that we're running on a GitHub Action
2. To do the Python equivalent of this `curl`:
    ```bash
    curl -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=sigstore"
    ```
3. Parse the JSON response and grab the `.value` key, which contains the actual JWT

jglick:
Jenkins [supports OIDC identities](https://plugins.jenkins.io/oidc-provider/) including customizable `aud`ience, but I am guessing this system only works from a SaaS for which you can register a specific permanent `iss`uer? The discussion of GitLab leaves it unclear whether self-hosted would be supported, or only the SaaS. (Of course if you run an agent on GCP then you could use its service account though that seems like a lower-level identity.)

jglick:
[...]

PR #59, opened 2022-04-30
internal/oidc, test: ambient credentials, refactoring

woodruffw:

#### Summary

Refactors `sigstore._internal.oidc` into a structured (directory) module and adds `sigstore._internal.oidc.ambient` for ambient credential detection. See #31.

#### Ticket Link

See #31. Does not close, since other detectors are needed.

### DIFF
--- a/sigstore/_internal/oidc/ambient.py
+++ b/sigstore/_internal/oidc/ambient.py
@@ -1,0 +1,92 @@
+# Copyright 2022 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Ambient OIDC credential detection for sigstore.
+"""
+
+import logging
+import os
+from typing import Callable, List, Optional
+
+import requests
+from pydantic import BaseModel
+
+from sigstore._internal.oidc import IdentityError
+
+logger = logging.getLogger(__name__)
+
+
+class AmbientCredentialError(IdentityError):
+    """
+    Raised when an ambient credential should be present, but
+    can't be retrieved (e.g. network failure).
+    """
+
+    pass
+
+
+def detect_credential() -> Optional[str]:
+    """
+    Try each ambient credential detector, returning the first one to succeed
+    or `None` if all fail.
+
+    Raises `AmbientCredentialError` if any detector fails internally (i.e.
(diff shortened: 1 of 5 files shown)


woodruffw:
N.B.: I also moved OAuth flow support into its own module (`sigstore._internal.oidc.oauth`) to keep the top-level relatively clean.

sigstore/_internal/oidc/ambient.py
```diff
@@ -0,0 +1,92 @@
...
+    # the environment and are only present if the workflow has sufficient permissions.
+    req_token = os.getenv("ACTIONS_ID_TOKEN_REQUEST_TOKEN")
+    req_url = os.getenv("ACTIONS_ID_TOKEN_REQUEST_URL")
+    if not req_token or not req_url:
+        raise AmbientCredentialError(
+            "GitHub: missing or insufficient OIDC token permissions?"
+        )
+
+    resp = requests.get(req_url, headers={"Authorization": f"bearer {req_token}"})
+    if not resp.ok:
```
tetsuo-cpp, line 83: Maybe this? I've been following this pattern so far and I find it useful because the stacktrace shows more information about the http error (it prints a text explanation matching the status code). I guess you can add the same information in the `AmbientCredentialError` message, but this seems easier to do consistently.
```suggestion
    try:
        resp.raise_for_status()
    except requests.HTTPError as http_error:
        raise AmbientCredentialError(
            f"GitHub: OIDC token request failed (code={resp.status_code})") from http_error
```

tetsuo-cpp (dismissed):
LGTM! I think it'd be good to get this in now since I have some changes that need to go on top of the `oauth` module.

@woodruffw I'll follow up with a change to amend the release workflow to use this.

sigstore/_internal/oidc/ambient.py
woodruffw: Yep, looks good. I need to get over my exception aversion 🙂

woodruffw:
> I'll follow up with a change to amend the release workflow to use this.

Sounds good!

sigstore/_internal/oidc/ambient.py
```diff
@@ -0,0 +1,97 @@
...
+class AmbientCredentialError(IdentityError):
+    """
+    Raised when an ambient credential should be present, but
+    can't be retrieved (e.g. network failure).
+    """
+
+    pass
+
+
+def detect_credential() -> Optional[str]:
```
woodruffw, line 40: We don't need to address this in this PR, but I think `Optional[Identity]` would be a better signature here -- that way we could assert our validity guarantees about the token (e.g. having the right audience and being issued by a party we know).

Thoughts @tetsuo-cpp?

woodruffw: (If we were to do this, we should also refactor our top-level `sign` API to take an `Identity` instead of constructing one internally.)

tetsuo-cpp: Yeah I think that's a good idea. I'm going to take a stab at #4 in the coming days so it might be good to tackle it then so we get the interfaces right.

tetsuo-cpp (approved):

PR #62, opened 2022-04-30
Support configurable OIDC issuers

tetsuo-cpp:
Signed-off-by: Alex Cameron <asc@tetsuo.sh>

#### Summary

This PR adds support for configurable OIDC issuers. An `oidc-issuer` flag has been added for this purpose. `sigstore` will query the issuer's `.well-known/openid-configuration` to find its endpoints and then use them to retrieve an OIDC token. The following flags have also been exposed since they need to be customised on a per issuer basis:
- `oidc-client-id`
- `oidc-client-secret`

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -44,6 +44,7 @@ def main():
     "--ctfe",
     type=click.File("rb"),
     default=resources.open_binary("sigstore._store", "ctfe.pub"),
+    help="A PEM-encoded public key for the CT log"
 )
 @click.option(
     "oidc_client_id",
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -51,6 +51,7 @@ def main():
     metavar="ID",
     type=click.STRING,
     default="sigstore",
+    help="The custom OpenID Connect client ID to use",
 )
 @click.option(
     "oidc_client_secret",
@@ -58,6 +59,7 @@ def main():
     metavar="SECRET",
     type=click.STRING,
     default=str(),
+    help="The custom OpenID Connect client secret to use",
 )
 @click.option(
     "oidc_issuer",
@@ -65,6 +67,7 @@ def main():
     metavar="URL",
     type=click.STRING,
     default="https://oauth2.sigstore.dev/auth",
+    help="The custom OpenID Connect issuer to use",
 )
 @click.option(
     "oidc_disable_ambient_providers",
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -28,7 +28,7 @@ def main():
 
 
 @main.command("sign")
-@click.option("identity_token", "--identity-token", type=click.STRING)
+@click.option("identity_token", "--identity-token", metavar="TOKEN", type=click.STRING)
 @click.option(
     "ctfe_pem",
     "--ctfe",
@@ -36,14 +36,23 @@ def main():
     default=resources.open_binary("sigstore._store", "ctfe.pub"),
 )
 @click.option(
-    "oidc_client_id", "--oidc-client-id", type=click.STRING, default="sigstore"
+    "oidc_client_id",
+    "--oidc-client-id",
+    metavar="ID",
+    type=click.STRING,
+    default="sigstore",
 )
 @click.option(
-    "oidc_client_secret", "--oidc-client-secret", type=click.STRING, default=str()
+    "oidc_client_secret",
+    "--oidc-client-secret",
+    metavar="SECRET",
+    type=click.STRING,
+    default=str(),
 )
 @click.option(
     "oidc_issuer",
     "--oidc-issuer",
+    metavar="URL",
     type=click.STRING,
     default="https://oauth2.sigstore.dev/auth",
 )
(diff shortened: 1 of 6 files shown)


tetsuo-cpp:
This works fine for the default issuer. I'm still looking for another issuer that I can use to test this without making some kind of billing account.

sigstore/_cli.py
```diff
@@ -34,10 +35,24 @@ def main():
...
+    "oidc_client_id", "--oidc-client-id", type=click.STRING, default="sigstore"
+)
+@click.option(
+    "oidc_client_secret", "--oidc-client-secret", type=click.STRING, default=str()
+)
+@click.option(
+    "oidc_issuer",
+    "--oidc-issuer",
+    type=click.STRING,
+    default="https://oauth2.sigstore.dev/auth",
```
woodruffw, line 48: Might want to give this a `metavar="URL"`, so that the CLI doesn't just render its as `--oidc-issuer TEXT`.

woodruffw, line 48: cc @di thoughts on removing the `oidc` prefix from these flags? AFAICT they don't have equivalents in `cosign`, so changing them here wouldn't be a compatibility concern and would make the CLI slightly less verbose.

sigstore/_internal/oidc/issuer.py
```diff
@@ -0,0 +1,54 @@
...
+            f"{base_url}/", ".well-known/openid-configuration"
+        )
+
+        resp: requests.Response = requests.get(oidc_config_url)
+        try:
+            resp.raise_for_status()
+        except requests.HTTPError as http_error:
+            raise IssuerError from http_error
+
+        self.struct = resp.json()
```
woodruffw, line 40: `self._struct` perhaps? Alternatively, if we don't need any other fields from the OpenID configuration payload, it might be worth not binding this to `self`.

woodruffw:
N.B.: I think `sign` will currently fail for issuers that aren't included in `OIDC_ISSUERS`, since we do this check:

```py
        if iss not in OIDC_ISSUERS:
            raise IdentityError(f"Not a valid OIDC issuer: {iss!r}")
```

sigstore/_cli.py
tetsuo-cpp: These flags come from `cosign sign`. They're experimental though which is why you may not have seen them. From the help text:
```
      --oidc-client-id string                                                                    [EXPERIMENTAL] OIDC client ID for application (default "sigstore")
      --oidc-client-secret string                                                                [EXPERIMENTAL] OIDC client secret for application
      --oidc-issuer string                                                                       [EXPERIMENTAL] OIDC provider to be used to issue ID token (default "https://oauth2.sigstore.dev/auth")
```

sigstore/_internal/oidc/issuer.py
tetsuo-cpp: [50e342b](https://github.com/sigstore/sigstore-python/pull/62/commits/50e342b05e24a7d52e5f2fd9254ec357b3647682)

tetsuo-cpp:
> N.B.: I think `sign` will currently fail for issuers that aren't included in `OIDC_ISSUERS`, since we do this check:
> 
> ```python
>         if iss not in OIDC_ISSUERS:
>             raise IdentityError(f"Not a valid OIDC issuer: {iss!r}")
> ```
> 
> But perhaps that's intended behavior?

Hmm, good point. I'll have a look and see how `cosign` handles this.

sigstore/_cli.py
woodruffw: Thanks! Okay, let's keep them as-is then.

di: +1, we should try to stay as close as possible to `cosign`'s CLI.

di:
As far as I can tell `cosign` doesn't handle it. I don't think it makes sense for us to maintain a separate list of the issuers that Fulcio supports, we should just let the user try to submit the certificate request and handle it if Fulcio fails.

tetsuo-cpp:
> As far as I can tell `cosign` doesn't handle it. I don't think it makes sense for us to maintain a separate list of the issuers that Fulcio supports, we should just let the user try to submit the certificate request and handle it if Fulcio fails.

Do you mean that it should just assume that the signature is the `email` signed? It's not going to work for Github Actions though. Maybe it should just default to using `email` for unrecognized issuers.

woodruffw:
> Do you mean that it should just assume that the challenge is the `email` signed? It's not going to work for Github Actions though. Maybe it should just default to using `email` for unrecognized issuers.

Based on my reading of https://github.com/sigstore/fulcio/blob/f6016fda1ec6f1bd550699c7b8f53c2057ada174/pkg/challenges/challenges.go#L437-L478, I think we should default to `sub` for most unrecognized issuers. That'll cause the proof of possession to fail if it's the wrong claim, but all of this should be obviated once we're using X.509 CSRs.

That's what I've done in [e812ec3](https://github.com/sigstore/sigstore-python/pull/62/commits/e812ec307489ad30b8a48437030bcf01113afebc).

tetsuo-cpp (commented):
Nice. This is looking good.

I can't approve this since I opened this PR. If one of you could approve, I think this is good to go.

PR #65, opened 2022-05-02
Refactor: add logging, better isolate output

woodruffw:
This is just general refactoring: it adds use of `logging` in the CLI and signing modules, and makes `sign` return a `SigningResult` for eventual rendering instead of plumbing `click.echo` through a non-CLI API.


di (approved):
LGTM, but let's make sure the order of the output remains the same as what `cosign` currently does.

PR #66, opened 2022-05-02
Publish cert to Rekor instead of pubkey

di:
This fixes a small bug where we were publishing the public key with the Rekor entry instead of the signing certificate (which includes the public key).

(cc @eddiezane)


woodruffw (dismissed):
LGTM!

woodruffw (approved):

PR #71, opened 2022-05-05
cli, verify: refactor, add logging

woodruffw:
Similar to the refactor in #65.

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -150,7 +150,7 @@ def _verify(files, certificate_path, signature_path, cert_email):
             certificate=certificate,
             signature=signature,
             cert_email=cert_email,
-        ) is not None:
+        ):
             click.echo(f"OK: {file.name}")
         else:
             click.echo(f"FAIL: {file.name}")
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -150,7 +150,7 @@ def _verify(files, certificate_path, signature_path, cert_email):
             certificate=certificate,
             signature=signature,
             cert_email=cert_email,
-        ):
+        ) is not None:
             click.echo(f"OK: {file.name}")
         else:
             click.echo(f"FAIL: {file.name}")
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -143,13 +143,18 @@ def _verify(files, certificate_path, signature_path, cert_email):
     logger.debug(f"Using signature from: {signature_path.name}")
     signature = signature_path.read()
 
+    verified = True
     for file in files:
-        click.echo(
-            verify(
-                file=file,
-                certificate=certificate,
-                signature=signature,
-                cert_email=cert_email,
-                output=click.echo,
-            )
-        )
+        if verify(
+            file=file,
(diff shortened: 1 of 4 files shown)


sigstore/_cli.py
```diff
@@ -143,13 +143,18 @@ def _verify(files, certificate_path, signature_path, cert_email):
...
+        if verify(
+            file=file,
+            certificate=certificate,
+            signature=signature,
+            cert_email=cert_email,
+        ):
+            click.echo(f"OK: {file.name}")
+        else:
+            click.echo(f"FAIL: {file.name}")
+            verified = False
```
woodruffw, line 157: N.B.: `cosign` doesn't seem to have a standard non-JSON output for verification yet, so I just made something up here. Longer term, we'll probably want some sort of `VerificationResult` model that's either truthy or falsey that we can display for each step.

lukehinds: I think that's fine creating your own, you don't need to replicate cosign, its possible you might think of a better way of implementation.

di: Generally our philosophy here is to be as much of a drop-in replacement for the parts of cosign that we care about as possible. But I agree, where cosign has no precedent we can do what's best for us, but we should also make sure to make an issue to 'upstream' this feature so we don't diverge significantly.

woodruffw: Makes sense. Given that, I think I'll spend some more time on this PR getting the `VerificationResult` in place. Thinking more about it, the signature should really be `Optional[VerificationResult]`, since verification is all-or-none.

di: Agreed!

sigstore/_cli.py
```diff
@@ -143,13 +143,18 @@ def _verify(files, certificate_path, signature_path, cert_email):
...
-                cert_email=cert_email,
-                output=click.echo,
-            )
-        )
+        if verify(
+            file=file,
+            certificate=certificate,
+            signature=signature,
+            cert_email=cert_email,
+        ) is not None:
```
di, line 153: I think we should probably do something slightly more here than check if it's not `None`. I realize that `VerificationResult`s are currently just a stub, but maybe we can add `VerificationResult.success` or `.verified` or even make it truthy/falsey so we can always return a result and do something like `if all(results):`?

woodruffw: Yeah, I'm of two minds on this: it would be nice for `VerificationResult` to be `__bool__` so that we could have a description of why verification failed, but the current failure modes are basically descriptionless (signature failure is a pure function of the inputs, and Rekor failure is essentially just "we couldn't find a record"). OTOH the success modes have _some_ possible describable state (such as a dump of the Rekor entry/entries).

But yeah, `is not None` is not very descriptive. I'll change it to be `__bool__` and make it an unconditional return.

woodruffw: [febfae7](https://github.com/sigstore/sigstore-python/pull/71/commits/febfae7cbaa2ceaa19248678c54f89803a276ca9)

woodruffw:
This should be good to go now.

di (approved):

PR #74, opened 2022-05-10
fulcio, _sign: Request certificates via CSR

### LINKED ISSUE #29: Request certificates via CSR
Fulcio will soon support Certificate Signing Requests (https://github.com/sigstore/fulcio/pull/527 is merged but not deployed at the time of writing).

Once deployed, we should switch to using CSRs instead, using roughly this pattern:

https://github.com/trailofbits/sigstore-python/blob/6cb0b25127ab21f90704145458df6e454d6a83e1/sigstore/_sign.py#L35-L50

tetsuo-cpp:
Closes #29

### DIFF
--- a/sigstore/_sign.py
+++ b/sigstore/_sign.py
@@ -17,15 +17,14 @@ import hashlib
 import logging
 from typing import BinaryIO
 
+import cryptography.x509 as x509
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.hazmat.primitives.serialization import load_pem_public_key
+from cryptography.x509.oid import NameOID
 from pydantic import BaseModel
 
-from sigstore._internal.fulcio import (
-    FulcioCertificateSigningRequest,
-    FulcioClient,
-)
+from sigstore._internal.fulcio import FulcioClient
 from sigstore._internal.oidc import Identity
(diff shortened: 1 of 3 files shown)


sigstore/_sign.py
```diff
@@ -63,41 +62,28 @@ def sign(file: BinaryIO, identity_token: str, ctfe_pem: bytes) -> SigningResult:
...
-
-    # NOTE: The proof here is not a proof of identity, but solely
-    # a proof that we possess the private key. Fulcio verifies this
-    # signature using the public key we give it and the "subject",
-    # which is a function of one or more claims in the identity token.
-    # An identity token whose issuer is unknown to Fulcio will fail this
-    # test, since Fulcio won't know how to construct the subject.
-    # See: https://github.com/sigstore/fulcio/blob/f6016fd/pkg/challenges/challenges.go#L437-L478
-    signed_proof = private_key.sign(
-        oidc_identity.proof.encode(), ec.ECDSA(hashes.SHA256())
```
woodruffw, line 98: I might be missing it, but how to we do the proof-of-possession now? Does it all happen internally within `builder`?

di: Yes, the CSR itself is the signed proof, see https://cryptography.io/en/latest/x509/tutorial/#creating-a-certificate-signing-request-csr.

di (approved):
LGTM, assuming that the Fulcio-specific serialization is correct.

PR #76, opened 2022-05-10
Update gitignore

di:
Adds the `build` directory and updates the file types we're ignoring to actually ignore them in the root directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ build
 # Don't ignore these files when we intend to include them
 !sigstore/_store/*.crt
 !sigstore/_store/*.pem
+!sigstore/_store/*.pub
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ build
 *.pem
 *.sh
 *.pub
+
+# Don't ignore these files when we intend to include them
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -6,11 +6,12 @@ __pycache__/
...
-./*.sig
-./*.pem
-./*.sh
-./*.pub
+*.txt
+*.crt
+*.sig
+*.pem
+*.sh
+*.pub
```
woodruffw, line 17: I think this will also ignore `sigstore/_store/*.{pem,crt}`, which isn't a problem for the already checked-in roots but might be a problem for future ones.

di: 4ae4097

.gitignore
```diff
@@ -6,11 +6,16 @@ __pycache__/
...
+
+# Don't ignore these files when we intend to include them
+!sigstore/_store/*.crt
+!sigstore/_store/*.pem
```
woodruffw, line 21:
```suggestion
!sigstore/_store/*.pem
!sigstore/_store/*.pub
```

woodruffw (approved):
LGTM!

PR #77, opened 2022-05-10
pyproject: pin pyjwt>=2.1

woodruffw:
We haven't tested with anything earlier.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
   "cryptography",
   "pem",
   "pydantic",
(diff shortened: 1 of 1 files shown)


di (commented):
This is pretty restrictive, seems like we could probably go back to `pyjwt>=2` without any problems.

woodruffw:
Looks like we might want to start at 2.1, since that's when PyJWT began splitting signature verification from claim verification: https://github.com/jpadilla/pyjwt/blob/master/CHANGELOG.rst#v210

di (approved):

Issue #27, opened 2022-04-12
Add CLI options compatible with `cosign` where applicable


tetsuo-cpp:
We should look at what flags that `cosign` provides and consider adding them to `sigstore-python`. Some examples that we've already identified include controlling the URL of the Fulcio and Rekor instances that we call out to.

woodruffw:
Closing this in favor of more precise issues going forward, now that we're at parity with the relevant parts of `cosign` 🙂

PR #84, opened 2022-05-11
Support verifying embedded SCTs

### LINKED ISSUE #30: Switch to verifying embedded SCTs
As mentioned at https://github.com/trailofbits/sigstore-python/pull/24#issuecomment-1102901494, Fulcio will soon be supporting embedded SCTs. We should switch over to using them at some point.

### LINKED ISSUE #95: Support other signature and hash types
We have some hardcoded assumptions that we're using ECDSA keys with SHA256 hashes to generate signatures. Fulcio supports other signing and hashing algorithms so we should allow them too. For example, the staging instance is currently using an RSA CTFE key for signing.

tetsuo-cpp:
Closes #30
Closes #95


haydentherapper:
I’d recommend continuing to support both detached and embedded SCTs. While the production environment will embed them, organizations may run their own instances of fulcio in a mode that only supports detached SCTs

woodruffw:
> I’d recommend continuing to support both detached and embedded SCTs. While the production environment will embed them, organizations may run their own instances of fulcio in a mode that only supports detached SCTs

Makes sense. Any thoughts on a search priority between the two? In other words, should we check for the sidecar version first and only fall back on the embedded path if it's not present, or should we opportunistically search for the embedded SCT?

haydentherapper:
Preferring embedded is probably the preferred approach since that'll be what's used in the community instance, and we haven't seen many examples of self-hosted Fulcio instances yet.

woodruffw:
@tetsuo-cpp in terms of verifying the embedded SCT: you might also find the code that does that for Chrome helpful: https://source.chromium.org/chromium/chromium/src/+/main:net/cert/ct_serialization.cc;l=258-284;drc=8a609a649e36c8099c6801d41aca1ec3b0b3af73;bpv=1;bpt=1

tetsuo-cpp:
Still not sure what's going on. I've opened #7267 for visibility as well.

Firstly, I verified that the TBS precertificate with the SCT list extension generated by this code is identical to what is produced by `certificate-transparency-go`.

Then I generated the entire digitally signed struct (the payload for the signature) in `certificate-transparency-go` and found that it is also identical to what this code produces. I had to hardcode a few things in the Go code during testing so there's a chance that I've reproduced my mistake in Go, but it seems unlikely.

These are some possibilities for what is wrong:

- The timestamp is in the wrong format. This seems unlikely but I did need to paste the timestamp integer into the Go code during my testing so if there's a problem with the format, my testing wouldn't have detected it. I think it's pretty straightforward though, just a milliseconds since epoch timestamp. My code does `int(sct.timestamp.timestamp() * 1e3)` which seems fine. `timestamp()` returns a float containing a "seconds since epoch" timestamp which we multiply by 1000 to get milliseconds. There doesn't appear to be an accuracy issue here as the seconds count has numbers after the decimal point.
- I've chosen the wrong issuer for the `issuer_key_hash` part of the digitally signed struct. I got my Go code to read the signing cert, chain, etc that my Python code produced so if I chose the wrong cert as the issuer, I'd have the same problem in Go. Seems unlikely because I've tried the SCT verification using every cert in the chain and none of them have worked.
- I'm using the wrong CTFE key. I got the key from https://storage.googleapis.com/tuf-root-staging which I think is correct.
- My verification call is wrong. So perhaps the digitally signed payload is fine but I'm verifying incorrectly somehow. I'm using `PKCS1v15` padding which matches what the IETF spec says.
- The signature is in the wrong format. I've read through [this article](https://letsencrypt.org/2018/04/04/sct-encoding.html#variable-length) again and had a look at how some TLS and DER prefixes had to be stripped away from the signature. I don't think this should apply to us as we would have already decoded this in the Rust. And if I do a `hexdump` on the signature, I'm not seeing anything that matches those prefixes.

haydentherapper:
> I've chosen the wrong issuer for the issuer_key_hash part of the digitally signed struct

I agree this is unlikely. The only reason that the issuer would differ between the precert and cert is in the case where the CA has issued a special signing certificate that only signs precerts. We don't use this, the same CA certificate signs both the precert and certificate.

> I'm using the wrong CTFE key. I got the key from https://storage.googleapis.com/tuf-root-staging which I think is correct.

I'm assuming you're testing against fulcio.sigstage.dev, and if so, this would be correct.

> PKCS1v15 padding

This could be a possibility. If you want to test if this is working with ECDSA signature instead, you could spin up your own local instance of Fulcio with `docker-compose up`, but there's a few more steps you'll need to follow to set up a signing backend that supports precertificate issuance. Lemme know if you're interested in this, I can write up instructions.

Some random guesses:
* The order of the extensions has changed? I ran into this issue when I was removing the poison extension in the TBS cert, I shuffled the extensions around
* Is the correct entry type specified?
* Is the correct log ID included? It's the SHA1 hash of the public key in the subjectpublickeyinfo asn.1 structure.

woodruffw:
> * The order of the extensions has changed? I ran into this issue when I was removing the poison extension in the TBS cert, I shuffled the extensions around

This is a good candidate; I'll look into this 🙂 

The entry type should be correct, since we're no longer hardcoding `0` and instead getting it directly from the SCT itself. I'll also check on the log ID.

woodruffw:
Just dumping for context: this is how I've added a development build of `cryptography` to my `sigstore-python` development environment:

```bash
LDFLAGS="-L$(brew --prefix openssl@1.1)/lib" \
  CFLAGS="-I$(brew --prefix openssl@1.1)/include" \
  pip install --editable git+https://github.com/trail-of-forks/cryptography.git@alex/tbs-precert#egg=cryptography
```

woodruffw:
> The entry type should be correct, since we're no longer hardcoding `0` and instead getting it directly from the SCT itself. I'll also check on the log ID.

Confirmed this:

```
DEBUG:sigstore._internal.sct:packing SCT: sct.version=<Version.v1: 0> sct.entry_type=<LogEntryType.PRE_CERTIFICATE: 1>
```

sigstore/_internal/sct.py
```diff
@@ -50,41 +64,69 @@ def _pack_digitally_signed(sct: DetachedFulcioSCT, cert: Certificate) -> bytes:
...
+        0,                                      # signature_type (certificate_timestamp(0))
+        int(sct.timestamp.timestamp() * 1000),  # timestamp (milliseconds)
+        sct.entry_type.value,                   # entry_type (x509_entry(0) | precert_entry(1))
+        hashlib.sha256(issuer_key).digest(),    # issuer_key_hash[32]
+        len1,                                   # \
+        len2,                                   # | opaque TBSCertificate<1..2^24-1> OR
+        len3,                                   # | opaque ASN.1Cert<1..2^24-1>
+        cert_der,                               # /
+        len(sct.extension_bytes),               # extensions (opaque CtExtensions<0..2^16-1>)
     )
```
woodruffw, line 90: I'm not sure this is the problem, but I'm realizing that the serialized structure here is only the _contents_ of the digitally signed element, not the `digitally-signed struct` itself. For that, we'd need the signature and hash algorithms, plus an `opaque<1..2^16-1>` for `data` itself. 

@haydentherapper do you know if the signature is meant to be over the contents of the `digitally-signed struct`, or whether it's supposed to include the envelope as well?

haydentherapper: See https://github.com/google/certificate-transparency-go/blob/100db00880e6af80af0132c01b09c9eb2c4ee2dc/serialization.go#L29 - The signature is over the contents of the `digitally-signed struct`. 

I agree that the SCT is missing the signature and hash. See https://letsencrypt.org/2018/04/04/sct-encoding.html for a really great blog diving into SCT encoding

haydentherapper: One issue might be that the SCT doesn't contain the TBS, that's just used to generate the signature. From the RFC:

```
struct {
   Version sct_version;
   LogID id;
   uint64 timestamp;
   CtExtensions extensions;
   digitally-signed struct {
       Version sct_version;
       SignatureType signature_type = certificate_timestamp;
       uint64 timestamp;
       LogEntryType entry_type;
       select(entry_type) {
           case x509_entry: ASN.1Cert;
           case precert_entry: PreCert;
       } signed_entry;
      CtExtensions extensions;
   };
} SignedCertificateTimestamp;
```

If I understand correctly, this means that the SCT should contain only:
* A version
* Log ID
* Timestamp
* Extensions
* Signature over a version, type, timestamp, entry type, and certificate (TBS precert or ASN.1 encoded cert) - The signature containing the hash algorithm and signature

haydentherapper: Reading https://datatracker.ietf.org/doc/html/rfc5246#section-4.7, yea, I think this is the issue, that the SCT should be composed of some metadata and a signature, rather than metadata and the certificate. It's unclear that `digitally-signed struct` means a signature over the contents of the struct.

woodruffw:
More braindump: here's how `certificate-transparency-go` converts a precertificate into a tbsCertificate: https://github.com/google/certificate-transparency-go/blob/100db00880e6af80af0132c01b09c9eb2c4ee2dc/x509/x509.go#L1227-L1303

woodruffw:
> The order of the extensions has changed? I ran into this issue when I was removing the poison extension in the TBS cert, I shuffled the extensions around

I just dumped our reconstructed `tbsCertificate`, and the order of the extensions _looks_ right, so I don't think this is it. I noticed that the `tbsCertificate` doesn't have an X.509 envelope (it's just the raw `SEQUENCE`), but IIUC that's correct (since it doesn't have a signature).

sigstore/_internal/sct.py
woodruffw: Yeah, so I _think_ our current construction is right: we only build the contents of the `digitally-signed`, which looks very close to the implementation you linked above. Closing this subthread since my initial hypothesis here was wrong (it's not an envelope problem).

woodruffw:
> * Is the correct log ID included? It's the SHA1 hash of the public key in the subjectpublickeyinfo asn.1 structure.

Where should the log ID be included in the digitall-signed payload? I see it in the SCT structure, but not in the `digitally-signed` struct that the signature is computed over.

woodruffw:
The mystery deepens: I copied `certificate-transparency-go`'s testvector for the digitally-signed payload and recreated it in our test suite, and our implementation of `_packed_digitally_signed` is consistent with theirs. So it's not the layout of the signature input and it's not the `tbsCertificate` (per @tetsuo-cpp's confirmation that our stripping is identical to theirs).

haydentherapper:
> Where should the log ID be included in the digitally-signed payload? I see it in the SCT structure, but not in the digitally-signed struct that the signature is computed over.

Correct, it's only in the SCT structure, not signed over

woodruffw:
I also added a test for the issuer key hash, although I don't have a known testvector for it.

haydentherapper:
Ah sorry, getting my wires crossed! Yes, it's sha256 - https://github.com/sigstore/cosign/blob/d5f2c329cef37770b7129963bd24580c894fed3f/pkg/cosign/tlog.go#L69

(SHA1 is for calculating the key ID for x.509 certificates)

Yes, the public key of the CT log.

sigstore/_internal/sct.py
```diff
@@ -16,75 +16,130 @@
...
 
-from sigstore._internal.fulcio import DetachedFulcioSCT
+logger = logging.getLogger(__name__)
 
 
-def _pack_digitally_signed(sct: DetachedFulcioSCT, cert: Certificate) -> bytes:
+def _cryptography_hash_to_enum(hash_: hashes.HashAlgorithm) -> int:
+    if type(hash_) is hashes.SHA256:
+        return 4
+    raise InvalidSctError(f"unexpected hash algorithm: {hash_!r}")
```
di, line 39: Unused?

woodruffw: Yep, that was from some earlier tests. I'll remove now.

woodruffw: 872bf8a..f8edf5e

sigstore/_internal/sct.py
```diff
@@ -16,75 +16,124 @@
...
-        len2,
-        len3,
-        cert_der,
-        len(sct.extensions),
+        sct.version.value,                      # sct_version
+        0,                                      # signature_type (certificate_timestamp(0))
+        int(sct.timestamp.timestamp() * 1000),  # timestamp (milliseconds)
+        sct.entry_type.value,                   # entry_type (x509_entry(0) | precert_entry(1))
+        issuer_key_hash,                        # issuer_key_hash[32]
+        len1,                                   # \
```
haydentherapper, line 75: (Feel free to ignore if you've already validated the contents here) What are each of these three lengths?

sigstore/_internal/sct.py
```diff
@@ -16,75 +16,124 @@
...
+    return hashlib.sha256(issuer_key).digest()
+
+
 class InvalidSctError(Exception):
     pass
 
 
 def verify_sct(
-    sct: DetachedFulcioSCT,
+    sct: SignedCertificateTimestamp,
```
haydentherapper, line 115: You've confirmed that the `SignedCertificateTimestamp` structure matches the RFC definition?

sigstore/_internal/sct.py
woodruffw: Yep, I have a unit test with testvectors validating this. They're the individual bytes of the `uint24_t` for the following cert/TBS data, since `struct` unfortunately doesn't support packing/unpacking 24-bit integers directly.

sigstore/_internal/sct.py
woodruffw: Yep. `SignedCertificateTimestamp` is `cryptography`'s interface, and they have a large testsuite for conformance.

sigstore/_internal/sct.py
```diff
@@ -16,75 +16,124 @@
...
+CERTIFICATE_TRANSPARENCY_OID = ObjectIdentifier("1.3.6.1.4.1.11129.2.4.4")
+
+
+def _is_preissuer(issuer: Certificate) -> bool:
+    ext_key_usage = issuer.extensions.get_extension_for_class(ExtendedKeyUsage)
+    return CERTIFICATE_TRANSPARENCY_OID in ext_key_usage.value
+
+
+def _get_issuer_cert(chain: List[Certificate]) -> Certificate:
+    issuer = chain[0]
```
haydentherapper, line 95: chain starts with the intermediate/issuing CA cert? Just want to check that chain doesn't start with the leaf

woodruffw: I believe so, but will confirm. Interestingly things *are* working for @di, just not locally for me.

haydentherapper: Which environment are you testing against, staging? And using the correct verification key?

woodruffw: > Which environment are you testing against, staging? And using the correct verification key?

Yep, staging, and we confirmed that we both have the correct key. @di is able to use both the OAuth and ambient flows correctly, so it's possible there's just something pathologically wrong with my local environment and/or a very subtle bug in OpenSSL that's only manifesting for me.

woodruffw:
I dumped the inner `tbsCertificate` for each `digitally-signed` payload, and nothing looks particularly out of place: the only differences (which are expected) are the serial number, time, public key, and subject key identifier.

Here's the diff:

```diff
--- /dev/fd/63  2022-05-27 12:47:09.000000000 -0400
+++ /dev/fd/62  2022-05-27 12:47:09.000000000 -0400
@@ -1,61 +1,61 @@
     0:d=0  hl=4 l= 433 cons: SEQUENCE          
     4:d=1  hl=2 l=   3 cons: cont [ 0 ]        
     6:d=2  hl=2 l=   1 prim: INTEGER           :02
-    9:d=1  hl=2 l=  20 prim: INTEGER           :2815BF5340ECBD5531F808860C76BFB913BC589F
+    9:d=1  hl=2 l=  20 prim: INTEGER           :404C0E5DAAC8FCBA5356AA9608CB4969E6F67B80
    31:d=1  hl=2 l=  10 cons: SEQUENCE          
    33:d=2  hl=2 l=   8 prim: OBJECT            :ecdsa-with-SHA384
    43:d=1  hl=2 l=  55 cons: SEQUENCE          
    45:d=2  hl=2 l=  21 cons: SET               
    47:d=3  hl=2 l=  19 cons: SEQUENCE          
    49:d=4  hl=2 l=   3 prim: OBJECT            :organizationName
    54:d=4  hl=2 l=  12 prim: PRINTABLESTRING   :sigstore.dev
    68:d=2  hl=2 l=  30 cons: SET               
    70:d=3  hl=2 l=  28 cons: SEQUENCE          
    72:d=4  hl=2 l=   3 prim: OBJECT            :commonName
    77:d=4  hl=2 l=  21 prim: PRINTABLESTRING   :sigstore-intermediate
   100:d=1  hl=2 l=  30 cons: SEQUENCE          
-  102:d=2  hl=2 l=  13 prim: UTCTIME           :220527155111Z
-  117:d=2  hl=2 l=  13 prim: UTCTIME           :220527160111Z
+  102:d=2  hl=2 l=  13 prim: UTCTIME           :220526201030Z
+  117:d=2  hl=2 l=  13 prim: UTCTIME           :220526202030Z
   132:d=1  hl=2 l=   0 cons: SEQUENCE          
   134:d=1  hl=2 l= 118 cons: SEQUENCE          
   136:d=2  hl=2 l=  16 cons: SEQUENCE          
   138:d=3  hl=2 l=   7 prim: OBJECT            :id-ecPublicKey
   147:d=3  hl=2 l=   5 prim: OBJECT            :secp384r1
   154:d=2  hl=2 l=  98 prim: BIT STRING        
-      0000 - 00 04 73 52 b3 60 80 94-f3 7b 62 f6 d9 7f e7 55   ..sR.`...{b....U
-      0010 - 8b 7f 25 e4 c5 ca ca dd-36 fa fa 57 7a 36 6a e3   ..%.....6..Wz6j.
-      0020 - 4f 24 39 1f 07 2f 89 e9-57 fe c8 bc d5 8c 40 40   O$9../..W.....@@
-      0030 - fb 7e 78 ea 16 da a8 64-d0 8f 2a ae 38 7b bb 70   .~x....d..*.8{.p
-      0040 - d1 42 7b 5c 0b 2e 45 6f-84 ee 13 19 fe 5e 26 2b   .B{\..Eo.....^&+
-      0050 - 90 9e fe 8a 55 2e 8c 50-ae 43 5c f1 39 43 00 4d   ....U..P.C\.9C.M
-      0060 - 12 1f                                             ..
+      0000 - 00 04 11 4a 98 7e a1 7e-3f 18 c4 53 58 bc 8e 88   ...J.~.~?..SX...
+      0010 - 0b 30 31 30 4e bd a7 54-6b a4 ae c6 67 31 6b 6c   .010N..Tk...g1kl
+      0020 - f8 f6 b1 29 4b 4a 28 65-c1 0e 51 2a 7a 19 0e 2b   ...)KJ(e..Q*z..+
+      0030 - 8f 60 f3 43 c2 d0 11 b4-30 ad 7f ec b3 ff 65 9f   .`.C....0.....e.
+      0040 - 74 27 5f 00 ae 0c 8e ca-2a 8b fd b5 fd f0 4b be   t'_.....*.....K.
+      0050 - 23 d3 61 e3 51 91 89 64-60 b9 0c 11 16 47 12 b6   #.a.Q..d`....G..
+      0060 - f1 7c                                             .|
   254:d=1  hl=3 l= 180 cons: cont [ 3 ]        
   257:d=2  hl=3 l= 177 cons: SEQUENCE          
   260:d=3  hl=2 l=  14 cons: SEQUENCE          
   262:d=4  hl=2 l=   3 prim: OBJECT            :X509v3 Key Usage
   267:d=4  hl=2 l=   1 prim: BOOLEAN           :255
   270:d=4  hl=2 l=   4 prim: OCTET STRING      
       0000 - 03 02 07 80                                       ....
   276:d=3  hl=2 l=  19 cons: SEQUENCE          
   278:d=4  hl=2 l=   3 prim: OBJECT            :X509v3 Extended Key Usage
   283:d=4  hl=2 l=  12 prim: OCTET STRING      
       0000 - 30 0a 06 08 2b 06 01 05-05 07 03 03               0...+.......
   297:d=3  hl=2 l=  29 cons: SEQUENCE          
   299:d=4  hl=2 l=   3 prim: OBJECT            :X509v3 Subject Key Identifier
   304:d=4  hl=2 l=  22 prim: OCTET STRING      
-      0000 - 04 14 b2 97 b3 1d fd 2c-a8 ed 9a 73 53 c5 db e7   .......,...sS...
-      0010 - 06 08 89 99 97 f6                                 ......
+      0000 - 04 14 58 55 86 dc fb 5c-32 c5 46 8d 8a cf a5 c9   ..XU...\2.F.....
+      0010 - 94 2d 9c c6 46 74                                 .-..Ft
   328:d=3  hl=2 l=  31 cons: SEQUENCE          
   330:d=4  hl=2 l=   3 prim: OBJECT            :X509v3 Authority Key Identifier
   335:d=4  hl=2 l=  24 prim: OCTET STRING      
       0000 - 30 16 80 14 71 86 30 a6-14 7c 62 6f f9 f7 d6 f4   0...q.0..|bo....
       0010 - 05 1a 7f 5f ff eb 6f ac-                          ..._..o.
   361:d=3  hl=2 l=  35 cons: SEQUENCE          
   363:d=4  hl=2 l=   3 prim: OBJECT            :X509v3 Subject Alternative Name
   368:d=4  hl=2 l=   1 prim: BOOLEAN           :255
   371:d=4  hl=2 l=  25 prim: OCTET STRING      
       0000 - 30 17 81 15 77 69 6c 6c-69 61 6d 40 79 6f 73 73   0...william@yoss
       0010 - 61 72 69 61 6e 2e 6e 65-74                        arian.net
   398:d=3  hl=2 l=  37 cons: SEQUENCE          
   400:d=4  hl=2 l=  10 prim: OBJECT            :1.3.6.1.4.1.57264.1.1
   412:d=4  hl=2 l=  23 prim: OCTET STRING      :github-sigstore-staging
```

@haydentherapper am I right in thinking that it's normal for the public key and subject key identifier to vary with each SCT's `digitally-signed`? The others make sense to me, but I wasn't sure about those two.

woodruffw:
> I assume this is for two different certificates, correct? The time of upload, public key, and key ID would differ.
> A precertificate and certificate's TBS representations should be identical though.

Yep, two different certificates. And yeah, the TBS representations are identical (modulo the extension we strip).

woodruffw:
@haydentherapper Turns out, it was a very silly mistake on our part: `cryptography` returns native `datetimes`, so we just needed to normalize them to UTC: [3dfe195](https://github.com/sigstore/sigstore-python/pull/84/commits/3dfe1952197a57afe2ca616478b375e837db678d)

woodruffw:
This should now be functional. I'll take some time this afternoon to do cleanup and add appropriate CLI support for switching between the prod and staging environments.

haydentherapper:
Great catch! Fantastic work on this y'all, this is a very hard-to-debug space

sigstore/_internal/sct.py
```diff
@@ -54,39 +66,58 @@ def _pack_digitally_signed(
...
         len3,
         cert_der,
-        len(sct.struct["extensions"]),
+        0,  # Extension Length
     )
 
     return data
 
 
+CERTIFICATE_TRANSPARENCY_OID = ObjectIdentifier("1.3.6.1.4.1.11129.2.4.4")
```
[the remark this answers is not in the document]
woodruffw: I don't mind it here, but maybe the `cryptography` devs want it anyways. What say you, @alex/@reaperhulk?

sigstore/_internal/oidc/oauth.py
```diff
@@ -119,7 +119,7 @@ def auth_request_params(self) -> Dict[str, str]:
 
     def auth_request(self) -> str:
         params = self.auth_request_params()
-        return "https://oauth2.sigstore.dev/auth/auth?" + urllib.parse.urlencode(params)
+        return "https://oauth2.sigstage.dev/auth/auth?" + urllib.parse.urlencode(params)
```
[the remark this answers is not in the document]
woodruffw: Done.

woodruffw:
This should be in good shape for an initial review. I confirmed locally that both the current prod and staging instances work, meaning that we're supporting both detached and embedded SCTs correctly.

In order to keep things convenient, we've added the `--staged` flag to `sigstore sign`. Using that flag will switch all of the URLs (and baked in root cert) over to the staging instances/endpoints.

sigstore/_internal/sct.py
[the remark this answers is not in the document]
alex: Sure, we're generally happy to include constants for significant OIDs.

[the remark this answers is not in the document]
woodruffw: Excellent. I'll roll it into our current PR.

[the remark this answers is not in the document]
woodruffw: Updated https://github.com/pyca/cryptography/pull/7279 to include this OID.

sigstore/_internal/sct.py
```diff
@@ -163,6 +163,11 @@ def verify_sct(
 
     digitally_signed = _pack_digitally_signed(sct, cert, issuer_key_hash)
 
+    if sct.signature_hash_algorithm == hashes.SHA256():
```
woodruffw, line 166: Did you mean `!=` here?

sigstore/_internal/sct.py
```diff
@@ -163,7 +163,7 @@ def verify_sct(
 
     digitally_signed = _pack_digitally_signed(sct, cert, issuer_key_hash)
 
-    if sct.signature_hash_algorithm == hashes.SHA256():
+    if sct.signature_hash_algorithm.name != hashes.SHA256().name:
```
woodruffw, line 166: Ugh, instance comparisons 🙂 -- maybe `type(sct.signature_hash_algorithm) is not hashes.SHA256`?

reaperhulk:
```suggestion
    if not isinstance(sct.signature_hash_algorithm,  hashes.SHA256):
```

tetsuo-cpp:
I believe these changes also address #95

di:
> In order to keep things convenient, we've added the --staged flag to sigstore sign. Using that flag will switch all of the URLs (and baked in root cert) over to the staging instances/endpoints.

Could we a) pull this out into a separate PR to discuss and b) potentially align this better with `cosign`'s `--rekor-url` flag?

woodruffw:
Yep, I can factor it out. And yeah, I’ll also look into unifying it with sigstore’s CLI.

woodruffw:
> I can remove this and redo it as part of https://github.com/sigstore/sigstore-python/issues/27 if that works for you. I think it's still useful to have a --staged flag for convenience but maybe it can just be shorthand for setting flags like --fulcio-url and --rekor-url.

Sounds good! Let's prioritize the `--rekor-url` and `--staging` PRs so that we can rebase this on top of them.

tetsuo-cpp:
I'll resolve these conflicts in just a moment. Let's get the `--staging` flag in first.

woodruffw:
Needs rebase, but otherwise the only thing blocking this is an upstream release of `cryptography`.

woodruffw:
We've hacked around the upstream requirements, for now. I'm going to push up some more changes to properly mark each hack, so that we can easily undo them later.

woodruffw (dismissed):
Provisionally approving, but @di and @tetsuo-cpp please review one last time!

(I've tested these changes locally and can confirm that `sigstore sign` and `sigstore sign --staged` both work. I don't know of a non-embedded SCT instance to test against any more, but these changes should work with one.)

tetsuo-cpp (commented):
LGTM! GH won't let me approve this since I opened the PR but the temporary changes you added look fine.

woodruffw (approved):

PR #88, opened 2022-05-13
Google Cloud ambient credential detection

di:
Adds ambient credential detection for Google Cloud. Towards #31.

### DIFF
--- a/sigstore/_internal/oidc/ambient.py
+++ b/sigstore/_internal/oidc/ambient.py
@@ -23,7 +23,7 @@ from typing import Callable, List, Optional
 import requests
 from pydantic import BaseModel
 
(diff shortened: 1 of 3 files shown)


sigstore/_internal/oidc/ambient.py
```diff
@@ -23,10 +23,15 @@
 import requests
 from pydantic import BaseModel
 
-from sigstore._internal.oidc import IdentityError
+from sigstore._internal.oidc import _AUDIENCE, IdentityError
```
woodruffw, line 26: Tiny nit: now that we're importing it somewhere else, maybe we should do `_AUDIENCE -> AUDIENCE`

woodruffw (approved):

PR #91, opened 2022-05-17
Support Google Cloud impersonation

di:
As a follow-on to #88, and towards #31, support service account impersonation, which makes it possible to use Google Cloud Build.

Also adds documentation for ambient credential detection.


woodruffw (approved):
LGTM!

PR #94, opened 2022-05-18
_internal/fulcio: refactor SCT model

woodruffw:
This replaces `FulcioSignedCertificateTimestamp` with `FulcioSCT`,
which inherits from `pydantic.BaseModel` and uses pydantic's
verification APIs instead of doing all verification ad-hoc in
the constructor.

The relationship to `cryptography.x509.SignedCertificateTimestamp`
is maintained via virtual subclassing.

This should make testing detached SCTs much easier. Marking as a draft until I add some corresponding tests.

### DIFF
--- a/sigstore/_internal/fulcio/client.py
+++ b/sigstore/_internal/fulcio/client.py
@@ -91,7 +91,7 @@ class DetachedFulcioSCT(BaseModel):
 
     version: Version = Field(..., alias="sct_version")
     log_id: bytes = Field(..., alias="id")
-    raw_timestamp: int = Field(..., alias="timestamp")
+    timestamp: datetime.datetime
     digitally_signed: bytes = Field(..., alias="signature")
     extensions: bytes
 
@@ -115,10 +115,6 @@ class DetachedFulcioSCT(BaseModel):
     @validator("extensions", pre=True)
     def _validate_extensions(cls, v: bytes) -> bytes:
         return base64.b64decode(v)
-
-    @property
-    def timestamp(self) -> datetime.datetime:
-        return datetime.datetime.fromtimestamp(self.raw_timestamp / 1000)
 
     @property
     def entry_type(self) -> LogEntryType:
--- a/sigstore/_internal/fulcio/client.py
+++ b/sigstore/_internal/fulcio/client.py
@@ -96,20 +96,20 @@ class FulcioSCT(BaseModel):
         arbitrary_types_allowed = True
 
     @validator("digitally_signed", pre=True)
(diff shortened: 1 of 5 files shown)


sigstore/_internal/fulcio/client.py
```diff
@@ -83,71 +84,65 @@ class FulcioSCTError(Exception):
     pass
 
 
-class FulcioSignedCertificateTimestamp(SignedCertificateTimestamp):
-    def __init__(self, b64_encoded_sct: str):
-        self.struct = json.loads(base64.b64decode(b64_encoded_sct).decode())
+class FulcioSCT(BaseModel):
+    version: Version = Field(..., alias="sct_version")
+    log_id: bytes = Field(..., alias="id")
+    raw_timestamp: int = Field(..., alias="timestamp")
```
woodruffw, line 90: N.B.: We have to keep this timestamp as a raw integer, since round-tripping it through `datetime.datetime` can result in us losing millisecond fidelity.

sigstore/_internal/fulcio/client.py
```diff
@@ -83,71 +84,69 @@ class FulcioSCTError(Exception):
     pass
 
 
-class FulcioSignedCertificateTimestamp(SignedCertificateTimestamp):
-    def __init__(self, b64_encoded_sct: str):
-        self.struct = json.loads(base64.b64decode(b64_encoded_sct).decode())
+class FulcioSCT(BaseModel):
```
woodruffw, line 87: N.B.: We could probably rename this to something like `DetachedFulcioSCT` once we merge embedded SCT support.

woodruffw: Nope, no reason. I can rename.

sigstore/_internal/fulcio/client.py
```diff
@@ -83,71 +84,69 @@ class FulcioSCTError(Exception):
...
-                f"got {len(digitally_signed[4:])}"
-            )
+    @validator("digitally_signed", pre=True)
+    def _validate_digitally_signed(cls, v: bytes) -> bytes:
+        digitally_signed = base64.b64decode(v)
 
-        # Finally, extract the underlying signature.
-        self.signature: bytes = digitally_signed[4:]
+        if len(digitally_signed) <= 4:
+            raise ValueError("impossibly small digitally-signed struct")
```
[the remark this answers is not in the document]
woodruffw: Yep -- each validator is called during construction, and `ValueError`s are promoted into `pydantic.ValidationError`s. We test this with the direct constructors in the unit tests, but I could add some for `parse_obj` as well.

di (approved):

PR #99, opened 2022-05-31
oauth: Fix race in OIDC token retrieval

### LINKED ISSUE #96: OAuth flow: intermittent 400 errors
**Description**

When running `sigstore` locally with the OAuth OIDC flow, I occasionally get HTTP 400s. Happens with both `sigstore.dev` and `sigstage.dev` as the instance. The error code indicates that it's almost certainly a payload construction error on our side, possibly due to raciness when setting up the local HTTP listener.

Example invocation:

```console
$ sigstore sign README.md
```

Example traceback:

```
BUG:urllib3.connectionpool:https://oauth2.sigstage.dev:443 "POST /auth/token HTTP/1.1" 400 70
  File "/Users/william/devel/sigstore-python/sigstore/_internal/oidc/oauth.py", line 193, in get_identity_token
    resp.raise_for_status()
  File "/Users/william/devel/sigstore-python/env/lib/python3.10/site-packages/requests/models.py", line 960, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://oauth2.sigstage.dev/auth/token

The above exception was the direct cause of the following exception:

    sys.exit(main())
  File "/Users/william/devel/sigstore-python/env/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/william/devel/sigstore-python/env/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/william/devel/sigstore-python/env/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/william/devel/sigstore-python/env/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/william/devel/sigstore-python/env/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Users/william/devel/sigstore-python/sigstore/_cli.py", line 110, in _sign
    identity_token = get_identity_token(
  File "/Users/william/devel/sigstore-python/sigstore/_internal/oidc/oauth.py", line 195, in get_identity_token
    raise IdentityError from http_error
sigstore._internal.oidc.IdentityError
```

tetsuo-cpp:
Closes #96

### DIFF
--- a/sigstore/_internal/oidc/oauth.py
+++ b/sigstore/_internal/oidc/oauth.py
@@ -45,6 +45,12 @@ class RedirectHandler(http.server.BaseHTTPRequestHandler):
 
     def do_GET(self) -> None:
         server = cast(RedirectServer, self.server)
+
+        # If the auth response has already been populated, the main thread will be stopping this
+        # thread and accessing the auth response shortly so we should stop servicing any requests.
+        if not server.active:
+            return
+
         r = urllib.parse.urlsplit(self.path)
 
         # Handle auth response
@@ -56,7 +62,7 @@ class RedirectHandler(http.server.BaseHTTPRequestHandler):
             self.end_headers()
             self.wfile.write(body)
             server.auth_response = urllib.parse.parse_qs(r.query)
-            return None
+            return
 
         # Any other request generates an auth request
         url = server.auth_request()
@@ -157,7 +163,7 @@ def get_identity_token(client_id: str, client_secret: str, issuer: Issuer) -> st
 
         if not server.is_oob:
             # Wait until the redirect server populates the response
-            while server.auth_response is None:
+            while server.active:
                 time.sleep(0.1)
             auth_error = server.auth_response.get("error")
             if auth_error is not None:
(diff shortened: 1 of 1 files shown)


tetsuo-cpp:
The issue is that sometimes the browser requests `/favicon.ico` after we handle the auth response. That code path doesn't seem to mutate the auth response but it does modify other fields in the server object as the main thread is reading the auth response. It's not 100% clear to me why that causes the auth response to get corrupted since I would think that the GIL prevents a torn write.

sigstore/_internal/oidc/oauth.py
```diff
@@ -45,6 +45,12 @@ def log_message(self, _format: str, *_args: Any) -> None:
 
     def do_GET(self) -> None:
         server = cast(RedirectServer, self.server)
+
+        # If the auth response has already been populated, the main thread will be stopping this
+        # thread and accessing the auth response shortly so we should stop servicing any requests.
+        if not server.active:
+            return
```
woodruffw, line 52: `return None` here and below? I'm surprised our linting tools don't enforce this...

woodruffw:
Nice debugging! Yeah, that's an interesting error case. I believe the GIL unfortunately does *not* prevent write tearing -- it really only protects the integrity of the interpreter itself 😞

di (approved):

woodruffw (approved):

PR #113, opened 2022-06-03
_cli, _verify: Wrap OpenSSL error with user-friendly text

tetsuo-cpp:


sigstore/_cli.py
```diff
@@ -294,15 +294,18 @@ def _verify(
...
             certificate=certificate,
             signature=signature,
             cert_email=cert_email,
-        ):
+        )
+        if result:
             click.echo(f"OK: {file.name}")
         else:
+            failure = cast(VerificationFailure, result)
+            click.echo(failure.reason)
```
tetsuo-cpp, line 308: This isn't related but I just noticed that the failure reason doesn't get printed at the moment.

sigstore/_verify.py
```diff
@@ -130,7 +135,13 @@ def verify(
...
     store.set_time(sign_date)
     store_ctx = X509StoreContext(store, openssl_cert)
-    store_ctx.verify_certificate()
+    try:
+        store_ctx.verify_certificate()
+    except X509StoreContextError as store_ctx_error:
+        return VerificationFailure(
+            reason="Failed to verify signing certificate, consider upgrading `sigstore` if a newer "
+            f"version is available: {store_ctx_error}"
+        )
```
woodruffw, line 144: LGTM. I think we'll want to think a bit more about how we want to render exceptions in general, but we can hold off on that.

woodruffw (approved):

PR #114, opened 2022-06-03
Add a requirements file w/ hashes, and additional instructions for installing

di:
This PR adds a requirements file that users can optionally use when installing to enable hash-checking mode.

It also turns on Dependabot, to keep this file up to date.


woodruffw (approved):
LGTM!

One thought: I wonder whether we'll need some additional CI work to apply updates when `pyproject.toml` is changed manually?

For example, I can imagine a case where we bump `cryptography` and our `requirements.txt` is out-of-date for approximately a day, causing stale (or broken) installs.

di:
Great point. I raised https://github.com/sigstore/sigstore-python/issues/115 to address this.

PR #116, opened 2022-06-04
workflows/release: upload assets, dogfood verification

### LINKED ISSUE #106: Dogfood verification
I'm not a complete beginner and I still have trouble understanding what exactly I can actually verify with sigstore-python at the moment... It would be helpful if there was an example in the documentation of verifying e.g. that a sigstore-python release was actually built by the github action of the project.

The status at the moment seems to be:
* certificate and signature could easily be published along the artifacts (since `sign --output-*` options were just added) but currently they are not published
* There is a somewhat undocumented `verify --cert-email` option that seems to be what I want, just slightly badly named: it looks like it would allow me to check that SAN contains `URI:https://github.com/sigstore/sigstore-python/.github/workflows/release.yml@refs/tags/$VERSION`
* I imagine in addition to SAN I should verify that the signing key is a github key? Is there a process for this or should the verifying script just hardcode the keyid? Should there be an option in `verify` for this?

I'd really like to use sigstore-python for this exact purpose so I'm willing to do some work here but I think I need a bit of guidance at least: Is my understanding of how to verify that build artifacts were built by a specific github action correct? Am I missing anything with regards to my last bullet point: how do I verify that the cert actually comes from GitHub?

woodruffw:
Needs testing with a prerelease.

Closes #106.

cc @jku

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,9 +5,12 @@ on:
 
 name: release
 
-# Needed to access the workflow's OIDC identity.
 permissions:
+  # Needed to access the workflow's OIDC identity.
   id-token: write
+
+  # Needed to upload release assets.
+  content: write
 
 jobs:
   pypi:
@@ -27,13 +30,21 @@ jobs:
     - name: sign
       run: |
         # we smoke-test sigstore by installing each of the distributions
-        # we've built in a fresh environment and using each to sign for itself,
-        # using the ambient OIDC identity
+        # we've built in a fresh environment and using each to sign and
+        # verify for itself, using the ambient OIDC identity
         for dist in dist/*; do
           python -m venv smoketest-env
 
           ./smoketest-env/bin/python -m pip install "${dist}"
-          ./smoketest-env/bin/python -m sigstore sign "${dist}"
+
+          ./smoketest-env/bin/python -m \
+            sigstore sign --output-signature --output-certificate "${dist}"
+
+          ./smoketest-env/bin/python -m \
+            sigstore verify \
+            --cert "${dist}.crt" --signature "${dist}.sig" \
+            --cert-oidc-issuer https://token.actions.githubusercontent.com \
+            "${dist}"
 
           rm -rf smoketest-env
         done
@@ -43,3 +54,11 @@ jobs:
       with:
         user: __token__
         password: ${{ secrets.PYPI_TOKEN }}
+
+    - name: upload artifacts to github
+      # Confusingly, this action also supports updating releases, not
+      # just creating them. This is what we want here, since we've manually
+      # created the release that triggered the action.
+      uses: softprops/action-gh-release@v1
+      with:
+        files: dist/*


.github/workflows/release.yml
```diff
@@ -27,13 +30,21 @@ jobs:
...
           ./smoketest-env/bin/python -m pip install "${dist}"
-          ./smoketest-env/bin/python -m sigstore sign "${dist}"
+
+          ./smoketest-env/bin/python -m \
+            sigstore sign --output-signature --output-certificate "${dist}"
+
+          ./smoketest-env/bin/python -m \
+            sigstore verify \
+            --cert "${dist}.crt" --signature "${dist}.sig" \
+            --cert-oidc-issuer https://token.actions.githubusercontent.com \
```
di, line 46: IMO we need some way to verify the workflow (org, repo, name) here as well. Maybe not a blocker for this but something to think about

woodruffw: Yep. These are the X.509v3 extensions we can verify against:

```
        X509v3 extensions:
            X509v3 Key Usage: critical
                Digital Signature
            X509v3 Extended Key Usage: 
                Code Signing
            X509v3 Basic Constraints: critical
                CA:FALSE
            X509v3 Subject Key Identifier: 
                D2:A5:AA:A9:BF:8D:EA:35:7D:C5:6B:16:A8:45:68:A2:9A:15:2C:FB
            X509v3 Authority Key Identifier: 
                58:C0:1E:5F:91:45:A5:66:A9:7A:CC:90:A1:93:22:D0:2A:C5:C5:FA
            X509v3 Subject Alternative Name: critical
                URI:https://github.com/sigstore/sigstore-python/.github/workflows/release.yml@refs/tags/v0.4.2
            1.3.6.1.4.1.57264.1.1: 
                https://token.actions.githubusercontent.com
            1.3.6.1.4.1.57264.1.6: 
                refs/tags/v0.4.2
            1.3.6.1.4.1.57264.1.3: 
                dd2be1d19eeb0fbffbc36afe27bba7ba03fe97cf
            1.3.6.1.4.1.57264.1.5: 
                sigstore/sigstore-python
            1.3.6.1.4.1.57264.1.2: 
                release
            1.3.6.1.4.1.57264.1.4: 
                release
```

So we could additionally do the SAN here (since we have that via the misleading `--cert-email` option), as well as the commit hash and symbolic `ref`.

woodruffw:
Example of us now attaching our signing results to each release:

Screen Shot 2022-06-04 at 2 43 26 PM

We run `sigstore verify` with these results as part of the release workflow as well.

.github/workflows/release.yml
jku: Yes, I would agree checking SAN would make sense. My security property assumptions (that I have not yet checked against github docs) are 
* GitHub asserts the workflow referred to by SAN was used for this build
* GitHub asserts it only gives OIDC tokens to trusted GitHub build systems

This still leaves some details (like the choice of using _mutable_ tag in SAN) that could be improved by checking the other extensions if/when that is possible, but I think SAN+OIDC provider is a very good starting point.

jku (commented):
Looks good. 

As a future consideration, if GitHub alone has six seemingly custom extensions that they think are useful, it looks like a somewhat generic `--cert-extension <id> <value>` option could maybe make sense -- so that callers can decide what they expect the certificate to contain without requiring specific support from sigmake sense -- so that callers can decide what they expect the certificate to contain without requiring specific support from sigstore-python?
store-python?

sigstore/__init__.py
```diff
@@ -16,4 +16,4 @@
 The `sigstore` APIs.
 """
 
-__version__ = "0.5.0"
+__version__ = "0.5.1rc2"
```
di, line 19:
```suggestion
__version__ = "0.5.0"
```

.github/workflows/release.yml
```diff
@@ -26,14 +29,28 @@ jobs:
...
+
+          ./smoketest-env/bin/python -m \
+            sigstore sign "${dist}" \
+            --output-signature smoketest-artifacts/"${dist_base}.sig" \
+            --output-certificate smoketest-artifacts/"${dist_base}.crt"
+
+          ./smoketest-env/bin/python -m \
+            sigstore verify "${dist}" \
+            --cert "smoketest-artifacts/${dist_base}.crt" \
+            --signature "smoketest-artifacts/${dist_base}.sig" \
```
di, line 52: Can we add comments about why these are going into a different directory than the default?

di (dismissed):

.github/workflows/release.yml
woodruffw: Yep, can do now.

di (approved):

PR #117, opened 2022-06-05
sigstore/verify: add more sigstore extension OIDs

woodruffw:
We don't currently use these, but we probably will in the future.


haydentherapper:
High level question - What should we be doing to make sure this list stays up to date as more OIDs are added?

@di, this can probably be solved with client conformance tests

di:
@haydentherapper I'm guessing you mean when we expose the ability to verify on these values? Yes, IMO this would be captured by the conformance tests.

Issue #25, opened 2022-04-10
Retrieve CTFE signing key via TUF


tetsuo-cpp:
At the moment, we've decided to check in the CTFE public key and use it to verify SCTs. The way this should really work is that we should check in the root key (`root.json`) and use it to download the CTFE key via TUF.

di:
Now that we're including an intermediate key, this should also include that as well.

wxjdsr:
Hi, we are a small team (@wxjdsr, @vcharapa) that is working with professor Santiago (@SantiagoTorres) on fixing this issue.

woodruffw:
Cool! Please let me or one of the other maintainers know if there's something we can do to help.

joshuagl:
👋 [python-tuf](https://github.com/theupdateframework/python-tuf) maintainer and [sigstore/root-signing](https://github.com/sigstore/root-signing) contributor here. 

I posted a [message on the #python Slack channel](https://sigstore.slack.com/archives/C024FPJKC6L/p1662561658491689) a few weeks back about some WIP changes I have to implement the certificate update workflow in python-sigstore. You can see my WIP branch here: https://github.com/joshuagl/sigstore-python/commits/joshuagl/tuf

Apologies for not posting about that work here, I didn't find this issue when I tried searching for something about the TUF workflow 🤦 

It would be cool to collaborate on this, or have your work build on top of mine, @wxjdsr and @vcharapa?

wxjdsr:
> 👋 [python-tuf](https://github.com/theupdateframework/python-tuf) maintainer and [sigstore/root-signing](https://github.com/sigstore/root-signing) contributor here.
> 
> I posted a [message on the #python Slack channel](https://sigstore.slack.com/archives/C024FPJKC6L/p1662561658491689) a few weeks back about some WIP changes I have to implement the certificate update workflow in python-sigstore. You can see my WIP branch here: https://github.com/joshuagl/sigstore-python/commits/joshuagl/tuf
> 
> Apologies for not posting about that work here, I didn't find this issue when I tried searching for something about the TUF workflow 🤦
> 
> It would be cool to collaborate on this, or have your work build on top of mine, @wxjdsr and @vcharapa?

Cool. I think we can collaborate on this.

joshuagl:
[...]

PR #121, opened 2022-06-07
workflows: add staging-tests

### LINKED ISSUE #120: Run scheduled tests against rekor/fulcio staging
**Description**

In order to catch regressions, we should run scheduled tests against rekor/fulcio staging.

We could optionally run these in CI as well, but allow failures.

woodruffw:
I need to configure this to not block normal PRs, but this is the basic idea.

Closes #120.


.github/workflows/staging-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+  # Needed to access the workflow's OIDC identity.
+  id-token: write
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  schedule:
+    - cron: '0 12 * * *'
```
di, line 13: Kind of inclined to make this more frequent, thoughts?

di (approved):

.github/workflows/staging-tests.yml
woodruffw: Makes sense to me. Every 4/6 hours maybe?

woodruffw:
Verification is failing, maybe we're missing a URL change somewhere:

```
Transparency log entry created at index: 227
Signature written to file README.md.sig
Certificate written to file README.md.crt
Failed to verify signing certificate
FAIL: README.md
This may be a result of an outdated `sigstore` installation.
Consider upgrading with:
    python -m pip install --upgrade sigstore
Additional context:
[30, 0, 'authority and subject key identifier mismatch']
```

Will investigate in a bit.

di:
Could also be a legitimate failure against staging 😉

woodruffw:
Oh, it's probably because we don't have the staging root and intermediate certs checked in, only the CTFE staging pubkey. I can add those.

woodruffw:
> Could also be a legitimate failure against staging 😉

Or that!

.github/workflows/staging-tests.yml
woodruffw: Set to 4 hours.

woodruffw:
This should be good for a review. I had to do a pretty significant refactor to plumb all of the needed keys and certificates into their needed locations without creating a lot of spaghetti.

To summarize:

* The top-level `sign` and `verify` APIs are now `Signer.sign` and `Verifier.verify`, where the corresponding classes are state managers for the relevant keys and certificates (and Rekor/Fulcio clients)
* We now have a general "production" and "staging" API pattern, where `{Signer,Verifier,FulcioClient,RekorClient}` all support `production()` and `staging()` factory methods that produce ready-to-use instances. This cuts down significantly on the amount of plumbing needed.

sigstore/_cli.py
```diff
@@ -312,11 +320,19 @@ def _verify(
...
+    elif rekor_url == DEFAULT_REKOR_URL:
+        verifier = Verifier.production()
+    else:
+        # TODO: We need CLI flags that allow the user to figure the Fulcio cert chain
+        # for verification.
+        click.echo(
+            "Custom Rekor and Fulcio configuration for verification isn't fully supported yet!",
+            err=True,
+        )
+        raise click.Abort
```
woodruffw, line 335: Flagging for review: we don't really have a clean way to do this yet, since it would require yet another flag (one that can be repeated) to add custom certificates for the custom Fulcio instance's chain.

I believe `cosign` sidesteps this entirely by relying heavily on TUF, so this will sort of resolve itself when #25 is resolved. Even then, however, the user will need a way to specify custom TUF root metadata so that we can bootstrap the TUF client, and the TUF repository will need to contain the various files at consistent locations/targets.

sigstore/_cli.py
```diff
@@ -151,6 +143,22 @@ def main() -> None:
...
+    help="A PEM-encoded public key for the CT log (conflicts with --staging)",
+)
+@click.option(
+    "rekor_root_pubkey",
+    "--rekor-root-pubkey",
+    metavar="FILE",
+    type=click.File("rb"),
+    default=BytesIO(DEFAULT_REKOR_ROOT_PUBKEY),
+    help="A PEM-encoded root public key for Rekor itself (conflicts with --staging)",
+)
```
woodruffw, line 161: Flagging for review: this is like https://github.com/sigstore/sigstore-python/pull/121/files#r891786806 below, except I went ahead and added the needed flag since it's less complicated. But I can remove it and have it return an error for the time being, instead, if desired.

di (approved):

Issue #124, opened 2022-06-08
Define a "verification configuration bundle"


di:
**Description**

I'd like to explore a format for a "verification configuration bundle". This would be:

* a single human-readable file (probably JSON)
* with a specific extension (like `.sigstore`)
* that includes the verification materials for one or more signatures for one or more artifacts (likely keyed by SHA256 hash)
* as well as configuration for additional extension verification

So for #106, this could be something like:

```json
{
  "config": {
    "cert-oidc-issuer": "https://token.actions.githubusercontent.com",
    ...
    "<some other extension>": "<some other value>"
  },
  "artifacts": {
    "<sha256 hash>": {
      "bundles": [
        {
          "sig": "<signature>"
          "crt": "<base64 encoded certificate>"
        },
        ...
        <potentially additional signatures>
     ]
    },
    "<another sha256 hash>": {
        <same as above>
    }
  }
}  
```

Which would be used like:

```
# For multiple artifacts:
$ python -m sigstore verify <something>.sigstore artifacts_to_verify/*

# Same bundle, but works for a single artifact:
$ python -m sigstore verify <something>.sigstore artifacts_to_verify/artifact.ext
```

Some open questions:
* What should the name of this file be, if it signs multiple artifacts?
* Should the "additional verification" be across all artifacts, or specific to individual artifacts? Or just potentially override-able?
* How would a user create/edit this file with a `sigstore` client?

jku:
This does have some advantages over my suggestion of github specific verify command:
* avoids being service specific (apart from requiring specific config being supported)
* allows more complex configuration in case it's needed. As an example (not saying this is a good design for SAN, just making an example):
```json
      "config": {
          "cert-oidc-issuer": "https://token.actions.githubusercontent.com",
          "subject-alternative-name": {
            "type": "URI",
            "value": "https://github.com/sigstore/sigstore-python/.github/workflows/release.yml@refs/tags/v0.5.1"
          },
        },
```

The only complaint I have with formats like these is that in practice as a user I'm still going to produce this file _at time of use_ -- in the github case I would want to fill in at least git tag and git hash at invocation time. That said a simple script (possibly that one actually should be github specific) that produces this json would not be hard to make.

di:
Thinking about this a bit more, one bundle that handles multiple files might be a bit unwieldy/unexpected for folks that expect verification materials to be per-artifact. There's also a lot of complexity about when it's OK to add to this file or when it should be regenerated.

[...]

PR #125, opened 2022-06-08
CLI: Add `--output` and `--force`

### LINKED ISSUE #122: `--output-signature` and `--output-certificate` flags are greedy
**Description**

When trying to use the `--output-*` flags with default values, invocation fails because they try to consume the filename of the artifact to be signed:

```
$ python -m sigstore sign --output-signature --output-certificate hi.txt
Usage: python -m sigstore sign [OPTIONS] FILE [FILE ...]
Try 'python -m sigstore sign --help' for help.

Error: Missing argument 'FILE [FILE ...]'.
```

### LINKED ISSUE #123: Warn when verification materials will be overwritten
**Description**

Before the signing process is invoked, we should warn (or perhaps refuse to sign / output verification materials) if the sign will overwrite existing files with the new verification materials.

We could possibly, additionally, provide a flag that permits the overwrite.

woodruffw:
This addresses the greedy output flags problem by giving them mandatory values, and adding a new option (`--output`) that toggles whether default output filenames are used instead.

It also addresses #123 by adding a new `--force` option that, if not passed, will refuse to continue with `sigstore sign` until pre-existing signing artifacts are moved or deleted from the overwriting path(s).

More generally, this PR refactors the entire `sigstore` CLI to use Python's default `argparse` instead of `click`. The primarily reason for this is to avoid `click`'s ambiguities around greedy value captures (https://github.com/pallets/click/issues/2307), but it'll also help us move towards a configuration model as part of #124.

Fixes #122.
Fixes #123.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,15 +45,16 @@ Top-level:
 
 <!-- @begin-sigstore-help@ -->
 ```
-Usage: sigstore [OPTIONS] COMMAND [ARGS]...
+usage: sigstore [-h] [-V] {sign,verify} ...
 
-Options:
-  --version  Show the version and exit.
-  --help     Show this message and exit.
+a tool for signing and verifying Python package distributions
 
-Commands:
-  sign
-  verify
+positional arguments:
+  {sign,verify}
+
+options:
+  -h, --help     show this help message and exit
+  -V, --version  show program's version number and exit
 ```
 <!-- @end-sigstore-help@ -->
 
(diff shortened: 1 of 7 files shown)


woodruffw:
Marking as a draft, since I still need to add some default CLI state.

sigstore/_cli.py
```diff
@@ -47,343 +41,360 @@
...
+            extant = None
+            if sig and sig.exists():
+                extant = sig
+            elif cert and cert.exists():
+                extant = cert
+
+            if extant:
+                args._parser.error(
+                    f"Refusing to overwrite pre-existing outputs without --force: {extant}"
+                )
```
di, line 296: It might be a little annoying that this only tells the user about one of possibly two files that would be extant. I think this should output _all_ extant files at once.

README.md
```diff
@@ -45,78 +45,115 @@ Top-level:
...
+                        files ({input}.sig and {input}.crt) (default: False)
+  --output-signature FILE
+                        Write a single signature to the given file; conflicts
+                        with --output and does not work with multiple input
+                        files (default: None)
+  --output-certificate FILE
+                        Write a single certificate to the given file;
+                        conflicts with --output and does not work with
+                        multiple input files (default: None)
+  --force               Overwrite preexisting signature and certificate
```
di, line 103: "force" seems a little too strong here on second thought, makes me think of force-pushing. Perhaps "overwrite" would be better?

sigstore/_cli.py
woodruffw: Yeah, agreed, will update.

di (approved):

Issue #126, opened 2022-06-09
Are there any example python packages in PyPI that have been already signed using sigstore?


yunhua-deng:
**Question**

Trying play `sigstore verify` but don't know where to start. Providing example PyPI packages would help newbies of sigstore.

jku:
the sigstore-python releases themselves are signed (by the build system). Copying an example from my slack comment:
```
VERSION=0.5.1
FNAME=sigstore-${VERSION}-py3-none-any.whl
BASEURL=https://github.com/sigstore/sigstore-python/releases/download/v${VERSION}

# Download artifact, certificate and signature
wget ${BASEURL}/${FNAME}{,.crt,.sig}

# Verify that artifact was built on GitHub
python -m sigstore verify ${FNAME} \
    --cert ${FNAME}.crt \
    --signature ${FNAME}.sig \
    --cert-oidc-issuer https://token.actions.githubusercontent.com
```

You could download the wheel file from pypi and verify that one (instead of the one from github releases page as my example does).

Note that while verify "works as intended" (it verifies that the certificates key has been used to produce the signature that matches the artifact), there is very little mechanism in place so far to make sure the signature actually means something -- as an example, was it signed by a project maintainer, a build system or someone completely unrelated? If the artifact was signed by a maintainer whose identity (email) you know, you could already use `verify --cert-email` but currently the published python-sigstore artifacts are only signed by the build system.

#106 and #124 have some related future work for verifying build system signatures. 

**Maybe it would make sense for the maintainers to manually upload their signature to the release artifacts, to provide an example that users could verify?**

woodruffw:
Thanks for the request @yunhua-deng! As @jku mentioned, we're currently self-signing `sigstore` releases with itself, but I'm not sure if any other packages are using it yet.

This has made me realize that we don't currently have that many (any?) examples of use in the README. I'll add some now.

> Maybe it would make sense for the maintainers to manually upload their signature to the release artifacts, to provide an example that users could verify?

I don't see any technical reason not to do this, although the naming scheme is going to be slightly annoying (since there are already signatures/certificates uploaded). In the mean time, I could attach a signature and certificate pair to this issue for the latest release.

woodruffw:
#129 should add some clarifying examples, at least for `sigstore sign`. We'll have some subsequent PRs to document expected workflow for `sigstore verify`, and some CLI changes to make `sigstore verify` more ergonomic.

yunhua-deng:
[...]

PR #129, opened 2022-06-09
README: add signing examples

woodruffw:
This should help with questions of the sort raised in #126.


woodruffw:
My plan is to add verification examples as well, once `sigstore verify` supports multiple verification inputs (which I'll do in a separate PR).

di (commented):
LGTM, my only note is we should use a simple filename like `foo.txt` instead of `foo.tar.gz` to reduce confusion.

We should also use `python -m sigstore` invocations everywhere.

di (approved):

PR #130, opened 2022-06-09
`sigstore verify`: support multiple inputs

### LINKED ISSUE #127: Infer signature and certificate paths from `sigstore verify`'s input
Right now, `sigstore verify` needs to be used like this:

```bash
sigstore verify --certificate foo.crt --signature foo.sig foo
```

This has a handful of downsides:

* It violates one of the Unix tool idioms, which is that flags should never be mandatory
* It fundamentally constrains the CLI to verifying one file at a time, since `--certificate` and `--signature` can only be passed once (in theory they could be made plural, but this results in poor UX where a user must either accept quadratic verification times *or* ensure that their flag order is exactly right)

One potential solution to this is to do the converse of what we've done with `sigstore sign`, and make `--certificate` and `--signature` optional. When absent, multiple inputs will be allowed, with the following lookup scheme:

```
input.ext -> [input.ext.crt, input.ext.sig]
```

This makes the happy path _much_ cleaner:

```bash
# automatically discovers foo.tar.gz.{crt,sig}
sigstore verify foo.tar.gz

# automatically discovers .crt and .sig for each of {foo,bar}.tar.gz
sigstore verify foo.tar.gz bar.tar.gz
```

Thoughts @di?

woodruffw:
Closes #127.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,8 +218,8 @@ namely the Fulcio's supported identity providers and the claims expected within 
 
 ### Verifying against a signature and certificate
 
-By default, `sigstore verify` will attempt to find a `foo.txt.sig` and `foo.txt.crt` adjacent
-to the file being verified:
+By default, `sigstore verify` will attempt to find a `<filename>.sig` and `<filename>.crt` in the
+same directory as the file being verified:
 
 ```console
 # looks for foo.txt.sig and foo.txt.crt
@@ -237,7 +237,10 @@ If your signature and certificate are at different paths, you can specify them
 explicitly (but only for one file at a time):
 
 ```console
-$ python -m sigstore verify --certificate foo.crt --signature foo.sig foo.txt
+$ python -m sigstore verify \
+    --certificate some/other/path/foo.crt \
+    --signature some/other/path/foo.sig \
+    foo.txt
 ```
 
 ### Extended verification against OpenID Connect claims
--- a/README.md
+++ b/README.md
@@ -216,6 +216,55 @@ $ python -m sigstore sign --identity-token YOUR-LONG-JWT-HERE foo.txt
 Note that passing a custom identity token does not circumvent Fulcio's requirements,
 namely the Fulcio's supported identity providers and the claims expected within the token.
 
+### Verifying against a signature and certificate
+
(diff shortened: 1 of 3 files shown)


woodruffw:
I'll also use this PR to update the README.

sigstore/_cli.py
```diff
@@ -362,6 +367,38 @@ def _sign(args: argparse.Namespace) -> None:
...
+
+        missing = []
+        if not sig.is_file():
+            missing.append(str(sig))
+        if not cert.is_file():
+            missing.append(str(cert))
+
+        if missing:
+            args._parser.error(
+                f"Missing requirements for {(file)}: {', '.join(missing)}"
```
di, line 397:
```suggestion
                f"Missing verification materials for {(file)}: {', '.join(missing)}"
```

sigstore/_cli.py
```diff
@@ -362,6 +367,38 @@ def _sign(args: argparse.Namespace) -> None:
 
 
 def _verify(args: argparse.Namespace) -> None:
+    # Fail if `--certificate` or `--signature` is specified and we have more than one input.
+    if (args.certificate or args.signature) and len(args.files) > 1:
+        args._parser.error(
+            "--certificate and --signature can only be used with a single input"
```
di, line 373:
```suggestion
            "--certificate and --signature can only be used with a single input file"
```

di (dismissed):

di (approved):

README.md
```diff
@@ -215,6 +216,55 @@ $ python -m sigstore sign --identity-token YOUR-LONG-JWT-HERE foo.txt
 Note that passing a custom identity token does not circumvent Fulcio's requirements,
 namely the Fulcio's supported identity providers and the claims expected within the token.
 
+### Verifying against a signature and certificate
+
+By default, `sigstore verify` will attempt to find a `foo.txt.sig` and `foo.txt.crt` adjacent
+to the file being verified:
```
di, line 222:
```suggestion
By default, `sigstore verify` will attempt to find a `<filename>.sig` and `<filename>.crt` in the
same directory as the file being verified:
```

tetsuo-cpp (approved):
LGTM!

PR #140, opened 2022-06-22
oauth: use a context manager for the server's thread

### LINKED ISSUE #139: Refactor the OAuth flow
The current OAuth implementation has some idempotency issues (#136) -- it's easy to misuse, which results in hard to debug errors.

I think the right solution here is to push everything into a context manager and collapse the lifetime distinction between the local listening server and the OAuth request.

Something like this (sketched):

```py
with OAuthFlow(...) as flow:
    # flow contains state for the server, switches to OOB as necessary
    # ...
```

woodruffw:
~~WIP.~~

Closes #139.

### DIFF
--- a/sigstore/_internal/oidc/oauth.py
+++ b/sigstore/_internal/oidc/oauth.py
@@ -46,17 +46,38 @@ AUTH_SUCCESS_HTML = """
 """
 
 
-class RedirectHandler(http.server.BaseHTTPRequestHandler):
+class OAuthFlow:
+    def __init__(self, client_id: str, client_secret: str, issuer: Issuer):
+        self._client_id = client_id
+        self._client_secret = client_secret
+        self._issuer = issuer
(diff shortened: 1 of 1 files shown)


woodruffw:
This should be good to go.

Key changes:

* Adds two new classes: `OAuthFlow` and `OAuthSession`. The former manages the entire flow lifecycle (including local HTTP server startup and teardown); the latter contains all of the state needed to create and verify an OAuth handshake. 
* The new `OAuthSession` API is slightly more misuse-resistant: it's idempotent across calls, *and* includes a poison state that catches potential misuse (e.g. someone attempting to generate the same OAuth request URL multiple times).

woodruffw:
This PR also preserves the "forced OOB" functionality that was already present, but I'm starting to think that it should be a CLI flag instead of an environment variable. So I can include that in these changes, if desired.

di (approved):

PR #151, opened 2022-07-05
cli: deprecate `verify --cert`

### LINKED ISSUE #149: Deprecate the `--cert` alias in favor of `--certificate`
We currently support both, but the former is a legacy alias inherited from the `cosign` CLI.

It violates the symmetry of the rest of the CLI (`--output-signature + --signature` and `--output-certificate` + `--certificate`), so we should deprecate and eventually remove it (probably with 1.0?).

cc @di for thoughts.

woodruffw:
The long form (`--certificate`) is the intended option.

Closes #149.


woodruffw:
As a demo:

```console
$ sigstore verify --cert README.md.crt README.md
WARNING:sigstore._cli:--cert has been replaced with --certificate and will be removed in an upcoming stable release
```

tetsuo-cpp:
Does `cosign` still use `--cert`? I had a quick grep and it looks like it still does.

I'll leave the judgement call to you but I tend to think that it's more important that our CLI is compatible with `cosign` than to be symmetrical or consistent in general.

tetsuo-cpp (approved):

woodruffw:
> I'll leave the judgement call to you but I tend to think that it's more important that our CLI is compatible with `cosign` than to be symmetrical or consistent in general.

Yeah, that's why I wanted to get @di's opinion -- IMO we're divergent enough at this point that it makes sense to make changes that improve our UX, but this is admittedly a very small example of one of those changes 🙂

di:
I think we should just support both as the same option for now, with no deprecation warning, to ensure this remains intuitive for `cosign` users as well as make sense for new users.

woodruffw:
Very stale and not unanimous, so closing 🙂

PR #153, opened 2022-07-07
_cli: Create `--signature` and `--certificate` aliases for the `--output-{signature,certificate}` flags

### LINKED ISSUE #150: UX: Drop `--output-` prefix from `--output-signature` and `--output-certificate`
The prefix here is arguably redundant -- the `sign` operation that these options belongs to implies that they're outputs, and we also emphasize that in the `--help` text. Making them just `--signature` and `--certificate` would make the `sigstore` CLI even more symmetric.

Similar to #149, we should perform a deprecation here before actually breaking the CLI.

Brought up by @sethmlarson.

cc @di for thoughts.

tetsuo-cpp:
Closes #150


sigstore/_cli.py
```diff
@@ -257,20 +259,20 @@ def main() -> None:
...
 
-    # Fail if `--output-signature` or `--output-certificate` is specified
-    # *and* we have more than one input.
-    if (args.output_signature or args.output_certificate) and len(args.files) > 1:
+    # Fail if `--signature` or `--certificate` is specified *and* we have more
+    # than one input.
+    if (args.signature or args.certificate) and len(args.files) > 1:
         args._parser.error(
-            "Error: --output-signature and --output-certificate can't be used with "
-            "explicit outputs for multiple inputs; consider using --output",
```
tetsuo-cpp, line 273: This is one more reference to `--output` that I missed in #148.

tetsuo-cpp:
As per our discussion on #151, I haven't put a deprecation notice here.

woodruffw:
> As per our discussion on https://github.com/sigstore/sigstore-python/pull/151, I haven't put a deprecation notice here.

I think the reasoning in #151 doesn't apply here 🙂 -- these flags are our own inventions, not compatibility flags with `cosign`. As such I think we should mark them for deprecation, since we're not planning on supporting them in the stable release in the interest of `cosign` compatibility.

tetsuo-cpp:
> > As per our discussion on #151, I haven't put a deprecation notice here.
> 
> I think the reasoning in #151 doesn't apply here 🙂 -- these flags are our own inventions, not compatibility flags with `cosign`. As such I think we should mark them for deprecation, since we're not planning on supporting them in the stable release in the interest of `cosign` compatibility.

I believe these names still come from `cosign`. From `cosign sign-blob --help`:
```
      --output-certificate string                                                                write the certificate to FILE
      --output-signature string                                                                  write the signature to FILE
```

PR #159, opened 2022-07-18
Switch to Fulcio v2 API

haydentherapper:
The main differences are:
* RootCert -> TrustBundle now returns a list of chains
* SigningCertificate differentiates between the embedded and detached
  SCT, and it's no longer returned in a header.
* Support for gRPC, but this continues to use HTTP.

No changes necessary for the request construction, `CertificateSigningRequest` is still the name of the field on the request.

Signed-off-by: Hayden Blauzvern <hblauzvern@google.com>

### DIFF
--- a/sigstore/_internal/fulcio/client.py
+++ b/sigstore/_internal/fulcio/client.py
@@ -29,13 +29,11 @@ from enum import IntEnum
 from typing import List, Optional
 from urllib.parse import urljoin
 
-import pem
 import requests
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.x509 import (
     Certificate,
     CertificateSigningRequest,
-    ExtensionNotFound,
     PrecertificateSignedCertificateTimestamps,
     load_pem_x509_certificate,
 )
@@ -51,8 +49,8 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_FULCIO_URL = "https://fulcio.sigstore.dev"
 STAGING_FULCIO_URL = "https://fulcio.sigstage.dev"
-SIGNING_CERT_ENDPOINT = "/api/v1/signingCert"
-ROOT_CERT_ENDPOINT = "/api/v1/rootCert"
+SIGNING_CERT_ENDPOINT = "/api/v2/signingCert"
+TRUST_BUNDLE_ENDPOINT = "/api/v2/trustBundle"
 
 
 class SCTHashAlgorithm(IntEnum):
@@ -162,10 +160,10 @@ class FulcioCertificateSigningResponse:
 
 
 @dataclass(frozen=True)
-class FulcioRootResponse:
-    """Root certificate response"""
+class FulcioTrustBundleResponse:
+    """Trust bundle response, containing a list of certificate chains"""
 
-    root_cert: Certificate
+    trust_bundle: List[List[Certificate]]
 
 
 class FulcioClientError(Exception):
@@ -212,16 +210,29 @@ class FulcioSigningCert(Endpoint):
             except (AttributeError, KeyError):
                 raise FulcioClientError from http_error
 
+        if resp.json().get("signedCertificateEmbeddedSct"):
+            sct_embedded = True
+            try:
+                certificates = resp.json()["signedCertificateEmbeddedSct"]["chain"]["certificates"]
+            except KeyError:
+                raise FulcioClientError("Fulcio response missing certificate chain")
+        else:
+            sct_embedded = False
+            try:
+                certificates = resp.json()["signedCertificateDetachedSct"]["chain"]["certificates"]
+            except KeyError:
+                raise FulcioClientError("Fulcio response missing certificate chain")
+
         # Cryptography doesn't have chain verification/building built in
         # https://github.com/pyca/cryptography/issues/2381
-        try:
-            cert_pem, *chain_pems = pem.parse(resp.content)
-            cert = load_pem_x509_certificate(cert_pem.as_bytes())
-            chain = [load_pem_x509_certificate(c.as_bytes()) for c in chain_pems]
-        except ValueError:
-            raise FulcioClientError(f"Did not find a cert in Fulcio response: {resp}")
+        if len(certificates) < 2:
+            raise FulcioClientError(
+                f"Certificate chain is too short: {len(certificates)} < 2"
+            )
+        cert = load_pem_x509_certificate(certificates[0].encode())
+        chain = [load_pem_x509_certificate(c.encode()) for c in certificates[1:]]
 
-        try:
+        if sct_embedded:
             # Try to retrieve the embedded SCTs within the cert.
             precert_scts_extension = cert.extensions.get_extension_for_class(
                 PrecertificateSignedCertificateTimestamps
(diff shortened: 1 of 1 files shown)


haydentherapper:
What's the best way to test this locally? I'm not familiar with the process for locally building and using packages.

di:
@haydentherapper Do you want to run tests/linting locally, or invoke the client with your changes?

haydentherapper:
Invoke the client with my changes, I’m working on getting things set up for local linting.

di:
You should be able to run:

```
$ python -m pip install -e .
```

to install the local source tree as an 'editable' install, meaning that invoking `python -m sigstore` anywhere will execute the files in your source tree every time.

sigstore/_internal/fulcio/client.py
```diff
@@ -212,16 +211,19 @@ def post(
             except (AttributeError, KeyError):
                 raise FulcioClientError from http_error
 
+        if resp.json().get("signedCertificateEmbeddedSct"):
+            embeddedSct = True
```
woodruffw, line 215: Tiny style nit, for consistency with the rest of the codebase (`sct_embedded` instead of `embedded_sct` to emphasize that it's a flag, not a variable holding an embedded SCT).

```suggestion
            sct_embedded = True
```

sigstore/_internal/fulcio/client.py
```diff
@@ -291,16 +293,24 @@ def _opaque16(value: bytes) -> bytes:
         return FulcioCertificateSigningResponse(cert, chain, sct, raw_sct)
 
 
-class FulcioRootCert(Endpoint):
-    def get(self) -> FulcioRootResponse:
-        """Get the root certificate"""
+class FulcioTrustBundle(Endpoint):
+    def get(self) -> FulcioTrustBundleResponse:
+        """Get the certificate chains from Fulcio"""
```
woodruffw, line 298: Maybe this?

```suggestion
        """Get the root certificate and its verification chain from Fulcio"""
```

woodruffw: Whoops, ignore that. I see that we have multiple chains below.

sigstore/_internal/fulcio/client.py
haydentherapper: Thanks, updated!

haydentherapper:
Thanks @di, that worked!

Tested with `python3 -m sigstore sign --identity-token <token> foo.txt`, it worked as expected. Also tested `trust_bundle` manually by calling it in the code - There's a bug we're fixing in Fulcio that is concatenating all certificates together, so `trust_bundle` only returns the first cert, but this should be corrected once the Fulcio fix is rolled out.

Not certain how to get the ambient signing test working.

sigstore/_internal/fulcio/client.py
```diff
@@ -212,16 +210,19 @@ def post(
...
+
         # Cryptography doesn't have chain verification/building built in
         # https://github.com/pyca/cryptography/issues/2381
-        try:
-            cert_pem, *chain_pems = pem.parse(resp.content)
-            cert = load_pem_x509_certificate(cert_pem.as_bytes())
-            chain = [load_pem_x509_certificate(c.as_bytes()) for c in chain_pems]
-        except ValueError:
-            raise FulcioClientError(f"Did not find a cert in Fulcio response: {resp}")
+        cert = load_pem_x509_certificate(certificates[0].encode())
```
tetsuo-cpp, line 222: Would be good to check for length here or at least catch `IndexError`.

sigstore/_internal/fulcio/client.py
```diff
@@ -262,14 +263,14 @@ def _opaque16(value: bytes) -> bytes:
...
-        except ExtensionNotFound:
+        else:
             # If we don't have any embedded SCTs, then we might be dealing
             # with a Fulcio instance that provides detached SCTs.
 
             # The SCT header is a base64-encoded payload, which in turn
             # is a JSON representation of the SignedCertificateTimestamp
             # in RFC 6962 (subsec. 3.2).
-            sct_b64 = resp.headers.get("SCT")
+            sct_b64 = resp.json()["signedCertificateDetachedSct"]["signedCertificateTimestamp"]
```
tetsuo-cpp, line 273: I think we need to use `get` in order for the check below to work. I expect that this will just raise a `KeyError`.

tetsuo-cpp:
Thanks @haydentherapper!

> Not certain how to get the ambient signing test working.

I can take a look at this if you'd like.

sigstore/_internal/fulcio/client.py
haydentherapper: Added a check for length.

haydentherapper:
@tetsuo-cpp Any help would be great! The error just says that ambient credentials aren't available, maybe it's just a flake? I see Dustin also filed a ticket to create a more actionable error message.

sigstore/_internal/fulcio/client.py
haydentherapper: Changed, thanks!

sigstore/_internal/fulcio/client.py
```diff
@@ -262,14 +267,14 @@ def _opaque16(value: bytes) -> bytes:
...
-        except ExtensionNotFound:
+        else:
             # If we don't have any embedded SCTs, then we might be dealing
             # with a Fulcio instance that provides detached SCTs.
 
             # The SCT header is a base64-encoded payload, which in turn
             # is a JSON representation of the SignedCertificateTimestamp
             # in RFC 6962 (subsec. 3.2).
-            sct_b64 = resp.headers.get("SCT")
+            sct_b64 = resp.json()["signedCertificateDetachedSct"].get("signedCertificateTimestamp")
```
di, line 277: We probably need to handle the case where `signedCertificateDetachedSct` is missing as well.

haydentherapper: Do we need to handle this case? That would be a malformed response from the server if it was missing.

haydentherapper: Same question for `resp.json()["signedCertificateEmbeddedSct"]["chain"]["certificates"]` - Do we need to guard against a malformed response, or are we ok with the key error that's returned?

di: IMO yes, we should wrap this whole thing in a `try`/`except KeyError` and produce a more meaningful error message than the stacktrace that would appear otherwise.

di:
> Maybe we should reserve this job just for when commits make their way into main and drop them from PR testing?

This seems fine to me. Either way, we should probably come up with some way to notify if this starts failing: https://github.com/sigstore/sigstore-python/issues/161

di:
Looks like all tests are passing now, we just need to resolve https://github.com/sigstore/sigstore-python/pull/159#discussion_r925047552

woodruffw:
> Looks like all tests are passing now, we just need to resolve [#159 (comment)](https://github.com/sigstore/sigstore-python/pull/159#discussion_r925047552)

I can take a stab at that in a moment.

haydentherapper:
Thanks y'all!

@woodruffw, I can resolve this too, unless you've already started it.

sigstore/_internal/fulcio/client.py
haydentherapper: Done! It's been awhile since I've written Python, forgot about the power of `try/except`.

sigstore/_internal/fulcio/client.py
```diff
@@ -262,15 +273,16 @@ def _opaque16(value: bytes) -> bytes:
...
 
             # The SCT header is a base64-encoded payload, which in turn
             # is a JSON representation of the SignedCertificateTimestamp
             # in RFC 6962 (subsec. 3.2).
-            sct_b64 = resp.headers.get("SCT")
-            if sct_b64 is None:
+            try:
+                sct_b64 = resp.json()["signedCertificateDetachedSct"]["signedCertificateTimestamp"]
+            except KeyError:
                 raise FulcioClientError("Fulcio response did not include a SCT header")
```
woodruffw, line 286: Sorry, one last nit: it looks like the detached SCT is no longer in the header, right? So this exception should probably be tweaked to something like...

```suggestion
                sct_b64 = resp.json()["signedCertificateDetachedSct"]["signedCertificateTimestamp"]
            except KeyError:
                raise FulcioClientError("Fulcio response did not include a detached SCT")
```

sigstore/_internal/fulcio/client.py
```diff
@@ -262,15 +273,16 @@ def _opaque16(value: bytes) -> bytes:
             raw_sct = _opaque16(_opaque16(raw_sct_list_bytes))
 
             sct = precert_scts_extension[0]
-        except ExtensionNotFound:
+        else:
             # If we don't have any embedded SCTs, then we might be dealing
             # with a Fulcio instance that provides detached SCTs.
 
             # The SCT header is a base64-encoded payload, which in turn
```
woodruffw, line 280:
```suggestion
            # The detached SCT is a base64-encoded payload, which in turn
```

sigstore/_internal/fulcio/client.py
haydentherapper: Yep, good catch! Fixed

woodruffw (approved):
LGTM!

PR #164, opened 2022-07-20
workflows/staging-tests: open an issue on failure

### LINKED ISSUE #161: Notify on CI failure
**Description**

We have some tests that run on a schedule to ensure the client continues to work against Sigstore's staging instance, but we have no way to be notified if they start failing, and we'll likely move these to only run on main/cron and not on PRs (ref: https://github.com/sigstore/sigstore-python/pull/159).

We should have some way to notify maintainers when these jobs start failing, perhaps by automatically opening an issue on this repository, or email notifications.

woodruffw:
Closes #161.

### DIFF
--- a/.github/workflows/staging-tests.yml
+++ b/.github/workflows/staging-tests.yml
@@ -24,26 +24,25 @@ jobs:
       - uses: actions/setup-python@7f80679172b057fc5e90d70d197929d454754a5a
 
       - name: staging tests
-        run: exit 1
-        # run: |
-        #   # This is similar to the "smoketest" that we run during the
-        #   # release workflow, except that we run against Sigstore's
-        #   # staging instances instead.
-        #   # We also don't bother to build distributions.
+        run: |
+          # This is similar to the "smoketest" that we run during the
+          # release workflow, except that we run against Sigstore's
+          # staging instances instead.
+          # We also don't bother to build distributions.
 
-        #   python -m venv staging-env
+          python -m venv staging-env
 
-        #   ./staging-env/bin/python -m pip install .
+          ./staging-env/bin/python -m pip install .
 
-        #   # Our signing target is not important here, so we just sign
-        #   # the README in the repository.
-        #   ./staging-env/bin/python -m sigstore sign --staging README.md
+          # Our signing target is not important here, so we just sign
+          # the README in the repository.
+          ./staging-env/bin/python -m sigstore sign --staging README.md
 
-        #   # Verification also requires a different Rekor instance, so we
-        #   # also test it.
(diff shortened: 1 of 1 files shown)


.github/workflows/staging-tests.yml
```diff
@@ -37,3 +42,32 @@ jobs:
...
+
+      - name: open an issue if the staging tests fail
+        if: failure()
+        uses: peter-evans/create-issue-from-file@v4
+        with:
+          title: "[CI] Integration failure: staging instance"
+          # created in the previous step
+          content-filepath: /tmp/staging-instance-issue.md
+          labels: bug,component:cicd,component:tests
+          assignees: woodruffw,di,tetsuo-cpp
```
woodruffw, line 73: Flagging: I put the three of us as assignees, but maybe someone from Fulcio or Rekor also makes sense?

woodruffw:
One thing: this will create a new issue each time the staging tests fail, which could happen several times (on the `cron` schedule) before a fix is implemented and applied. I'm not sure if that's worth addressing now.

.github/workflows/staging-tests.yml
di: I think this is fine for us to triage.

di:
> One thing: this will create a new issue each time the staging tests fail, which could happen several times (on the `cron` schedule) before a fix is implemented and applied. I'm not sure if that's worth addressing now.

Might get annoying if staging is broken for a long time, but I figure we can cross that bridge when we get to it.

.github/workflows/staging-tests.yml
```diff
@@ -37,3 +42,32 @@ jobs:
...
+
+          This suggests one of three conditions:
+
+          * A backwards-incompatible change in a Sigstore component;
+          * A regression in \`sigstore python\`;
+          * A transient error.
+
+          The full CI failure can be found here:
+
+          $GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID
```
di, line 62: Nice!

di (approved):

jku:
> > One thing: this will create a new issue each time the staging tests fail, which could happen several times (on the `cron` schedule) before a fix is implemented and applied. I'm not sure if that's worth addressing now.
> 
> Might get annoying if staging is broken for a long time, but I figure we can cross that bridge when we get to it.

when you do, a rough bridge is here: https://github.com/theupdateframework/python-tuf/blob/develop/.github/workflows/specification-version.yml :wink:

woodruffw:
Nice! Yeah, avoiding a third party action where we can do so conveniently seems like an easy win to me.

PR #169, opened 2022-07-21
Makefile: Make `lint` exit non-zero with unstaged changes

### LINKED ISSUE #168: CI: Reformatting from Black does not break `lint` workflow
A change recently went in that did not conform to Black's formatting rules. Our `lint` workflow did not catch this for whatever reason. I'm guessing that `make lint` does not exit with a non-zero exit code now.

tetsuo-cpp:
Closes #168

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,8 +47,9 @@ lint:
 		isort $(ALL_PY_SRCS) && \
 		flake8 $(ALL_PY_SRCS) && \
 		mypy $(PY_MODULE) && \
-		bandit -c pyproject.toml -r $(PY_MODULE) && \
-	        git diff --exit-code
+		bandit -c pyproject.toml -r $(PY_MODULE)
+
+	  git diff --exit-code
 
 .PHONY: test
 test:
--- a/Makefile
+++ b/Makefile
@@ -47,10 +47,8 @@ lint:
 		isort $(ALL_PY_SRCS) && \
 		flake8 $(ALL_PY_SRCS) && \
 		mypy $(PY_MODULE) && \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -47,10 +47,8 @@ lint:
 		isort $(ALL_PY_SRCS) && \
 		flake8 $(ALL_PY_SRCS) && \
 		mypy $(PY_MODULE) && \
-		bandit -c pyproject.toml -r $(PY_MODULE)
-# TODO: Reenable interrogate
-#		interrogate -v -c pyproject.toml . && \
```
tetsuo-cpp, line 52: I don't think we need to leave this here. We have an issue to track it already.

tetsuo-cpp:
When we merge #167 in, the `lint` workflow should start passing again.

woodruffw (approved):

Makefile
```diff
@@ -48,9 +48,8 @@ lint:
 		flake8 $(ALL_PY_SRCS) && \
 		mypy $(PY_MODULE) && \
 		bandit -c pyproject.toml -r $(PY_MODULE)
-# TODO: Reenable interrogate
-#		interrogate -v -c pyproject.toml . && \
-#	        git diff --exit-code
+
+	  git diff --exit-code
```
di, line 52: I'm not a fan of this because it makes linting fail when attempting to lint unstaged changes. I think we should be running `black --check` instead.

woodruffw: Ah, yep. Okay, I'll do a follow-on to drop this and use `black --check` instead.

di: (And maybe add a `make reformat` command that actually runs the autoformatters we use)

PR #170, opened 2022-07-21
oidc, _cli: Add more detailed error messages when ambient credential detection fails in GitHub Actions

### LINKED ISSUE #160: More detailed error messages when ambient credential detection fails
**Description**

In the CI for https://github.com/sigstore/sigstore-python/pull/159, ambient credential detection is failing. There are a few reasons why this might be happening, but the error message just indicates that detection fails. Ideally, we would add more detail about which specific things are missing that are required for credential detection.

tetsuo-cpp:
Closes #160

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,23 +89,3 @@ jobs:
             <( \
               make run ARGS="verify --help" \
             )
-      - name: staging tests
-        run: |
-          # This is similar to the "smoketest" that we run during the
-          # release workflow, except that we run against Sigstore's
-          # staging instances instead.
-          # We also don't bother to build distributions.
-
-          python -m venv staging-env
-
-          ./staging-env/bin/python -m pip install .
-
-          # Our signing target is not important here, so we just sign
-          # the README in the repository.
-          ./staging-env/bin/python -m sigstore sign --staging README.md
-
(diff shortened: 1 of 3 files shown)


tetsuo-cpp:
From [c941d87](https://github.com/sigstore/sigstore-python/pull/170/commits/c941d874e67c26b196d996a8336271e6de616dda):

```
Insufficient permissions for GitHub Actions workflow.

The most common reason for this is due to incorrect
configuration of the top-level `permissions` setting of the
workflow YAML file. It should be configured like so:

    permissions:
      id-token: write

Relevant documentation here:

    https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#adding-permissions-settings

Another possible reason is that the workflow run has been
triggered by a PR from a forked repository. PRs from forked
repositories typically cannot be granted write access.

Relevant documentation here:

    https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

Additional context:

GitHub: missing or insufficient OIDC token permissions, the ACTIONS_ID_TOKEN_REQUEST_TOKEN environment variable was unset

Error: Process completed with exit code 1.
```

woodruffw (approved):
LGTM! Great detail.

PR #171, opened 2022-07-21
Tweak linting target

woodruffw:
This follows the same pattern as Warehouse, and splits `make lint` into two separate targets:

* `make lint`: Run all linters, exiting with nonzero if any fail
* `make reformat`: Run the subset of linters that can automatically rewrite the source

Closes #169.


tetsuo-cpp (approved):
LGTM!

woodruffw:
Ah, maybe I should stop enabling auto-merge, since @di probably wanted to also give feedback on this 😅

di (commented):
👍

PR #174, opened 2022-07-28
_verify: Create a new `X509Store` for each `verify` call

### LINKED ISSUE #173: Verifying multiple files causes `certificate is not yet valid` from OpenSSL
I was verifying some artifacts and noticed that they fail to verify when I pass all of the artifacts to `sigstore-python` in one invocation. However, they verify fine when I pass them in separately in their own invocations.

So for example:
```
(env) tetsuo@Alexs-MacBook-Pro sigstore-python % sigstore verify --staging ~/Code/gh-action-sigstore-python/test/artifact.txt ~/Code/gh-action-sigstore-python/test/artifact1.txt
OK: /Users/tetsuo/Code/gh-action-sigstore-python/test/artifact.txt
FAIL: /Users/tetsuo/Code/gh-action-sigstore-python/test/artifact1.txt
Failure reason: Failed to verify signing certificate

This may be a result of an outdated `sigstore` installation.

Consider upgrading with:

    python -m pip install --upgrade sigstore

Additional context:

[9, 0, 'certificate is not yet valid']

(env) tetsuo@Alexs-MacBook-Pro sigstore-python % sigstore verify --staging ~/Code/gh-action-sigstore-python/test/artifact1.txt
OK: /Users/tetsuo/Code/gh-action-sigstore-python/test/artifact1.txt
```

tetsuo-cpp:
Closes #173

### DIFF
--- a/sigstore/_verify.py
+++ b/sigstore/_verify.py
@@ -93,12 +93,7 @@ class Verifier:
         establishing the trust chain for the signing certificate and signature.
         """
         self._rekor = rekor
-
-        self._store = X509Store()
-        for parent_cert_pem in fulcio_certificate_chain:
-            parent_cert = load_pem_x509_certificate(parent_cert_pem)
-            parent_cert_ossl = X509.from_cryptography(parent_cert)
-            self._store.add_cert(parent_cert_ossl)
+        self._fulcio_certificate_chain = fulcio_certificate_chain
 
     @classmethod
     def production(cls) -> Verifier:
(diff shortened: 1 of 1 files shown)


tetsuo-cpp:
I think there might be an underlying issue to report to either `pyopenssl` or `openssl` itself since, if the `set_time` call succeeds, users would understandably expect it to use the new time.

In the meantime, let's just get it working again with this patch.

woodruffw:
> I think there might be an underlying issue to report to either `pyopenssl` or `openssl` itself since, if the `set_time` call succeeds, users would understandably expect it to use the new time.

Yeah, that smells like a PyOpenSSL (or underlying OpenSSL) bug to me, or at least a documentation error.

I'd suggest filing it here: https://github.com/pyca/pyopenssl

woodruffw (dismissed):
LGTM!

sigstore/_verify.py
```diff
@@ -144,6 +139,15 @@ def verify(
         success.
         """
 
+        # NOTE: The `X509Store` object currently cannot have its time reset once the `set_time`
+        # method been called on it. To get around this, we construct a new one for every `verify`
+        # call.
+        store = X509Store()
+        for parent_cert_pem in self._fulcio_certificate_chain:
+            parent_cert = load_pem_x509_certificate(parent_cert_pem)
+            parent_cert_ossl = X509.from_cryptography(parent_cert)
```
tetsuo-cpp, line 148: Actually it'd probably be better to do this in the constructor and just store the `parent_cert_ossl`s for each chain member.

tetsuo-cpp:
Ok, I've removed the duplicate work. I'll let you take another look and pull the trigger on it.

woodruffw (approved):
LGTM structurally. We should probably get some tests in the CI that exercise this code path, so that we don't regress here in the future.

PR #178, opened 2022-07-30
✨ Enable Scorecard badge

azeemshaikh38:
#### Summary

Enable the beta version of scorecard-action:v2 which adds a Scorecard badge to the repo.

#### Release Note

NONE
#### Documentation
NONE


.github/workflows/scorecards-analysis.yml
```diff
@@ -20,15 +20,15 @@ jobs:
       security-events: write
       actions: read
       contents: read
-
+      id-token: write
```
woodruffw, line 23: Why does action need this permission?

azeemshaikh38: We use GitHub's OIDC token to sign the results which ensures integrity of the result displayed in the badge.

woodruffw: Would you mind adding a comment to that effect? I realize the other permissions aren't documented, but it'd be nice to (eventually) have them all documented so that future maintainers understand the security scope of the workflow.

azeemshaikh38: Done.

woodruffw:
LGTM structurally, although I have a (weak) preference for waiting for v2 to be stabilized and officially released before we merge it here. 

cc @di however; if you're content as-is then I have no objections 🙂

.github/workflows/scorecards-analysis.yml
woodruffw: Thanks! I'll admit that I'm still a little murky on what this means -- does it mean that the Scorecard server is verifying the OIDC-minted JWT to confirm that the results are actually being created from a particular run of this workflow?

azeemshaikh38: Yep, that's correct.

azeemshaikh38:
> LGTM structurally, although I have a (weak) preference for waiting for v2 to be stabilized and officially released before we merge it here.
> 
> cc @di however; if you're content as-is then I have no objections 🙂

If it helps, this version is already installed and running on [cosign](https://github.com/sigstore/cosign/blob/72c42246eecb4cfcb35e2dda9a2debbe5138ac8e/.github/workflows/scorecard_action.yml#L31), [rekor](https://github.com/sigstore/rekor/blob/0140a44cc063ab3f08884f685dd36efadd719570/.github/workflows/scorecard_action.yml#L31) and [fulcio](https://github.com/sigstore/fulcio/blob/0b23eb66b00690fbd8469c82e8bb397077b7421e/.github/workflows/scorecard_action.yml#L31) :)

di (approved):

.github/workflows/scorecards-analysis.yml
woodruffw: Thanks for clarifying!

di:
Looks like https://api.securityscorecards.dev/projects/github.com/sigstore/sigstore-python is 404? Is that expected?

woodruffw:
> Looks like https://api.securityscorecards.dev/projects/github.com/sigstore/sigstore-python is 404? Is that expected?

I think it was waiting for the action to finish; it returns JSON for me now.

PR #180, opened 2022-08-01
test: add an "online" marker for online tests

### LINKED ISSUE #177: Mark tests as "online"
Similarly to what we've done for `pip-audit`'s tests: we should mark tests that require network activity with `pytest.mark.online`, so that a user can opt-out of them in situations where networked tests are undesired or impractical.

woodruffw:
~~I might be missing a couple, but this puts the scaffolding in place.~~

Edit: The CI now runs the tests with the networking namespace disabled, ensuring that the `pytest.mark.online` markers are complete.

Closes #177.


woodruffw:
N.B.: This is mostly for developer experience, since both signing and verifying in the context of Sigstore are fundamentally online operations.

.github/workflows/ci.yml
```diff
@@ -27,6 +27,8 @@ jobs:
         run: make dev
       - name: test
         run: make test
+      - name: test (offline)
+        run: sudo unshare -n make test TEST_ARGS="--skip-online"
```
tetsuo-cpp, line 31: What's with the use of `sudo` and `unshare`? I think this deserves a comment.

tetsuo-cpp (dismissed):
LGTM!

.github/workflows/ci.yml
woodruffw: Yep, thanks for calling this out -- `unshare` is used to "un-share" the default networking namespace to the target program, meaning that it can't make network connections. In turn, that means that we exercise the correctness of our `pytest.mark.online` markers, since any tests that are missing markers would fail.

I'll summarize that in a comment! I think I can also work around the need for `sudo`.

woodruffw: [8cd6330](https://github.com/sigstore/sigstore-python/pull/180/commits/8cd633011f6a59e9b248be05aa98d5669b10b100)

tetsuo-cpp: Oh I see. Thanks for that.

tetsuo-cpp (approved):

PR #192, opened 2022-09-03
Skip curl to download requirements.txt

### LINKED ISSUE #191: README uses curl unnecessarily
**Description**

pip can download the file automatically, so the curl call is not needed. Unless you trust curl more than requests, which I guess is not entirely unreasonable :)

uranusjr:
#### Summary
pip can download the file automatically, so the curl call is not needed. Unless you trust curl more than requests, which I guess is not entirely unreasonable :)

Close #191.


di:
TIL! Thank you! Will update this on https://www.python.org/download/sigstore/ as well.

di (approved):

PR #199, opened 2022-09-07
treewide: upgrade to cryptography 38

### LINKED ISSUE #142: Revert workarounds to missing cryptography APIs
See #84 -- we have a variety of hacks in place due to missing APIs in a yet-unreleased version of `cryptography`.

We expect those changes to land with `cryptography` 38, at which point we can delete all of the hacks (including our dependency on `pyasn1`).

No current action needed/possible, since we don't currently have a release to upgrade to.

### LINKED ISSUE #195: [CI] Integration failure: staging instance
## Staging instance failure

A scheduled test against Sigstore's staging instance has failed.

This suggests one of three conditions:

* A backwards-incompatible change in a Sigstore component;
* A regression in `sigstore python`;
* A transient error.

The full CI failure can be found here:

https://github.com/sigstore/sigstore-python/actions/runs/3004260616

woodruffw:
This brings a couple of new (and much-needed) SCT APIs, allowing us to remove a variety of nasty unpacking hacks.

Closes #195.

Closes #142.

### DIFF
--- a/sigstore/_internal/rekor/client.py
+++ b/sigstore/_internal/rekor/client.py
@@ -203,9 +203,7 @@ class RekorClient:
         pubkey = serialization.load_pem_public_key(pubkey)
         if not isinstance(
             pubkey,
-            (
-                ec.EllipticCurvePublicKey,
-            ),
+            ec.EllipticCurvePublicKey,
         ):
             raise RekorClientError(f"Invalid public key type: {pubkey}")
         self._pubkey = pubkey
--- a/sigstore/_internal/rekor/client.py
+++ b/sigstore/_internal/rekor/client.py
@@ -27,6 +27,7 @@ from urllib.parse import urljoin
 
 import requests
 from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import ec, rsa
 from pydantic import BaseModel, Field, validator
 
 DEFAULT_REKOR_URL = "https://rekor.sigstore.dev"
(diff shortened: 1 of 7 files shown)


di:
This should also close #142, correct?

woodruffw:
> This should also close #142, correct?

Yep! I'll update the top comment to auto-close that as well.

woodruffw:
Typechecks are snarled because of some obsolete dependencies in typeshed: https://github.com/python/typeshed/issues/8699

di:
I'd be OK with temporarily ignoring these typecheck failures in the interest of getting this release out.

sigstore/_internal/rekor/client.py
```diff
@@ -199,8 +200,26 @@ def __init__(self, url: str, pubkey: bytes, ctfe_pubkey: bytes) -> None:
...
-        self._ctfe_pubkey = serialization.load_pem_public_key(ctfe_pubkey)
+        pubkey = serialization.load_pem_public_key(pubkey)
+        if not isinstance(
+            pubkey,
+            (
+                ec.EllipticCurvePublicKey,
+            ),
+        ):
+            raise RekorClientError(f"Invalid public key type: {pubkey}")
+        self._pubkey = pubkey
```
woodruffw, line 211: Flagging for review: this was already an implicit constraint (due to how we do the SET verification; it assumes an EC key in the cryptography APIs we use), but we didn't catch it until we switched to the upstream cryptography type hints.

I think this is _technically_ overly conservative of us, as the Rekor root pubkey could be RSA instead. But neither the production nor staging instance currently does that (and both of the keys we have checked in are EC).

di: There has actually been some discussion about changing the key to RSA if I understand correctly, cc @haydentherapper @asraa.

asraa: Hmmmm I don't think so. Maybe you are confusing it with me debating (no longer) this week on changing our TUF root management sign-off keys to RSA for better compatibility. But we are good to go on that front as well.

On my end you're fine here. The cosign code presumes Rekor to always be ecdsa too

di: Oh, I was mixing those up. Thanks!

haydentherapper: Right now, there are some assumptions in Rekor and Cosign that mandate that the key be EC. This is not true with CT however, which is why you've seen RSA and EC keys for staging/prod respectively. 

It'd be nice to support different types of keys in Rekor if possible, but this is not planned to be implemented any time soon. (I also prefer EC far more due to the reduced signature size).

woodruffw: Thanks @asraa and @haydentherapper!

PR #202, opened 2022-09-07
Staging workflow improvements

### LINKED ISSUE #200: Small improvements to smoketest/staging tests
**Description**

In #195 we found an example of the client failing against staging, and in #196, #197, #198, the issue was repeated.

Based on this, ideally:

- The issue body contains the exception or log of the failed job run
- The issue title contained the last line of the traceback
- The workflow could de-duplicate based on the title and prevent duplicate issues from being opened

This is probably not super critical given the infrequency in which this happens, but would be a nice to have.

woodruffw:
Closes #200.


.github/workflows/staging-tests.yml
```diff
@@ -12,7 +12,7 @@ on:
     branches:
       - main
   schedule:
-    - cron: '0 */4 * * *'
+    - cron: '0 */8 * * *'
```
woodruffw, line 15: NB: I changed this to run every 8 hours instead of 4, under the reasoning that 8 hours corresponds roughly to 3 different working times each day and gives us enough time between failures to actually institute a fix without too many duplicates piling up.

woodruffw:
I think we may be able to use https://github.com/peter-evans/find-comment to deduplicate automatically created issues. I'll look into that a bit more later.

woodruffw:
Hmm, so my cheeky idea won't work here: we can't search for a special sentinel comment, since `find-comment` can only look for a comment within a *specific* issue, not all issues. But we're trying to find the specific issue to begin with, so it's a chicken-and-egg problem.

di:
This is a pretty small diff, did something get lost?

woodruffw:
Nope; I pruned out the issue deduplication stuff because I couldn't figure out a reliable way to make it work without some pretty gnarly octokit scripting + a PAT.

di (approved):

PR #223, opened 2022-09-26
Add SLSA provenance generator to release; closes #222

### LINKED ISSUE #222: Integration with SLSA security framework
Hello, I'm working on behalf of Google and the [Open Source Security Foundation][https://openssf.org/] to help essential open-source projects improve their supply-chain security. I’m creating this issue to suggest the usage of the [SLSA](https://slsa.dev/) framework on your project, and expose my availability to create a PR with it.

In essence, SLSA aims to improve open source security by suggesting best practices regarding processes of artifact building and source code management, also creating [levels](https://slsa.dev/spec/v0.1/levels) of 0 to 4 that tells how closely the project is following the guidelines. Concerning the build process, SLSA highlights the importance of granting the integrity of binaries or artifacts, so that it's ensured that a binary's code really comes from the source code you’re relying on. This assurance relies on the usage of a [provenance](slsa.dev/provenance), which is verifiable and relatable to the artifact, describing where, when and how it was produced.

The sigstore-python already has solid build processes and code management, so for reaching SLSA 3 it misses only the generation of the build provenance on its releases. It would enable users to verify the integrity of the binaries and make the project even more reliable.

Would you be interested in a PR adding the provenance generation on your current build process?

diogoteles08:
#### Summary

This pull request adds a provenance generator to the Github Action build workflow. It enhances the security of the end-user by allowing them to verify the integrity of the artifact they are consuming.

Also, this PRs divides the build workflow into different jobs, each of them with their specific permissions.

IMPORTANT: The new build flow was not able to be fully tested because I did not have the credentials to publish on PYPI and also could not upload the release on Github because I did not added tags. But until this point of the build, everything worked out fine.

#### Documentation

I also changed the README to instruct the users on how to use the provenance to verify their artifacts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       id-token: write
+    outputs:
+      hashes: ${{ steps.hash.outputs.hashes }}
     steps:
       - uses: actions/checkout@d171c3b028d844f2bf14e9fdec0c58114451e4bf
 
@@ -49,6 +51,15 @@ jobs:
             rm -rf smoketest-env
           done
       
+      - name: Generate hashes for provenance
+        shell: bash
+        id: hash
+        run: |
+          # sha256sum generates sha256 hash for all artifacts.
+          # base64 -w0 encodes to base64 and outputs on a single line.
+          # sha256sum artifact1 artifact2 ... | base64 -w0
+          echo "::set-output name=hashes::$(sha256sum ./dist/* | base64 -w0)"
+      
       - name: Upload built packages
         uses: actions/upload-artifact@v3
         with:
@@ -63,8 +74,21 @@ jobs:
           path: smoketest-artifacts/
           if-no-files-found: warn
 
-  release-pypi:
+  generate-provenance:
     needs: [build]
+    name: Generate build provenance
+    permissions:
+      actions: read   # To read the workflow path.
+      id-token: write # To sign the provenance.
+      contents: write # To add assets to a release.
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
+    with:
+      attestation-name: sigstore-python-provenance.intoto.jsonl
+      base64-subjects: "${{ needs.build.outputs.hashes }}"
+      upload-assets: true
+
+  release-pypi:
+    needs: [build, generate-provenance]
     runs-on: ubuntu-latest
     permissions:
       id-token: write
@@ -82,7 +106,7 @@ jobs:
           packages_dir: built-packages/
 
   release-github:
-    needs: [build]
+    needs: [build, generate-provenance]
     runs-on: ubuntu-latest
     permissions:
       # Needed to upload release assets.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,6 @@
 name: Release
 
 on:
-  workflow_dispatch:
   release:
     types:
       - published
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,7 @@
 name: Release
 
 on:
+  workflow_dispatch:
   release:
     types:
       - published
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,78 +1,102 @@
+name: Release
+
 on:
   release:
     types:
       - published
 
-name: release
-
-permissions:
-  # Needed to access the workflow's OIDC identity.
-  id-token: write
-
-  # Needed to upload release assets.
-  contents: write
-
 jobs:
-  pypi:
-    name: Build, sign and publish release to PyPI
+  build:
+    name: Build and sign artifacts
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
     steps:
-    - uses: actions/checkout@d171c3b028d844f2bf14e9fdec0c58114451e4bf
+      - uses: actions/checkout@d171c3b028d844f2bf14e9fdec0c58114451e4bf
 
-    - uses: actions/setup-python@7f80679172b057fc5e90d70d197929d454754a5a
+      - uses: actions/setup-python@7f80679172b057fc5e90d70d197929d454754a5a
 
-    - name: deps
-      run: python -m pip install -U build
+      - name: deps
+        run: python -m pip install -U build
 
-    - name: build
-      run: python -m build
+      - name: build
+        run: python -m build
 
-    - name: sign
-      run: |
-        mkdir -p smoketest-artifacts
+      - name: sign
+        run: |
+          mkdir -p smoketest-artifacts
+          # we smoke-test sigstore by installing each of the distributions
+          # we've built in a fresh environment and using each to sign and
+          # verify for itself, using the ambient OIDC identity
+          for dist in dist/*; do
+            dist_base="$(basename "${dist}")"
+            python -m venv smoketest-env
+            ./smoketest-env/bin/python -m pip install "${dist}"
+            # NOTE: signing artifacts currently go in a separate directory,
+            # to avoid confusing the package uploader (which otherwise tries
+            # to upload them to PyPI and fails). Future versions of twine
+            # and the gh-action-pypi-publish action should support these artifacts.
+            ./smoketest-env/bin/python -m \
+              sigstore sign "${dist}" \
+              --output-signature smoketest-artifacts/"${dist_base}.sig" \
+              --output-certificate smoketest-artifacts/"${dist_base}.crt"
+            ./smoketest-env/bin/python -m \
+              sigstore verify "${dist}" \
+              --cert "smoketest-artifacts/${dist_base}.crt" \
+              --signature "smoketest-artifacts/${dist_base}.sig" \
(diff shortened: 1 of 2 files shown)


woodruffw (dismissed):
LGTM. I can try doing a pre-release to shake things out once this is merged!

diogoteles08:
Great! Should I squash commits for you?

di:
Not necessary, thanks!

.github/workflows/release.yml
```diff
@@ -1,78 +1,130 @@
...
+  generate-provenance:
+    needs: [build]
+    name: Generate build provenance
+    permissions:
+      actions: read   # To read the workflow path.
+      id-token: write # To sign the provenance.
+      contents: write # To add assets to a release.
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
+    with:
+      attestation-name: sigstore-python-provenance.intoto.jsonl
```
di, line 91: This seems arbitrary. Is it possible to make this based on the name of the artifact that the provenance is generated for? Also, how does this handle multiple artifacts? Are they all bundled in the same file?

E.g., could we make the provenance for `sigstore-0.6.5.tar.gz` be `sigstore-0.6.5.tar.gz.intoto.jsonl`, and so on?

README.md
```diff
@@ -305,6 +306,12 @@ Everyone interacting with this project is expected to follow the
 Should you discover any security issues, please refer to sigstore's [security
 process](https://github.com/sigstore/.github/blob/main/SECURITY.md).
 
+### SLSA Provenance
+This project emits a SLSA provenance on it's release! This enables you to verify the integrity
+of the downloaded artifacts and ensured that the binary's code really comes from this source code.
+To do so, please follow the instructions [here](https://github.com/urllib3/urllib3/blob/main/.github/workflows/publish.yml).
```
di, line 312: I don't think we should link to another project's workflow here, but I'm also not seeing instructions here on how to verify integrity.

woodruffw: Agreed. Is there an official (and human-first) doc we can link to instead, @diogoteles08?

.github/workflows/release.yml
diogoteles08: Based on my local testings, you can use that unique provenance file to verify both artifacts. That's why I used an "arbitrary" name

diogoteles08: We could indeed separate the provenances, but AFAIK for that I would need to create and output two different hashes and then call the slsa-github-generator twice, which seems a bit redundant

woodruffw: I'm personally okay with them being bundled together, although I also have a slight preference for the attestation filename having the version number in it somewhere (that way, someone who wants to pull down every attestation we've ever generated doesn't need to do the renaming on their own).

di: Perhaps this is better as a feature request upstream for the generator. I think in the Python world, it makes more sense to have provenance be per-artifact instead.

di: I'm fine with this for now if we can open an issue upstream as well.

README.md
diogoteles08: Ups, good catch! It was my mistake, for some reason I put the wrong URL. It was supposed to be this one: https://github.com/slsa-framework/slsa-github-generator#verification-of-provenance
If that makes sense now, please react and I'll commit the change

woodruffw: That looks good, thanks!

di: I think that's fine for now.

.github/workflows/release.yml
diogoteles08: I understand. I'll talk to someone that I know who works on SLSA project and ask if they have this feature in mind and possibly raise the issue.
In the meanwhile, I'll try to change the code to automatically fetch the release version and create the provenance with the name `provenance-sigstore-{version number}.intoto.jsonl`. Does it sound good?

diogoteles08: Just commited the proposed changes, please take a look

laurentsimon: Hey, @laurentsimon from the upstream SLSA generator project.

The SLSA specs currently encourage the use of a single `multiple.intoto.jsonl` provenance covering all artifacts (see https://github.com/in-toto/attestation/issues/107). Having a version is definitely good feedback for them, so I replied to the thread.

> I think in the Python world, it makes more sense to have provenance be per-artifact instead.

Can you explain why it makes more sense?

di: It's not the case here, but when Python builds include native code, builds have to be run across multiple platforms and architectures. In a hypothetical future, we might have a project that needs to generate provenance from multiple discrete SLSA-compatible builders, for multiple artifacts. So, the first problem is: what we name that provenance, so at verification time, users know what to use with a given artifact?

Then, if we then want to distribute provenance along with the artifact, we have the same problem: on a repository like PyPI, we have distribution files, and then additional artifacts that directly relate to those files. So for example, if I was signing something with PGP, I would upload `sigstore-0.6.5.tar.gz` and `sigstore-0.6.5.tar.gz.asc`. I think we want to follow a similar pattern here.

laurentsimon: Makes sense. @diogoteles08 when you get to a project like the one @di describes, let me know. Feel free to create a tracking issue on the slsa-github-generator repo as well now

.github/workflows/release.yml
```diff
@@ -1,78 +1,130 @@
...
+    with:
+      attestation-name: provenance-sigstore-${{ github.event.release.tag_name }}.intoto.jsonl
+      base64-subjects: "${{ needs.build.outputs.hashes }}"
+      upload-assets: true
+
+  release-pypi:
+    needs: [build, generate-provenance]
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
```
jku, line 99: is id-token needed here?

woodruffw: I don't think so -- @diogoteles08 could you remove it?

di: So, this will be necessary once we move to OIDC-based publishing (instead of stored secrets). But agreed it is not actually needed right now.

diogoteles08: Sure, good point! I didn't now if it was needed, did not know how to find out (tried to look into pypa/gh-action-pypi-publish but no success) nor test it, so just kept it. Thanks for the info!

laurentsimon: @di "OIDC-based publishing": do you mean to authenticate to the registry? Or to sign the package?

woodruffw: @laurentsimon The former -- PyPI will soon support an OIDC authentication flow, where trusted workflows will be allowed to publish for a particular package. But that feature hasn't actually landed yet 🙂

README.md
```diff
@@ -305,6 +306,12 @@ Everyone interacting with this project is expected to follow the
 Should you discover any security issues, please refer to sigstore's [security
 process](https://github.com/sigstore/.github/blob/main/SECURITY.md).
 
+### SLSA Provenance
+This project emits a SLSA provenance on it's release! This enables you to verify the integrity
```
woodruffw, line 310:
```suggestion
This project emits a SLSA provenance on its release! This enables you to verify the integrity
```

README.md
```diff
@@ -305,6 +306,12 @@ Everyone interacting with this project is expected to follow the
...
+of the downloaded artifacts and ensured that the binary's code really comes from this source code.
+To do so, please follow the instructions [here](https://github.com/slsa-framework/slsa-github-generator#verification-of-provenance).
```
woodruffw, line 312: Just to make it flow a little better.

```suggestion

To do so, please follow the instructions [here](https://github.com/slsa-framework/slsa-github-generator#verification-of-provenance).
```

woodruffw (commented):
Two small formatting/typo nits; I'll merge these using the GH web flow.

.github/workflows/release.yml
```diff
@@ -1,78 +1,129 @@
...
+          if-no-files-found: warn
+
+  generate-provenance:
+    needs: [build]
+    name: Generate build provenance
+    permissions:
+      actions: read   # To read the workflow path.
+      id-token: write # To sign the provenance.
+      contents: write # To add assets to a release.
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
```
woodruffw, line 89: Nonblocking: for consistency with the other actions in this workflow, we should probably use a hash rev instead of a symbolic (tag) rev.

woodruffw (approved):

.github/workflows/release.yml
di: Agreed, let's update this.

diogoteles08: I agree that it would maintain the consistency and also follow the [best practice](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) rules, but currently SLSA generator does not support hash pinning, although there already is an issue about that. More details [here](https://github.com/slsa-framework/slsa-github-generator#verification-of-provenance)

di (dismissed):

.github/workflows/release.yml
diogoteles08: However I think it would be a good idea to add a comment on code explaining this decision, because it's indeed contradictory and more people can question that. Will work on it

laurentsimon: +1, with a link to the issue

di (approved):

PR #238, opened 2022-10-06
rekor, verify: replace unstable API use

### LINKED ISSUE #237: Rekor Verification Reliability: Using /entries/retrieve rather than /index/retrieve
https://github.com/sigstore/sigstore-python/blob/67b0c0dda70859ce5d8912ad47c9a63ee2975b4f/sigstore/_verify.py#L251-L252

Hi all!

I was cleaning up an AI due to a post-mortem (https://github.com/sigstore/public-good-instance/pull/871, I'm aware this is not a public repository). This postmortem was on the issue that Rekor production was sharded, and originally Rekor was only searching for entries on the current active shard through the `/entries/retrieve` endpoint. This would mean that if a signature was uploaded before the instance sharded, it would not be found.

As a result, I had an AI to inform clients that they may need to be shard-aware. Users are able to upload the same entry to multiple shards, which means that when clients use `/entries/retrieve` to get an entry by the (sig, msg, cert) pair, that they may return multiple entries (on multiple shards). Obviously, the earliest entry should be used (because the integratedTime there is TRUE, later entries on later shards are "dupes" that someone uploaded later).

Then, I did some inspection into sigstore-python, and I found it was using `/index/retrieve` to get UUIDs for signature entries. Did you know that the endpoint is experimental and the index has no stability guarantee of verifiability?
https://github.com/sigstore/rekor/issues/1021

Ideally you should not be using this endpoint.

Because you have the certificate in that `_verify` function, you should be querying Rekor by the proposed entry with the (sig, msg, cert) pair through `/entries/retrieve`. This is a verifiable endpoint.

That being said, after sharding, clients have to be aware that they need to extract the earliest entry retrieved (the entry on the earliest shard, as described above).

I can provide more details later, or even help to craft a PR!

woodruffw:
~~WIP; needs cleanup.~~

Fixes #237.

### DIFF
--- a/sigstore/_internal/rekor/client.py
+++ b/sigstore/_internal/rekor/client.py
@@ -247,7 +247,7 @@ class RekorEntriesRetrieve(Endpoint):
         try:
             resp.raise_for_status()
         except requests.HTTPError as http_error:
-            if http_error.status_code == 404:
+            if http_error.response.status_code == 404:
                 return None
             raise RekorClientError(resp.json()) from http_error
 
@@ -269,7 +269,10 @@ class RekorEntriesRetrieve(Endpoint):
         # The response is a list of `{ uuid: LogEntry }` objects.
         # We select the oldest entry to actually return, since a malicious
         # actor could conceivably spam the log with newer duplicate entries.
-        entry = min(body[0].items(), key=lambda tup: tup[1]["integratedTime"])
+        entry = min(
+            body[0].items(),
+            key=lambda tup: tup[1]["integratedTime"],  # type: ignore[no-any-return]
+        )
 
         return RekorEntry.from_response(dict([entry]))
 
--- a/sigstore/_internal/rekor/client.py
+++ b/sigstore/_internal/rekor/client.py
@@ -215,6 +215,13 @@ class RekorEntriesRetrieve(Endpoint):
         sha256_artifact_hash: str,
         b64_cert: str,
     ) -> Optional[RekorEntry]:
+        """
+        Retrieves an extant Rekor entry, identified by its artifact signature,
+        artifact hash, and signing certificate.
+
+        Returns None if Rekor has no entry corresponding to the signing
+        materials.
+        """
         data = {
             "entries": [
                 {
--- a/sigstore/_internal/rekor/client.py
+++ b/sigstore/_internal/rekor/client.py
@@ -19,6 +19,7 @@ Client implementation for interacting with Rekor.
 from __future__ import annotations
 
 import json
+import logging
 from abc import ABC
 from dataclasses import dataclass
 from importlib import resources
@@ -29,6 +30,8 @@ import requests
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import ec, rsa
 from pydantic import BaseModel, Field, StrictInt, StrictStr, validator
+
+logger = logging.getLogger(__name__)
 
 DEFAULT_REKOR_URL = "https://rekor.sigstore.dev"
 STAGING_REKOR_URL = "https://rekor.sigstage.dev"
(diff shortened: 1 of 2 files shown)


sigstore/_internal/rekor/client.py
```diff
@@ -219,14 +194,57 @@ def post(
...
+            ]
+        }
+
+        resp: requests.Response = self.session.post(self.url, json=data)
+        try:
+            resp.raise_for_status()
+        except requests.HTTPError as http_error:
+            raise RekorClientError(resp.json()) from http_error
+
+        return RekorEntry.from_response(resp.json()[0])
```
asraa, line 246: I'm not totally confident that entries are received in earliest entry order...
I think in practice it actually goes active -> inactives
https://github.com/sigstore/rekor/blob/01c7b55405527d40de664d81696b5e78e7abbb0d/pkg/api/entries.go#L423

I can't decide if it make sense for this to return the array of RekorEntry's or to be handling the  parsing, ordering, and returning of a single internally. Maybe the latter is better for use-ability outside the package... but fairly complex for a client.

sigstore/_verify.py
```diff
@@ -249,53 +249,40 @@ def verify(
...
+            verify_merkle_inclusion(inclusion_proof, entry)
+        except InvalidInclusionProofError as inval_inclusion_proof:
+            return VerificationFailure(
+                reason=f"invalid Rekor inclusion proof: {inval_inclusion_proof}"
+            )
 
-            # 4) Verify the inclusion proof supplied by Rekor for this artifact
-            inclusion_proof = RekorInclusionProof.parse_obj(
-                entry.verification.get("inclusionProof")
+        # 5) Verify the Signed Entry Timestamp (SET) supplied by Rekor for this artifact
```
asraa, line 269: perfect!!!

sigstore/_verify.py
```diff
@@ -249,53 +249,40 @@ def verify(
         )
 
         # Retrieve the relevant Rekor entry to verify the inclusion proof and SET
-        uuids = self._rekor.index.retrieve.post(sha256_artifact_hash, pub_b64.decode())
+        entry = self._rekor.log.entries.retrieve.post(
```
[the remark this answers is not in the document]
woodruffw: Experimentally it looks like the endpoint returns `404` instead of an empty list, but I'll add the check for resiliency. 

(I'm also going to update the Python API here to return an `Optional[RekorEntry]`, since that's much more ergonomic than tracking an HTTP-sourced exception that could be either `404` or something else.)

sigstore/_internal/rekor/client.py
woodruffw: > I can't decide if it make sense for this to return the array of RekorEntry's or to be handling the parsing, ordering, and returning of a single internally. Maybe the latter is better for use-ability outside the package... but fairly complex for a client.

Yeah, I think for now I'll make the client do it internally (especially since these are all currently internal API details for `sigstore-python`).

Longer term it might make sense to make a separate Rekor API client that `sigstore-python` depends on, which in turn should isolate these concerns. But that's not a project for today 🙂

sigstore/_internal/rekor/client.py
```diff
@@ -219,14 +193,86 @@ def post(
...
+            return None
+        elif len(body) > 1:
+            raise RekorClientError(
+                f"expected exactly one response for entry query, but got {len(body)}"
+            )
+
+        # The response is a list of `{ uuid: LogEntry }` objects.
+        # We select the oldest entry to actually return, since a malicious
+        # actor could conceivably spam the log with newer duplicate entries.
+        entry = min(body[0].items(), key=lambda tup: tup[1]["integratedTime"])
```
woodruffw, line 272: This is a little ugly, but it accomplishes the filter without surfacing it beyond the client API. Thoughts @asraa @di?

sigstore/_verify.py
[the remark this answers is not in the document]
woodruffw: https://github.com/sigstore/sigstore-python/pull/238/files#diff-45c49b779739a3e08051f2c6afe622eb6b693d600dcc0ad9052ac30139a19a62R261-R267

asraa (approved):
Looks good to me!!

Side note: I love how nice and easy Python makes it to do min with a configured parameter.

sigstore/_internal/rekor/client.py
```diff
@@ -219,14 +193,89 @@ def post(
...
+                return None
+            raise RekorClientError(resp.json()) from http_error
+
+        # Rekor should return a 404 when there are no results rather than
+        # an empty list, but we check for the latter just in case.
+        #
+        # Similarly, we expect exactly one top-level response, since we made
+        # exactly one entry request. The inner entry set may have more than
+        # one result, but anything other than 1 at the top-level indicates
+        # a bug in Rekor.
```
di, line 260: I'm a little confused by this comment -- does this mean we're expecting a list at the top level with exactly one object in it? The top level response isn't a JSON object? "one top-level response" doesn't parse for me.

sigstore/_verify.py
```diff
@@ -240,62 +236,45 @@ def verify(
...
+            signature.decode(),
+            sha256_artifact_hash,
+            base64.b64encode(certificate).decode(),
+        )
+        if entry is None:
+            return VerificationFailure(
+                reason="Rekor has no entry for these verification materials"
```
di, line 247: While we're changing this, we probably could make this a bit more user friendly and explain what this means.

di (approved):
Overall LGTM, just some nits.

di:
Dang automerge!

woodruffw:
Ah crap, that was my fault. I can either revert or do a follow-up to address those.

sigstore/_internal/rekor/client.py
woodruffw: The top-level response is a JSON array, with each element corresponding to a "search result" for the query sent to the endpoint. We only send one query, so we only expect one response back.

(I agree the language is confusing; I can reword it to something close to what I just wrote.)

PR #241, opened 2022-10-07
_verify: make the failure reason more detailed when rekor lookup fails

woodruffw:

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -37,6 +37,7 @@ from sigstore._internal.rekor.client import DEFAULT_REKOR_URL, RekorClient
 from sigstore._sign import Signer
 from sigstore._verify import (
     CertificateVerificationFailure,
+    RekorEntryMissing,
     VerificationFailure,
     Verifier,
 )
@@ -423,11 +424,11 @@ def _verify(args: argparse.Namespace) -> None:
     for file, inputs in input_map.items():
         # Load the signing certificate
         logger.debug(f"Using certificate from: {inputs['cert']}")
-        certificate = inputs["cert"].read_bytes()
+        certificate = inputs["cert"].read_bytes().rstrip()
 
         # Load the signature
         logger.debug(f"Using signature from: {inputs['sig']}")
-        signature = inputs["sig"].read_bytes()
+        signature = inputs["sig"].read_bytes().rstrip()
 
         logger.debug(f"Verifying contents from: {file}")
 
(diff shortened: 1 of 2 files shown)


sigstore/_verify.py
```diff
@@ -244,7 +244,11 @@ def verify(
         )
         if entry is None:
             return VerificationFailure(
-                reason="Rekor has no entry for these verification materials"
+                reason=(
+                    "Rekor has no entry for these verification materials; "
+                    "this suggests that the signing certificate was not published "
+                    "to the log at the time the signature was created"
```
woodruffw, line 250: I'm curious to hear feedback on this -- Rekor missing the certificate is (IMO) the most likely reason for failure here, although it could also happen because the user searches the wrong tlog (e.g. staging instead of production) or if they signed multiple artifacts using the certificate but only uploaded signatures for a subset of them.

tetsuo-cpp (approved):

sigstore/_verify.py
woodruffw: Yeah, agreed. I think I'm going to do what we did in `pip-audit` here and specialize `VerificationFailure` a bit more, so that we can push some of this error handling into the CLI. That in turn will allow us to include the filename and tlog information, since we have more context there.

sigstore/_cli.py
```diff
@@ -423,11 +424,11 @@ def _verify(args: argparse.Namespace) -> None:
     for file, inputs in input_map.items():
         # Load the signing certificate
         logger.debug(f"Using certificate from: {inputs['cert']}")
-        certificate = inputs["cert"].read_bytes()
+        certificate = inputs["cert"].read_bytes().rstrip()
 
         # Load the signature
         logger.debug(f"Using signature from: {inputs['sig']}")
-        signature = inputs["sig"].read_bytes()
+        signature = inputs["sig"].read_bytes().rstrip()
```
woodruffw, line 431: N.B.: This wasn't causing failures, but I noticed that these two inputs can sometimes have trailing newlines in them (due to being text files).

sigstore/_cli.py
```diff
@@ -468,6 +469,33 @@ def _verify(args: argparse.Namespace) -> None:
...
+                # wasn't logged after creation or because the user requested the
+                # wrong Rekor instance (e.g., staging instead of production).
+                # The latter is significantly more likely, so we add
+                # some additional context to the output indicating it.
+                #
+                # NOTE: Even though the latter is more likely, it's still extremely
+                # unlikely that we'd hit this -- we should always fail with
+                # `CertificateVerificationFailure` instead, as the cert store should
+                # fail to validate due to a mismatch between the leaf and the trusted
+                # root + intermediates.
```
woodruffw, line 483: Flagging for review: the likelihood of different failure modes is speculation on my part, here. I figure it's much less likely for a sigstore client to forget to publish to the tlog than for a user to accidentally hit the wrong tlog.

However, as the `NOTE:` says, it would still be pretty difficult for a user to hit this *particular* error case, as there would have to be an earlier bug in cert chain validation. To actually trigger this for testing, I had to disable our cert chain validation.

sigstore/_verify.py
woodruffw: [967a587](https://github.com/sigstore/sigstore-python/pull/241/commits/967a58751e2e1ac88b6ab586f2ed4944da18ddf4) addresses most of this -- the artifact filename is emitted as part of the top-level failure message, and we now emit additional context (like the artifact hash and signature) that should help a user do a manual tlog lookup.

di (dismissed):

sigstore/_cli.py
asraa: I see: I just checked validation code, and I agree: in order for a user to hit this bug, it would somehow mean that the certificate pool was configured with the CORRECT root, but that the tlog was pointing to a different instance of the ecosystem.

A mismatch between the fulcio roots and tlog are unlikely I agree: both the incorrect tlog request & the certificate verification errors are pointing to errors in the client itself: the only time users would see this and it WOULDN'T be the client's fault is LIKELY a malicious signature event. 

IMO: It is worth a callout that this may actually be a failure like "The signing event could not be verified using the configured log.  This may be a result of connecting to the wrong Rekor instance..." like you say.

woodruffw: Sounds good; I'll tweak to cover that!

asraa: lgtm!

di (approved):

PR #244, opened 2022-10-11
_cli: add boolean envvar defaults

woodruffw:
This rounds out the envvar changes in #242.

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -75,6 +75,22 @@ def _boolify_env(envvar: str) -> bool:
     return val.lower() in {"true", "yes", "1"}
 
 
+def _add_shared_instance_options(group: argparse._ArgumentGroup) -> None:
+    group.add_argument(
+        "--staging",
+        action="store_true",
+        default=_boolify_env("SIGSTORE_STAGING"),
+        help="Use sigstore's staging instances, instead of the default production instances",
+    )
+    group.add_argument(
+        "--rekor-url",
+        metavar="URL",
+        type=str,
+        default=os.getenv("SIGSTORE_REKOR_URL", DEFAULT_REKOR_URL),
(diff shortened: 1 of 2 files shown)


sigstore/_cli.py
```diff
@@ -62,6 +62,19 @@ def __repr__(self) -> str:
...
+    """
+    An `argparse` helper for turning an environment variable into a boolean.
+
+    The only valid `True` states are (case-insensitive) "true", "yes", and "1".
+    Anything else, including the absence of the variable, is `False`.
+    """
+    val = os.getenv(envvar)
+    if val is None:
+        return False
+    return val.lower() in {"true", "yes", "1"}
```
di, line 75: Can we make this an implementation of https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool?

woodruffw: Yep, doing that now.

woodruffw: Done; the only deviation is for `None`, since we need that to default to `False` to prevent the absence of envvars from causing errors.

woodruffw: N.B.: This doesn't produce the prettiest user errors, since the `ValueError` isn't caught in the normal `argparse` handler:

```console
$ SIGSTORE_OVERWRITE=x sigstore sign README.md
    sys.exit(main())
  File "/Users/william/devel/sigstore-python/sigstore/_cli.py", line 284, in main
    parser = _parser()
  File "/Users/william/devel/sigstore-python/sigstore/_cli.py", line 190, in _parser
    default=_boolify_env("SIGSTORE_OVERWRITE"),
  File "/Users/william/devel/sigstore-python/sigstore/_cli.py", line 83, in _boolify_env
    raise ValueError(f"can't coerce '{val}' to a boolean")
ValueError: can't coerce 'x' to a boolean
```

di (approved):

PR #247, opened 2022-10-11
Offline Rekor bundle generation and verification

### LINKED ISSUE #52: Support verifying offline Rekor bundles
**Description**

As described in the [specification](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md), a Rekor response can be stored and verified without hitting the log. The client should:
* Verify the SET
* Verify that the signature over the blob matches what's in the payload's body
* Compare the integrated time against the certificate
* Compare hashes of the Rekor payload body to the provided blob

See [code in Cosign](https://github.com/sigstore/cosign/blob/b9152aa67931923be7da2cf59093df88b4dc4002/pkg/cosign/verify.go#L730) for more details.

### LINKED ISSUE #194: Support generating offline Rekor bundles
**Description**

Similar to `cosign`'s `--bundle` flag, we should support generating offline Rekor bundles as well:

E.g. from `cosign`:
```
--bundle string         write everything required to verify the blob to a FILE
```

This is likely a logical first step before #52.

woodruffw:
~~WIP.~~

Closes #52.

Closes #194.


sigstore/_internal/rekor/client.py
```diff
@@ -48,13 +49,47 @@
 
 @dataclass(frozen=True)
 class RekorEntry:
-    uuid: str
+    uuid: Optional[str]
```
woodruffw, line 52: cc @asraa: I had to make this optional, since offline Rekor bundles don't include their entry UUID. Just a heads up, since we talked a bit about the inclusion of this field before 🙂

woodruffw: (The other option here is to do some subclassing, i.e. an abstract `RekorEntry` and `OnlineEntry` and `OfflineEntry` as subclasses.)

woodruffw:
This is ready for an initial review, but it needs some more testing before merge (I haven't actually confirmed offline bundles work yet, only that the changes don't break online verification).

My plan was to implement #194 in this PR as well, to complete the whole picture (and round out testing). But I can do that separately and use a bundle generated by `cosign` instead, if we'd like to keep the diff small(er).

sigstore/_internal/rekor/client.py
asraa: I'd vote for optional! 

I suppose the UUID information should just be informational (e.g. the hash can be reconstructed from the entry content itself, and the tree ID should just be a hint on which tree the entry is)

README.md
```diff
@@ -146,8 +146,9 @@ Verifying:
 <!-- @begin-sigstore-verify-help@ -->
 ```
 usage: sigstore verify [-h] [--certificate FILE] [--signature FILE]
-                       [--cert-email EMAIL] [--cert-oidc-issuer URL]
-                       [--staging] [--rekor-url URL]
+                       [--bundle FILE] [--cert-email EMAIL]
```
haydentherapper, line 149: nit: We probably should name this something else like `rekor-bundle` to avoid confusion when the "sigstore bundle" spec is completed.

asraa:
> But I can do that separately and use a bundle generated by cosign instead, if we'd like to keep the diff small(er).

I may not necessarily totally us cosign's scheme... I'm not really a fan of the duplicated fields as I mentioned in https://github.com/sigstore/sigstore-python/issues/194#issuecomment-1240976109

But I hear the burden of implementing something twice. On the other hand, I think keeping it absolutely minimal to just the Rekor entry stored offline would allow building up to the "Sigstore bundle" (which is WIP design https://github.com/sigstore/cosign/pull/2204/ away from the sorta-faulty cosign bundle)

sigstore/_cli.py
```diff
@@ -245,6 +250,13 @@ def _parser() -> argparse.ArgumentParser:
         default=os.getenv("SIGSTORE_SIGNATURE"),
         help="The signature to verify against; not used with multiple inputs",
     )
+    input_options.add_argument(
+        "--bundle",
+        metavar="FILE",
+        type=Path,
+        default=os.getenv("SIGSTORE_BUNDLE"),
```
haydentherapper, line 257: Same thing here, we should refer to this as a REKOR_BUNDLE

woodruffw:
> But I hear the burden of implementing something twice. On the other hand, I think keeping it absolutely minimal to just the Rekor entry stored offline would allow building up to the "Sigstore bundle" (which is WIP design [sigstore/cosign#2204](https://github.com/sigstore/cosign/pull/2204) away from the sorta-faulty cosign bundle)

Makes sense. Yeah, this PR is limited to only the "Rekor" bundle, not the bigger thing that I understand (maybe incorrectly?) to be unstable/soft-deprecated on `cosign`'s side.

sigstore/_internal/rekor/client.py
```diff
@@ -71,10 +106,47 @@ def from_response(cls, dict_: Dict[str, Any]) -> RekorEntry:
...
+        """
+
+        payload = dict_["payload"]
+        return cls(
+            uuid=None,
+            body=payload["body"],
+            integrated_time=payload["body"],
+            log_id=payload["body"],
+            log_index=payload["body"],
+            inclusion_proof=None,
```
asraa, line 130: thx for the comment above explaining why this is optional!

sigstore/_cli.py
```diff
@@ -261,6 +273,12 @@ def _parser() -> argparse.ArgumentParser:
         default=os.getenv("SIGSTORE_CERT_OIDC_ISSUER"),
         help="The OIDC issuer URL to check for in the certificate's OIDC issuer extension",
     )
+    verification_options.add_argument(
+        "--offline",
+        action="store_true",
+        default=_boolify_env("SIGSTORE_OFFLINE"),
```
haydentherapper, line 279: When you integrate with TUF, will this flag be respected? TUF by default will make network calls to fetch the latest timestamp if expired. We might want to scope this to only Rekor.

sigstore/_verify.py
```diff
@@ -136,6 +137,10 @@ def verify(
 
         `expected_cert_oidc_issuer` is the expected OIDC Issuer Extension within `certificate`.
 
+        `offline_rekor_entry` is an optional offline `RekorEntry` to verify against. If supplied,
```
asraa, line 140: nit to add in comment: this only does SignedEntryTimestamp verification, but not inclusion proof verification. offline verification is slightly weaker and does not provide guarantees that the log has a consistent view

asraa (approved):

sigstore/_cli.py
woodruffw: Sounds good! We're using the `SIGSTORE_` prefix everywhere, so this will become `SIGSTORE_REKOR_BUNDLE` under the current naming scheme 😅

sigstore/_cli.py
asraa: Or make explicit a default exception for TUF: TUF really can't be done totally offline (or guaranteed offline, unless you want to accept some verification failure here that you might have expired TUF metadata if you can't have a network call.)

sigstore/_internal/merkle.py
```diff
@@ -91,10 +91,11 @@ def _hash_leaf(leaf: bytes) -> bytes:
     return hashlib.sha256(data).digest()
 
 
-def verify_merkle_inclusion(
-    inclusion_proof: RekorInclusionProof, entry: RekorEntry
-) -> None:
+def verify_merkle_inclusion(entry: RekorEntry) -> None:
```
haydentherapper, line 94: Not related to this PR, but something we need to add - We now return checkpoints (https://github.com/sigstore/rekor/blob/main/openapi.yaml#L608), signed root hashes. We need to add verification of these when verifying the inclusion proof, otherwise we're just verifying a list of hashes without any connection to the log.

haydentherapper: cc @asraa

asraa: This is only for online rekor entries! But yes, I was going to file an issue on this after reading this.

sigstore/_verify.py
woodruffw: 94db410 should elaborate on this!

sigstore/_internal/merkle.py
woodruffw: Please do, thanks!

woodruffw: #248.

woodruffw:
Sanity checking myself again: in its current form, this PR isn't sufficient, right? I also need to check the offline Rekor bundle's consistency against the certificate and signature, since someone could conceivably ask me to verify a completely unrelated (but valid) Rekor bundle.

asraa:
> I also need to check the offline Rekor bundle's consistency against the certificate and signature, since someone could conceivably ask me to verify a completely unrelated (but valid) Rekor bundle.

YES!

It would maybe look something like: reconstruct the rekor entry with the hashed_rekord entry, and compare hashes/fields

woodruffw:
> It would maybe look something like: reconstruct the rekor entry with the hashed_rekord entry, and compare hashes/fields

Cool, makes sense! Just making sure I understand: the `hashedrekord` here should have the same layout as the one we POST to Rekor's REST API, right? Plus canonicalization and base64 encoding, of course.

woodruffw:
> Correct. The canonicalization is a tricky point though. E.g. if a client passes in a DER encoded cert, will you canonicalize to the PEM? Rekor canonicalized to de-dupe those.

At least for sigstore-python's purposes, we only accept PEM-encoded certs. So, assuming that Rekor's body is always PEM-encoded, that part should be fine!

README.md
```diff
@@ -162,13 +163,17 @@ Verification inputs:
...
+                        multiple inputs (default: None)
 
 Extended verification options:
   --cert-email EMAIL    The email address to check for in the certificate's
                         Subject Alternative Name (default: None)
   --cert-oidc-issuer URL
                         The OIDC issuer URL to check for in the certificate's
                         OIDC issuer extension (default: None)
+  --rekor-offline       Perform offline Rekor verification using a bundle;
+                        implied by --bundle (default: False)
```
di, line 176:
```suggestion
                        implied by --rekor-bundle (default: False)
```
Right? TBH, I'm slightly concerned about too much rekor-specific stuff sneaking into the UI, especially for something that a user might want to use (bundles) that doesn't require knowledge of rekor specifically. Maybe this should just be `--bundle`?

sigstore/_cli.py
```diff
@@ -261,6 +273,12 @@ def _parser() -> argparse.ArgumentParser:
         default=os.getenv("SIGSTORE_CERT_OIDC_ISSUER"),
         help="The OIDC issuer URL to check for in the certificate's OIDC issuer extension",
     )
+    verification_options.add_argument(
+        "--rekor-offline",
+        action="store_true",
+        default=_boolify_env("SIGSTORE_REKOR_OFFLINE"),
+        help="Perform offline Rekor verification using a bundle; implied by --bundle",
```
di, line 280:
```suggestion
        help="Perform offline Rekor verification using a bundle; implied by --rekor-bundle",
```

sigstore/_cli.py
```diff
@@ -403,10 +421,12 @@ def _sign(args: argparse.Namespace) -> None:
 
 
 def _verify(args: argparse.Namespace) -> None:
-    # Fail if `--certificate` or `--signature` is specified and we have more than one input.
-    if (args.certificate or args.signature) and len(args.files) > 1:
+    # Fail if --certificate, --signature, or --bundle is specified and we have more than one input.
```
di, line 424:
```suggestion
    # Fail if --certificate, --signature, or --rekor-bundle is specified and we have more than one input.
```

sigstore/_cli.py
```diff
@@ -416,24 +436,31 @@ def _verify(args: argparse.Namespace) -> None:
...
         missing = []
         if not sig.is_file():
             missing.append(str(sig))
         if not cert.is_file():
             missing.append(str(cert))
+        if not bundle.is_file() and args.rekor_offline:
+            # NOTE: We only produce errors on missing bundle files
+            # if the user has explicitly requested offline-only verification.
+            # Otherwise, we fall back on online verification.
+            missing.append(str(bundle))
```
di, line 456: This seems like it'd be somewhat surprising behavior -- if the bundle file isn't valid, we'll silently attempt online verification and then potentially fail if that's unavailable? Can't the presence of `--rekor-bundle` imply that the user wanted offline verification of _only_ that bundle?

asraa: +1, this seems like a case where we want to alert the user that their argument was invalid.

We used to fallback on online verification when the bundle produced a failure in cosign, and then we decided it was not correct. Granted, here it's not even getting to that point.

README.md
woodruffw: I tend to agree, although I think @haydentherapper's point was that `--bundle` will eventually be the expected CLI option for the "sigstore bundle", whereas we're currently consuming the (confusingly named) "rekor-only bundle". So having it as `--rekor-bundle` will make the migration a little smoother for users, as they won't need to change their inputs for the same flag between sigstore-python versions.

That being said, maybe I could add a warning log when this flag is used, to emphasize to users that it'll eventually (soon?) be replaced by `--bundle`?

sigstore/_cli.py
```diff
@@ -403,10 +421,12 @@ def _sign(args: argparse.Namespace) -> None:
...
 def _verify(args: argparse.Namespace) -> None:
-    # Fail if `--certificate` or `--signature` is specified and we have more than one input.
-    if (args.certificate or args.signature) and len(args.files) > 1:
+    # Fail if --certificate, --signature, or --bundle is specified and we have more than one input.
+    if (args.certificate or args.signature or args.rekor_bundle) and len(
+        args.files
+    ) > 1:
         args._parser.error(
-            "--certificate and --signature can only be used with a single input file"
+            "--certificate, --signature, and --bundle can only be used with a single input file"
```
woodruffw, line 429:
```suggestion
            "--certificate, --signature, and --rekor-bundle can only be used with a single input file"
```

sigstore/_cli.py
woodruffw: Yeah, I agree that this is surprising as-is: we should treat `--rekor-bundle` as implying `--rekor-offline`. I'll tweak to that.

README.md
```diff
@@ -162,13 +163,17 @@ Verification inputs:
...
+                        multiple inputs (default: None)
 
 Extended verification options:
   --cert-email EMAIL    The email address to check for in the certificate's
                         Subject Alternative Name (default: None)
   --cert-oidc-issuer URL
                         The OIDC issuer URL to check for in the certificate's
                         OIDC issuer extension (default: None)
+  --rekor-offline       Perform offline Rekor verification using a bundle;
+                        implied by --rekor-bundle (default: False)
```
jku, line 176: Why is the default false? Wouldn't it make sense that in this case:
1. `--rekor-bundle` is not set 
2. but `FILE.bundle` is found

you would use offline verification -- this corresponds to automatically finding and using `FILE.sig` and `FILE.cert`.

Maybe the default should be rekor-offline=TRUE if there is a bundle (that is if `--rekor-bundle` is set or `FILE.bundle` is found). Otherwise rekor-offline should be FALSE. Regardless of the files existence, user should be able to override to either direction: offline or online)

sigstore/_internal/rekor/client.py
```diff
@@ -71,10 +106,47 @@ def from_response(cls, dict_: Dict[str, Any]) -> RekorEntry:
...
+        Creates a `RekorEntry` from an offline Rekor bundle.
+
+        See: <https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md#properties>
+        """
+
+        payload = dict_["payload"]
+        return cls(
+            uuid=None,
+            body=payload["body"],
+            integrated_time=payload["body"],
```
tetsuo-cpp, line 127: How come everything is being initialised with `payload["body"]`?

README.md
woodruffw: > you would use offline verification -- this corresponds to automatically finding and using FILE.sig and FILE.cert if they are not specified.

That's the current default! I can see how the flag's name is a little confusing: what it actually does is *force* offline Rekor verification, and fails if it's impossible. By default, offline Rekor verification is opportunistic. 

In other words, here's the current set of behaviors:

* `sigstore verify foo.txt`: look for `foo.txt.{crt,sig,bundle}` and use all three if they exist, otherwise fall back on online Rekor verification
* `sigstore verify --rekor-offline foo.txt`: look for `foo.txt.{crt,sig,bundle}` and fail unless all three exist

woodruffw: In other words, maybe the right flag name is `--force-rekor-offline` or similar, to emphasize that this doesn't affect the "opportunistic" behavior. CC @asraa and @di for thoughts.

sigstore/_internal/rekor/client.py
woodruffw: (This did make me realize that the `payload` key wasn't correctly capitalized, though, so thanks for that!)

tetsuo-cpp: What I mean is, shouldn't it read something like:
```python
integrated_time=payload["integratedTime"],
```

woodruffw: Oh yep, that's another copy-paste error 😅 

Thanks for catching, and sorry for the initial misunderstanding!

sigstore/_cli.py
woodruffw: https://github.com/sigstore/sigstore-python/pull/247/commits/34a1e4a4291c952c4bc6640aaa5dd1f754386cf2 accomplishes this!

woodruffw:
Just for tracking purposes, there are a few things this PR still needs:

- [ ] Offline Rekor bundle _generation_ (per #194), so that we can do round-trip testing
- [ ] Cross-checking the offline Rekor bundle against the certificate and artifact, to ensure we're verifying the correct tlog entry
- [ ] Anything else I missed?

README.md
jku: thanks -- I should have checked the code before complaining.

I think the name might be fine, it's just the docstring that could maybe be tweaked, maybe `...implied by --rekor-bundle or the existence of FILE.bundle` (although that is not super clear either)

haydentherapper:
I'd also check the Cosign security advisories, https://github.com/sigstore/cosign/security/advisories?state=published, since most of them involved bundle verification.

woodruffw:
Thanks for linking that! https://github.com/sigstore/cosign/security/advisories/GHSA-8gw7-4j42-w388 looks like the relevant one in terms of confusable Rekor verification.

README.md
woodruffw: Maybe I could change it from "perform" to "require", i.e.:

```
  --rekor-offline       Require offline Rekor verification using a bundle;
                        implied by --rekor-bundle (default: False)
```

That emphasizes that `--rekor-offline` forces bundle use, but doesn't imply that other modes won't do so opportunistically.

woodruffw:
Just for anyone playing along at home: offline Rekor bundle generation now works, and can dogfood itself:

```console
# generates README.md.{sig,crt,bundle}
sigstore sign README.md

sigstore verify --rekor-offline README.md
```

Question for @haydentherapper and @asraa: should I be using the `.bundle` suffix in this way? I can imagine that that suffix will eventually be used for the "full" Sigstore bundle, so maybe I should use `.rekor` or something similar to avoid an eventual conflict.

haydentherapper:
We said `.sigstore` for the bundle. I'd be fine with `.rekor`

woodruffw:
[64370f5](https://github.com/sigstore/sigstore-python/pull/247/commits/64370f5720ff547234ead7a22ad972d5e4cc12c0) should address the consistency requirements between the offline Rekor bundle and the other signing artifacts. I'll refactor it into an independent API, however, since that top level `verify` API function is rapidly ballooning...

woodruffw:
> We said `.sigstore` for the bundle. I'd be fine with `.rekor`

Done! We now emit `{input}.rekor` and attempt to load that file by default, for offline Rekor bundles.

sigstore/_verify.py
```diff
@@ -236,27 +246,91 @@ def verify(
...
-            signature.decode(),
-            sha256_artifact_hash,
-            base64.b64encode(certificate).decode(),
-        )
-        if entry is None:
-            return RekorEntryMissing(
-                signature=signature.decode(), sha256_artifact_hash=sha256_artifact_hash
+        entry: Optional[RekorEntry]
+        if offline_rekor_entry is not None:
+            # TODO(ww): This should all go in a separate API, probably under the
```
woodruffw, line 251: NTS: I put this here just so people can review it, but I'll factor it into a separate API as part of this PR.

README.md
di: The doc changes lgtm, maybe `--require-rekor-offline`?

di (approved):

README.md
woodruffw: Sounds good to me!

sigstore/_verify.py
```diff
@@ -136,6 +138,13 @@ def verify(
 
         `expected_cert_oidc_issuer` is the expected OIDC Issuer Extension within `certificate`.
 
+        `offline_rekor_entry` is an optional offline `RekorEntry` to verify against. If supplied,
+        verification will be done against this entry rather than the against the online
+        transparency log. Offline Rekor entries do not carry their Merkle inclusion
+        proofs, and as such are verified only against their Signed Entry Timestamps.
+        This is a slightly weaker verification verification mode, as it does not guarantee
+        the log's consistency.
```
haydentherapper, line 146: nit: Even with an online lookup for inclusion, there's no guarantee of consistency, unless a consistency proof is requested and verified against a previously persisted checkpoint/signed root hash. To be more precise, I'd say that "it does not demonstrate inclusion in the log" (I'm using demonstrate instead of guarantee since technically to guarantee it, you'd need witnessing too)

sigstore/_verify.py
```diff
@@ -154,17 +163,18 @@ def verify(
...
-        # 2) Verify that the signing certiticate belongs to the signer
-        # 3) Verify that the signature was signed by the public key in the signing certificate
-        #
-        # And optionally, if we're performing verification online:
-        #
-        # 4) Verify the inclusion proof supplied by Rekor for this artifact
-        # 5) Verify the Signed Entry Timestamp (SET) supplied by Rekor for this artifact
-        # 6) Verify that the signing certificate was valid at the time of signing by comparing the
-        #    expiry against the integrated timestamp
+        # 1) Verify that the signing certificate is signed by the root
```
haydentherapper, line 166: nit: certificate chain, not just a root certificate

sigstore/_verify.py
```diff
@@ -154,17 +163,18 @@ def verify(
...
+        #    certificate and that the signing certificate was valid at the time
+        #    of signing.
+        # 2) Verify that the signing certificate belongs to the signer.
+        # 3) Verify that the signature was signed by the public key in the
```
haydentherapper, line 170:
```suggestion
        # 3) Verify that the artifact signature was signed by the public key in the
```

haydentherapper: Just to be more precise, to differentiate between artifact sigs and cert sigs

sigstore/_verify.py
```diff
@@ -236,27 +246,91 @@ def verify(
...
-            base64.b64encode(certificate).decode(),
-        )
-        if entry is None:
-            return RekorEntryMissing(
-                signature=signature.decode(), sha256_artifact_hash=sha256_artifact_hash
+        entry: Optional[RekorEntry]
+        if offline_rekor_entry is not None:
+            # TODO(ww): This should all go in a separate API, probably under the
+            # RekorEntry class.
+            logger.debug("offline Rekor entry: checking consistency")
```
haydentherapper, line 253: nit: Remove debug statement? Not sure if you want to keep it or not (if so, change "consistency")

sigstore/_verify.py
```diff
@@ -236,27 +246,91 @@ def verify(
...
+            except Exception:
+                return VerificationFailure(
+                    reason="couldn't parse offline Rekor entry's body"
+                )
+
+            # The Rekor entry's body should be a hashedrekord object.
+            # TODO: This should use a real data model, ideally generated from
+            # Rekor's official JSON schema.
+            kind, version = entry_body.get("kind"), entry_body.get("apiVersion")
+            if kind != "hashedrekord" or version != "0.0.1":
```
haydentherapper, line 266: out of scope for this, is there a tracking issue for supporting other types?

sigstore/_verify.py
```diff
@@ -236,27 +246,91 @@ def verify(
...
+                    reason=(
+                        f"Rekor entry is of unsupported kind ('{kind}') or API "
+                        f"version ('{version}')"
+                    )
+                )
+
+            spec = entry_body["spec"]
+            expected_sig, expected_cert, expected_hash = (
+                spec["signature"]["content"],
+                load_pem_x509_certificate(
```
haydentherapper, line 277: This field can contain either a certificate, a certificate chain, or a public key, so we'll need to handle this. Not sure if there's other parts of the code that make this assumption also.

sigstore/_cli.py
```diff
@@ -416,24 +465,31 @@ def _verify(args: argparse.Namespace) -> None:
...
             cert = file.parent / f"{file.name}.crt"
+        if bundle is None:
+            bundle = file.parent / f"{file.name}.rekor"
 
         missing = []
         if not sig.is_file():
             missing.append(str(sig))
         if not cert.is_file():
             missing.append(str(cert))
+        if not bundle.is_file() and args.require_rekor_offline:
```
haydentherapper, line 481: Are you requiring that the cert and sig be passed in addition to the bundle? For example, with Cosign, if you give us a bundle, we'll extract the cert and signature automatically - https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/verify/verify_blob.go#L173

cc @asraa for more thoughts

sigstore/_verify.py
woodruffw: Yeah, bad word choice on my part (I'm checking consistency between the entry and the signing artifacts, not consistency with the log). I'll reword.

sigstore/_verify.py
woodruffw: I believe this is the only place we make that assumption, since the only other places we touch `hashedrekord` objects directly are during log entry submission and querying. 

In this case, it's okay for us to fail if we don't see a PEM-encoded certificate, right? This code only runs (currently) in the context of verification against a PEM-encoded certificate, so anything else would imply that the user has given us the wrong Rekor entry.

sigstore/_cli.py
woodruffw: Yep -- right now the only "bundle" we support is the "Rekor only" bundle, i.e. the one in the [signature spec](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md). We don't currently support the "cosign-style" bundle with embedded cert and signature (although we plan to support the "sigstore-style" one, once it's finalized).

TL;DR `sigstore-python` currently expects 2+ independent verification inputs for each `verify` invocation: a separate `.crt`, separate `.sig`, and optionally a separate `.rekor` containing the JSON encoded Rekor bundle.

sigstore/_verify.py
woodruffw: Actually, I'm going to postpone this to #250, since that refactor will include this functionality.

sigstore/_verify.py
woodruffw: (More generally, I think this demonstrates that I probably need to invert how I'm matching against the offline Rekor entry -- instead of unpacking the `body` and making sure it matches the cert/sig/etc, I should pack the latter into a `hashedrekord` payload and make sure it's strictly equal to the claimed entry.)

sigstore/_verify.py
woodruffw: https://github.com/sigstore/sigstore-python/issues/252

sigstore/_cli.py
haydentherapper: For Cosign, it's the "rekor bundle" where we extract the signature and certificate. You could argue it's putting additional trust in Rekor though.

sigstore/_verify.py
haydentherapper: So there's no support for verification by signing key rather than cert? (If not, feature request!)

Assuming only verification by cert is currently supported, I do think we should handle a certificate chain too.

woodruffw: Correct! I'll file an issue for that.

jku (dismissed):
The rekor specific bits fly over my head a bit but this looks ok to me. 
* I think `verify()` needs a refactor though: it's now very long and quite complicated -- that could happen in another issue though
* your comment WRT inverting entry matching seems reasonable: that would likely go some way towards fixing the previous complaint?

jku:
> * your comment WRT inverting entry matching seems reasonable: that would likely go some way towards fixing the previous complaint?

actually: inverting the entry matching maybe isn't a good idea if the rekor entry could contain full certificate chains but we are ok with just the single certificate in our verfication input ?

woodruffw:
> * I think `verify()` needs a refactor though: it's now very long and quite complicated -- that could happen in another issue though

Yes, absolutely agreed. I have that earmarked in #250.

di (dismissed):

woodruffw:
Not sure why DCO is stalled here...

woodruffw:
Needs re-app, since I added some warnings to emphasize that the Rekor bundle isn't intended to be a long-term format.

di (approved):

PR #262, opened 2022-10-20
ctfe: add staging targets

asraa:
Signed-off-by: Asra Ali <asraa@google.com>

The first is from `curl https://storage.googleapis.com/tuf-root-staging/targets/ctfe_2022.pub`, the second was verbally messaged to me... @haydentherapper I think I got this right.


haydentherapper (approved):

haydentherapper:
these are correct

haydentherapper:
We needed to test automation so we did a second sharding.

woodruffw (approved):

haydentherapper:
@woodruffw You'll also need to update production - https://github.com/sigstore/root-signing/blob/main/targets/ctfe_2022.pub

We're planning to shard tomorrow

PR #267, opened 2022-10-20
sigstore: add a CT keyring, use it for SCT verification

### LINKED ISSUE #260: [CI] Integration failure: staging instance
## Staging instance failure

A scheduled test against Sigstore's staging instance has failed.

This suggests one of three conditions:

* A backwards-incompatible change in a Sigstore component;
* A regression in `sigstore-python`;
* A transient error.

The full CI failure can be found here:

https://github.com/sigstore/sigstore-python/actions/runs/3292043400

### LINKED ISSUE #263: Refactor SCT verification/CT key handling
This has come to light during the staging instance's recent sharding: there are now multiple CT signing keys, instead of just `ctfe.pub`. The newer signing keys aren't necessarily compatible with the older ones (i.e. they can be ECDSA instead of RSA), and new CT entries are signed with the latest key rather than the older ones.

As a result, our assumption of a single CT key (`ctfe.{staging.}pub`) is no longer correct. This makes SCT verification fail for newer SCTs, since they're no longer signed with the key that we've vendored.

To fix this, we need to refactor our CT key handling. In particular:

* Instead of a single CT key, we need a "keychain" of CT keys. That keychain will be essentially a `dict[KeyID, Key]`, where `KeyID` is the key ID format described in [RFC 6962 s. 3.2](https://www.rfc-editor.org/rfc/rfc6962#section-3.2). 
* When verifying an SCT, the keyring will select the correct CT key using the SCT's `logID` field, which is really just the key ID.

For the time being, we'll create this keychain by vendoring more CT keys (#262). Longer term, the TUF integration planned under #25 will be expected to retrieve all CT key targets, either via their usage metadata or some other identifying feature in the TUF repo.

woodruffw:
~~WIP; just to get something up.~~

Some TODOs:

- [x] The `--ctfe` flag behavior needs to be updated to work with the new `CTKeyring` API
- [x] Unit testing

Closes #263.

Closes #260.


sigstore/_internal/ctfe.py
```diff
@@ -0,0 +1,106 @@
...
+from cryptography.hazmat.primitives.asymmetric import ec, rsa
+
+from sigstore._utils import PublicKey, key_id, load_pem_public_key
+
+
+class CTKeyringError(Exception):
+    pass
+
+
+class CTKeyring:
```
woodruffw, line 37: Flagging for review: I welcome a better name for this class (and its corresponding exception class) 🙂

woodruffw:
This is good for an initial review. Key changes:

* We no longer assume that there's only one CTFE signing key during SCT verification. Instead, we have a `CTKeyring` that contains one or more signing keys, mapped by their key ID (per RFC 6962), and each SCT is verified against the signing key that it identifies.
* The `sigstore._utils` module contains a handful of cross-module helpers that I broke out as part of these changes, particularly around having a checked union type for the kinds of keys we support (RSA and EC).

sigstore/_internal/ctfe.py
```diff
@@ -0,0 +1,106 @@
...
+            keys.append(key)
+
+        return cls(keys)
+
+    def verify(self, *, key_id: bytes, signature: bytes, data: bytes) -> None:
+        key = self._keyring.get(key_id)
+        if key is None:
+            # If we don't have a key corresponding to this key ID, we can't
+            # possibly verify the signature.
+            raise CTKeyringError(f"no known key for key ID {key_id.hex()}")
```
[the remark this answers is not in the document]
woodruffw: Yeah, great point. I'll tweak the error to communicate those cases.

sigstore/_internal/ctfe.py
```diff
@@ -0,0 +1,106 @@
...
+            if not fnmatch.fnmatch(resource, "ctfe*.pub") or fnmatch.fnmatch(
+                resource, "ctfe*.staging.pub"
+            ):
+                continue
+
+            key_pem = resources.read_binary("sigstore._store", resource)
+            key = load_pem_public_key(key_pem)
+            keys.append(key)
+
+        return cls(keys)
```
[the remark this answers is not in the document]
di: Yeah, this feels redundant, but IIUC this will go away anyways once we integrate with TUF?

Another thought I had was that while this feels pretty fancy, it might be better to be more explicit about which keys we're adding to the keyring, like `keyring.add(<path>)`. But again, if this is short-lived, I think it's probably fine for now.

di (dismissed):
LGTM with or without the nits.

sigstore/_internal/ctfe.py
```diff
@@ -0,0 +1,106 @@
...
+                    padding=padding.PKCS1v15(),
+                    algorithm=hashes.SHA256(),
+                )
+            elif isinstance(key, ec.EllipticCurvePublicKey):
+                key.verify(
+                    signature=signature,
+                    data=data,
+                    signature_algorithm=ec.ECDSA(hashes.SHA256()),
+                )
+            else:
```
haydentherapper, line 102: LGTM for key type support - IIRC CT-go doesn't support ED25519, but I see no reason to switch to it over ECDSA.

sigstore/_internal/ctfe.py
```diff
@@ -0,0 +1,106 @@
...
+                # Unreachable.
+                raise CTKeyringError("unreachable")
```
haydentherapper, line 104: nit: Error could be "unsupported key type" to be more descriptive

sigstore/_internal/rekor/client.py
```diff
@@ -287,31 +289,20 @@ def __init__(self, url: str, pubkey: bytes, ctfe_pubkey: bytes) -> None:
...
+        self._ct_keyring = ct_keyring
 
     def __del__(self) -> None:
         self.session.close()
 
     @classmethod
     def production(cls) -> RekorClient:
         return cls(
-            DEFAULT_REKOR_URL, _DEFAULT_REKOR_ROOT_PUBKEY, _DEFAULT_REKOR_CTFE_PUBKEY
+            DEFAULT_REKOR_URL, _DEFAULT_REKOR_ROOT_PUBKEY, CTKeyring.production()
```
haydentherapper, line 300: Hm, this would require larger changes than should be in this PR, but is there a separate client for CT? The CT log and Rekor are separate logs. The CT log is accessed at ctfe.sigstore.dev/<shard>, and while all Sigstore clients don't need to do online verification (since they only verify the SCTs offline), it's separate from rekor. I'd recommend splitting these apart, or at least associated CT with Fulcio rather than Rekor

sigstore/_internal/ctfe.py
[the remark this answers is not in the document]
woodruffw: > but IIUC this will go away anyways once we integrate with TUF?

Yep! This is all a stand-in for retrieving these as TUF targets.

I like that `keyring.add` API idea :slightly_smiling_face: -- the only hiccup is that it'll probably need to be `keyring.add(key_pem)` rather than a filesystem path, since the Python 3.7 compatible `importlib` APIs don't support treating resources as paths.

sigstore/_internal/sct.py
```diff
@@ -48,18 +47,18 @@ def _pack_signed_entry(
         pack_format = "!BBB{cert_der_len}s"
         cert_der = cert.public_bytes(encoding=serialization.Encoding.DER)
     elif sct.entry_type == LogEntryType.PRE_CERTIFICATE:
-        if not issuer_key_hash or len(issuer_key_hash) != 32:
+        if not issuer_key_id or len(issuer_key_id) != 32:
             raise InvalidSctError("API misuse: issuer key hash missing")
```
haydentherapper, line 51: nit: comment update to `issuer key ID`

haydentherapper (dismissed):
LGTM, just some nits and a comment for unrelated changes

sigstore/_internal/rekor/client.py
woodruffw: Yeah, @asraa brought this up as feedback here: https://github.com/sigstore/sigstore-python/issues/263#issuecomment-1286108466

I agree that this should be broken out, and I'll do that in a follow-up :slightly_smiling_face:

sigstore/_internal/sct.py
woodruffw: Meaning you Want me to add an inline source comment here? I made this change primarily to make the naming consistent with RFC 6962, which refers to this identifier as a "key ID" rather than an issuer hash.

I'm happy to say as much in a source comment, as well.

haydentherapper: I just meant in the comment "API misuse: issuer key hash missing"

sigstore/_internal/ctfe.py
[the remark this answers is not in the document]
woodruffw: Done. Let me know what you think!

sigstore/_internal/ctfe.py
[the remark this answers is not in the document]
woodruffw: Okay, I've specialized `CTKeyringError` to provide a more detailed `InvalidSctError` (ugh, inconsistent capitalization in our error classes...).

Longer term, I probably need to refactor the top-level `sign(...)` API to have `SigningResult` be similar to `VerificationResult`, so that we can do this error messaging at that layer instead.

sigstore/_internal/sct.py
woodruffw: Gotcha, thanks for catching that!

di (approved):

Issue #273, opened 2022-10-21
Implement a sign/verify API callable from Python


mbestavros:
**Description**

Currently, `sigstore-python` only exposes a simple sign/verify API through its CLI. It's likely that some Python projects interested in integrating sigstore would benefit from a similar interface available directly in code.

cc @lukehinds @mayaCostantini

di:
I think this is probably a duplicate of https://github.com/sigstore/sigstore-python/issues/4?

woodruffw:
Yes, I believe so. We do currently have simple sign/verify APIs (under `sigstore._sign` and `sigstore._verify`), but they're considered internal for the moment and are subject to major changes between releases.

mbestavros:
Yep, I'll close this.

Issue #250, opened 2022-10-13
Chore: Refactor how we handle verification materials


woodruffw:
Conceptually, there are three mandatory pieces of sigstore verification material, and one optional piece:

* The artifact's SHA256 hash (naive datatype: `str`)
* The artifact's signature (naive datatype: `bytes`, or `str` if base64-encoded)
* The artifact's PEM-formatted signing certificate (naive datatype: `str`)
* (Optionally) the offline Rekor entry (JSON, deserialized to `RekorBundle` or `RekorEntry` in #247)

Right now, the `verify` API takes these as independent parameters, along with some other options (which will grow over time, as we support different configurations and verification policies):

```python
    def verify(
        self,
        input_: bytes,
        certificate: bytes,
        signature: bytes,
        expected_cert_email: Optional[str] = None,
        expected_cert_oidc_issuer: Optional[str] = None,
        offline_rekor_entry: Optional[RekorEntry] = None,
    ) -> VerificationResult:
```

This is going to become unwieldly, so we should probably rethink the top-level API here.

Related:

* #247 
* The "sigstore bundle" design in https://github.com/sigstore/cosign/pull/2204, which will contain most (all?) of these inputs

woodruffw:
One thought: a top-level `verify` API like this might be nice:

```python
def verify(bundle: VerificationBundle, options: VerificationOptions) -> VerificationResult
```

...where `VerificationBundle` could be constructed from a Sigstore bundle, from individual inputs, etc., and `VerificationOptions` could contain settings for the expected OIDC issuer, etc.

woodruffw:
More thinking out loud: `VerificationBundle` should be correct by construction, meaning that it should be impossible to create a `VerificationBundle` that includes an offline `RekorEntry` **without** checking that entry's relevancy to the rest of the verification materials.

woodruffw:
Temporarily blocked on #289 and #288, since we want this functionality in before a more general refactor.

mayaCostantini:
Was a verification bundle schema already discussed somewhere? I would be interested in taking up this issue if it has no assignee yet.

di:
@mayaCostantini Yes, here: https://github.com/sigstore/cosign/issues/2131

woodruffw:
Additionally, the protobuf definition for the bundle is being developed here: https://github.com/sigstore/protobuf-specs

mayaCostantini:
@di @woodruffw Thank you for the info, I will look into it soon :+1:

[...]

PR #274, opened 2022-10-21
Implement a Python-native sign/verify API

mbestavros:
#### Summary

Linked issue: #273 

This PR exposes a similar sign/verify interface to what's currently available from the `sigstore-python` CLI in native Python. This will allow other Python projects to sign and verify directly, without needing to wrap the CLI.

cc @lukehinds @mayaCostantini

Signed-off-by: Mark Bestavros <mbestavr@redhat.com>

### DIFF
--- a/sigstore/_api.py
+++ b/sigstore/_api.py
@@ -1,0 +1,88 @@
+from textwrap import dedent
+from typing import cast
+
+from sigstore._internal.oidc.ambient import (
+    detect_credential,
+    GitHubOidcPermissionCredentialError
+)
+from sigstore._internal.oidc.issuer import Issuer
+from sigstore._internal.oidc.oauth import (
+    DEFAULT_OAUTH_ISSUER,
+    get_identity_token,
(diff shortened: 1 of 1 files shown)


woodruffw (changes requested):
I don't think we want to provide this quite yet -- the underlying APIs here are still undergoing a decent amount of flux (especially around more fine-grained claim/identity verification), and I plan on generally refactoring the (internal) verification APIs in #250 in a manner that probably won't mesh well with this.

IMO, we should accomplish a stable public API here by polishing the current APIs under `sigstore._sign` and `sigstore._verify`, and then making those modules public (e.g. `from sigstore.sign import sign`).

woodruffw:
(That being said, thank you for opening this PR! It'll definitely help serve as a frame of reference for the APIs that Python-level consumers might expect.)

woodruffw:
I'm fleshing out the public API in #299, so closing. Thanks again!

PR #288, opened 2022-11-03
_verify: Check for URI SANs when verifying certificate emails

tetsuo-cpp:

PR #289, opened 2022-11-03
Support `--cert-identity`

### LINKED ISSUE #108: Idea: alias `--cert-email` as `--cert-identity`/`--cert-san`
Breaking this out for discussion from #107: `--cert-email` can be misleading when the OIDC identity in question isn't an email, but rather a GitHub workflow or other URL. We could provide CLI aliases that make this clearer.

cc @di for thoughts.

woodruffw:
Closes #108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,8 @@ Verifying:
 <!-- @begin-sigstore-verify-help@ -->
 ```
 usage: sigstore verify [-h] [--certificate FILE] [--signature FILE]
-                       [--rekor-bundle FILE] [--cert-email EMAIL]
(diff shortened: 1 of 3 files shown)


woodruffw:
NB: We'll also need to update the GitHub Action, once this makes it into a release.

tetsuo-cpp:
> NB: We'll also need to update the GitHub Action, once this makes it into a release.

Sure, I can do that.

Let's cut a release once #288 makes its way in.

tetsuo-cpp (approved):
LGTM!

PR #298, opened 2022-11-10
workflows: Add conformance testing workflow

### LINKED ISSUE #297: Add conformance testing workflow
We've recently pushed a v0.0.1 release for `sigstore-conformance` so we should add this to the workflow for `sigstore-python`.

tetsuo-cpp:
Closes #297


.github/sigstore-python-conformance
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
+
+"""
+A wrapper to convert `sigstore-conformance` CLI protocol invocations to match `sigstore-python`.
```
tetsuo-cpp, line 4: Not really sure where the right spot for this is.

It doesn't seem like it belongs in the actual package itself. But on the other hand, I want to access this wrapper from the `sigstore-conformance` tests so that when we iterate on the action, we have visibility when we're breaking basic things.

woodruffw: IMO `test/integration/sigstore-python-conformance` would be a reasonable path -- that would reuse our existing test structure but shouldn't interfere with `pytest`.

If it does, we could move our current unit tests from `test/` to `test/unit.`.

woodruffw, line 4: Separate tiny nit: could we expand this comment to include a link to the `sigstore-conformance` repo?

tetsuo-cpp: Good idea. I think that's definitely cleaner.

tetsuo-cpp: [5637528](https://github.com/sigstore/sigstore-python/pull/298/commits/563752850e7d1812ca70bc8b9b0dd53bc5f31a3e)

tetsuo-cpp: [3da269a](https://github.com/sigstore/sigstore-python/pull/298/commits/3da269ad139b52ead77b627f564cea68e0b9abd7)

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,23 @@
...
+    permissions:
+      # Needed to access the workflow's OIDC identity.
+      id-token: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@d171c3b028d844f2bf14e9fdec0c58114451e4bf
+      - uses: actions/setup-python@7f80679172b057fc5e90d70d197929d454754a5a
+      - name: install sigstore-python
+        run: python -m pip install .
+      - uses: trailofbits/sigstore-conformance@v0.0.1
```
jku, line 21: nit: could use hash pinning like the other actions do -- dependabot* should still follow your latest release.

*) just realized dependabot does not seem to be set up for GH actions currently: will file an issue

tetsuo-cpp: [6ca2ed7](https://github.com/sigstore/sigstore-python/pull/298/commits/6ca2ed7d0212af7dddfefe92d68cf88445bcf9c9)

jku:
Github is currently censoring me: leaving a review doesn't do anything... So here it is as a comment:

--- 
(Github ate my original review: apologies if this is a duplicate)

Looks good to me. The brutally simple test scaffold seems like a good start: it may be too simplistic but this should be a good way to find out.

Can you verify the effects of `id-toke: write`: We are delegating the current toplevel workflow identity (so in this PR "sigstore/sigstore-python/.github/workflows/conformance.yml") to sigstore-conformance to use, correct? So security-wise this project just needs to remember to not include this workflow in release.yml (to avoid using the release identity for this).

woodruffw (approved):
LGTM! :shipit:

PR #307, opened 2022-11-21
workflows/staging-tests: add missing identity check

woodruffw:
#299 made `--cert-identity` mandatory, so this should fix the staging tests by using it.

### DIFF
--- a/.github/workflows/staging-tests.yml
+++ b/.github/workflows/staging-tests.yml
@@ -43,6 +43,7 @@ jobs:
           # also test it.
           ./staging-env/bin/python -m sigstore verify --staging \
             --cert-oidc-issuer https://token.actions.githubusercontent.com \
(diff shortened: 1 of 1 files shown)


woodruffw:
I'm going to push up a temporary commit enabling it on this branch to confirm it works, then revert.

di (approved):

di:
> Failure reason: Certificate's SANs do not match https://github.com/sigstore/sigstore-python/.github/.workflows/staging-tests.yml@refs/heads/ww/fix-integration

Might be nice to have this say what the SAN actually is/are.

woodruffw:
> Might be nice to have this say what the SAN actually is/are.

Yeah, I was thinking about how best to expose this: the cert can potentially have multiple SANs of multiple types (I believe Fulcio currently only allows one SAN, but maybe we shouldn't assume that?).

I'll do that in a follow-up.

woodruffw:
Confirmed that the conformance suite is passing, so this should be good to go.

woodruffw:
Opened https://github.com/sigstore/sigstore-python/pull/309 for the failure reason side.

di (approved):

PR #323, opened 2022-11-29
_cli: Add `--certificate-chain` and support `--rekor-url` for verification

### LINKED ISSUE #296: Support custom Fulcio and Rekor configuration for verification
We're going to need this for https://github.com/trailofbits/sigstore-conformance/issues/25 as this is needed to point the client at mock servers during verification. At the moment, we expose the flag but don't support it:
```python
    if args.staging:
        logger.debug("verify: staging instances requested")
        verifier = Verifier.staging()
    elif args.rekor_url == DEFAULT_REKOR_URL:
        verifier = Verifier.production()
    else:
        # TODO: We need CLI flags that allow the user to figure the Fulcio cert chain
        # for verification.
        args._parser.error(
            "Custom Rekor and Fulcio configuration for verification isn't fully supported yet!",
        )
```

tetsuo-cpp:
Closes #296

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -42,7 +42,11 @@ from sigstore._internal.rekor.client import (
     RekorEntry,
 )
 from sigstore._sign import Signer
-from sigstore._utils import load_pem_public_key
+from sigstore._utils import (
+    SplitCertificateChainError,
+    load_pem_public_key,
+    split_certificate_chain,
+)
 from sigstore._verify import (
     CertificateVerificationFailure,
     RekorEntryMissing,
@@ -278,7 +282,7 @@ def _parser() -> argparse.ArgumentParser:
     verification_options.add_argument(
         "--certificate-chain",
         metavar="FILE",
-        type=argparse.FileType("rb"),
+        type=argparse.FileType("r"),
         help=(
             "Path to a list of CA certificates in PEM format which will be needed when building "
             "the certificate chain for the signing certificate"
@@ -551,9 +555,7 @@ def _verify(args: argparse.Namespace) -> None:
             )
 
         try:
-            certificate_chain = _split_certificate_chain(
-                args.certificate_chain.read_text()
-            )
+            certificate_chain = split_certificate_chain(args.certificate_chain.read())
         except SplitCertificateChainError as error:
             args._parser.error(f"Failed to parse certificate chain: {error}")
 
@@ -717,35 +719,3 @@ def _get_identity_token(args: argparse.Namespace) -> Optional[str]:
             issuer,
         )
     return token
-
-
-class SplitCertificateChainError(Exception):
-    pass
-
-
-def _split_certificate_chain(chain_pem: str) -> List[bytes]:
-    PEM_BEGIN_CERTIFICATE = "-----BEGIN CERTIFICATE-----"
-
-    # Check for no certificates
-    if not chain_pem:
-        raise SplitCertificateChainError("empty PEM file")
-
-    # Use the "begin certificate" marker as a delimiter to split the chain
-    certificate_chain = chain_pem.split(PEM_BEGIN_CERTIFICATE)
-
-    # The first entry in the list should be empty since we split by the "begin certificate" marker
-    # and there should be nothing before the first certificate
-    if certificate_chain[0]:
-        raise SplitCertificateChainError(
-            "encountered unrecognized content before first PEM entry"
-        )
-
-    # Remove the empty entry
-    certificate_chain = certificate_chain[1:]
-
-    # Add the delimiters back into each entry since this is required for valid PEM
-    certificate_chain = [
-        (PEM_BEGIN_CERTIFICATE + c).encode() for c in certificate_chain
-    ]
-
-    return certificate_chain
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -557,7 +557,7 @@ def _verify(args: argparse.Namespace) -> None:
                 args.certificate_chain.read_text()
             )
         except SplitCertificateChainError as error:
-            args._parser.error(f"Failed to split certificate chain: {error}")
+            args._parser.error(f"Failed to parse certificate chain: {error}")
 
         verifier = Verifier(
             rekor=RekorClient(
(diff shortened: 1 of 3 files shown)


sigstore/_cli.py
```diff
@@ -692,3 +719,35 @@ def _get_identity_token(args: argparse.Namespace) -> Optional[str]:
             issuer,
         )
     return token
+
+
+class SplitCertificateChainError(Exception):
+    pass
+
+
+def _split_certificate_chain(chain_pem: str) -> List[bytes]:
```
tetsuo-cpp, line 728: This is pretty nasty. Not sure if there is a cleaner way to do it.

I believe that the `load_pem...`functions in `cryptography` load the first valid PEM entry and leave the rest so I don't think they can help us here.

sigstore/_cli.py
```diff
@@ -536,10 +547,26 @@ def _verify(args: argparse.Namespace) -> None:
...
+            certificate_chain = _split_certificate_chain(
+                args.certificate_chain.read_text()
+            )
+        except SplitCertificateChainError as error:
+            args._parser.error(f"Failed to split certificate chain: {error}")
+
+        verifier = Verifier(
+            rekor=RekorClient(
+                url=args.rekor_url,
+                pubkey=args.rekor_root_pubkey.read(),
```
tetsuo-cpp, line 565: I wonder whether we should warn if a user supplies `--rekor-root-pubkey` but don't set the `--rekor-url`. Same with the new `--certificate-chain`. At the moment they just get ignored which might be confusing.

This issue also exists in the signing code path.

sigstore/_cli.py
woodruffw: Yeah, this isn't ideal, but it looks like neither `cryptography` nor `pyOpenSSL` exposes a "load an entire chain from PEMs" API. 

I _think_ the corresponding OpenSSL API is `SSL_CTX_get0_chain_certs`, so this might be worth a patch to pyOpenSSL in the medium term. But for now, something like this approach is fine 🙂

sigstore/_cli.py
woodruffw: Yeah, it would be good to catch this and flag it as a warning at the minimum (or maybe even an error, since we expect it to fail anyways).

sigstore/_cli.py
```diff
@@ -692,3 +719,35 @@ def _get_identity_token(args: argparse.Namespace) -> Optional[str]:
...
+
+def _split_certificate_chain(chain_pem: str) -> List[bytes]:
+    PEM_BEGIN_CERTIFICATE = "-----BEGIN CERTIFICATE-----"
+
+    # Use the "begin certificate" marker as a delimiter to split the chain
+    certificate_chain = chain_pem.split(PEM_BEGIN_CERTIFICATE)
+
+    # Check for no certificates.
+    if not certificate_chain:
+        raise SplitCertificateChainError("empty PEM file")
```
woodruffw, line 736: I think this won't actually catch the empty PEM file case, since `split` on an empty string always returns `['']`, not `[]`. Maybe test the `chain_pem` instead?

```suggestion
    # Check for no certificates.
    if not chain_pem:
        raise SplitCertificateChainError("empty PEM file")
```

sigstore/_cli.py
```diff
@@ -536,10 +547,26 @@ def _verify(args: argparse.Namespace) -> None:
...
+            args._parser.error(
+                "Custom Rekor URL used without specifying --certificate-chain"
+            )
+
+        try:
+            certificate_chain = _split_certificate_chain(
+                args.certificate_chain.read_text()
+            )
+        except SplitCertificateChainError as error:
+            args._parser.error(f"Failed to split certificate chain: {error}")
```
woodruffw, line 560: Just to make it a little more generic:

```suggestion
            args._parser.error(f"Failed to parse certificate chain: {error}")
```

sigstore/_cli.py
```diff
@@ -275,6 +275,17 @@ def _parser() -> argparse.ArgumentParser:
...
     verification_options = verify.add_argument_group("Extended verification options")
+    verification_options.add_argument(
+        "--certificate-chain",
+        metavar="FILE",
+        type=argparse.FileType("rb"),
+        help=(
+            "Path to a list of CA certificates in PEM format which will be needed when building "
+            "the certificate chain for the signing certificate; must start with the parent "
+            "intermediate CA certificate of the signing certificate and end with the root "
+            "certificate"
```
woodruffw, line 286: I might be wrong, but I don't think we actually require this ordering: this cert chain ends up in an `X509Store`, which AFAICT allows certs to be added in any order.

sigstore/_cli.py
woodruffw, line 728: What do you think about moving this helper + exception to the `_utils` module? We already have a couple of PEM helpers in there, and it would make unit testing a little easier.

sigstore/_cli.py
tetsuo-cpp: True, I'll remove that. I just pasted that from `cosign`.

tetsuo-cpp: [b3f3847](https://github.com/sigstore/sigstore-python/pull/323/commits/b3f3847d70cbd71d0523091a5460ef45a2dd11fc)

sigstore/_cli.py
tetsuo-cpp: [ce3dc89](https://github.com/sigstore/sigstore-python/pull/323/commits/ce3dc8921a065472b2583784a49604f428fb0135)

sigstore/_cli.py
tetsuo-cpp: [da25f55](https://github.com/sigstore/sigstore-python/pull/323/commits/da25f55dbf89d8b906205a9dad4707c92f21939a)

sigstore/_cli.py
woodruffw: Let's track this with a follow-up issue.

woodruffw (approved):

sigstore/_cli.py
woodruffw: NB: I opened https://github.com/pyca/cryptography/pull/7878 to add this API, so we'll be able to use it starting with cryptography 39 (assuming they accept it).

tetsuo-cpp: Awesome, thanks @woodruffw.

PR #331, opened 2022-12-08
Run tests in CI with full verbosity

di:
[A recent test run](https://github.com/sigstore/sigstore-python/actions/runs/3641503672/jobs/6147515676) seems to have caused a 500-level error with Rekor, but the full certificate that was uploaded in the test was truncated.

This PR ensures CI tests are run with full verbosity so long strings like certificates appear intact in the logs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       - name: deps
         run: make dev SIGSTORE_EXTRA=test
       - name: test
-        run: make test
+        run: make test TEST_ARGS="-vv"
       - name: test (offline)
         run: |
           # We use `unshare` to "un-share" the default networking namespace,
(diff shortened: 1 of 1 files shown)


jku:
The output truncation is unfortunate but yes, looks like GitHubOidcPermissionCredentialError.

tetsuo-cpp (approved):

tetsuo-cpp:
I believe it's because this PR is from a fork rather than a branch. I'm not sure whether there's a way around this.

woodruffw:
> I believe it's because this PR is from a fork rather than a branch. I'm not sure whether there's a way around this.

It's not a good solution, but the trick we use on `gh-action-sigstore-python` is simply to skip tests that require the CI's OIDC credential:

```yaml
    if: (github.event_name != 'pull_request') || !github.event.pull_request.head.repo.fork
```

This would effectively disable the conformance suite on third-party PRs, though.

di:
Reopened as https://github.com/sigstore/sigstore-python/pull/334 to get tests to pass.

PR #342, opened 2022-12-13
workflows/conformance: bump sigstore-conformance action

woodruffw:
Doing this from a fork, to confirm that it works as expected.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -4,8 +4,9 @@ on:
   push:
     branches:
       - main
-  pull_request:
   workflow_dispatch:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


woodruffw:
Hmm, I added the appropriate label, but it looks like it hasn't triggered the conformance suite. Will debug.

woodruffw:
It triggers on `pull_request` (and fails, as expected), but doesn't trigger on `pull_request_target`. Hmm.

woodruffw:
Oh, apparently this is a known limitation of `pull_request_target`: it needs to be merged into `main` before it takes effect.

di (approved):

Issue #324, opened 2022-11-30
CLI: Limit user confusion over Rekor/Fulcio instance and state URLs


woodruffw:
Follow-up from https://github.com/sigstore/sigstore-python/pull/323#discussion_r1034822945: the various CLI options that control the Fulcio and Rekor instance state aren't "all or nothing," meaning that a user might forget to pass one and end up with a confusing error somewhere deep in signing or verification (e.g., when they reach SCT verification).

This isn't *dangerous*, per se, but we can definitely improve the UX.

jku:
#351 is going to make this even more visible in at least a couple of places. It does things like:
```python
        updater = TrustUpdater.production()
        if args.ctfe_pem is not None:
            ctfe_keys = [args.ctfe_pem.read()]
        else:
            ctfe_keys = updater.get_ctfe_keys()
```

This happens in a place where either rekor or fulcio URL is non-default... but we end up loading production ctfe keys if none are provided by user

dmitris:
just a thought - wouldn't it be better to call the flag **`--trusted-roots`** (plural)? In case the user needs to provide multiple roots, it would be clear and intuitive from the flag name itself that it is a valid and supported use case.  And it still should be clear to those who have just one trusted root, that it is a valid much for the {N>=1} cardinality implied by the name 😄  It would also cover better (IMO) the fact that with this flag, the user can provide CA roots for Fulcio, Rekor, BYO CA etc.

jku:
> In case the user needs to provide multiple roots, it would be clear and intuitive from the flag name itself that it is a valid and supported use case

I've seen this idea mentioned but  I don't understand what exactly it would mean. I definitely don't think it's intuitively clear and would like to see the concept explained.

woodruffw:
> just a thought - wouldn't it be better to call the flag **`--trusted-roots`** (plural)? In case the user needs to provide multiple roots

At least in my English dialect, this can go either way: the "trusted root" is either the thing containing a whole bunch of trust material, *or* it's a single piece of trust material in a larger "trust bundle." The protobuf-specs use it in the former way (the `TrustedRoot` message), so IMO we should just be consistent with them here 🙂

dmitris:
ok that's fine, thanks 👍

PR #351, opened 2022-12-20
Use TUF to download key/cert material

### LINKED ISSUE #25: Retrieve CTFE signing key via TUF

jku:
Should fix #25 by using TUF to download all certs and keys. It is based on code by @joshuagl, @wxjdsr (co-authors on first commit in this branch) but it is a significant rewrite compared to their branches.

FYI also @woodruffw, @di,  @SantiagoTorres. 

I'm allowing edits by maintainers, go wild if you feel like it. 

Status (updated dec 22):
* Works for signing and verifying
* The embedded keys/certs are no longer used or included in git (except staging keys as test assets)
* if the default (production) or --staging flows are used, TUF is used to fetch key material
* TUF is also used in some other cases (code does the same thing as the previous implementation, and uses production key material if none is provided)
* the implementation peeks into python-tuf internals currently (to get access to top-level target metadata). This should be okay for now but a long-term solution might be  https://github.com/theupdateframework/python-tuf/issues/1995
* When the caches are hot (certs, keys and metadata have been downloaded already), every "sigstore verify" and "sigstore sign" (with the production or staging flows) now results in a minimum of two network  requests (404 and a small download). This [may be avoidable](https://github.com/theupdateframework/python-tuf/issues/2225) in future but currently always happens
* with empty caches we download ~10 files, in the order of 20kB altogether. These files are then cached for later use

Some TODO items (there are certainly more)
- testing: many existing tests now require network
  - [x] we could mock the requests for staging in the tests
  - [ ] longer term could make it possible to not do network requests when metadata is not expired
- [x] testing: the TUF module has no tests (it does get tested by all tests using staging though)
- [x] TUF metadata is stored in `~/.local/share/sigstore-python/tuf/<url>/` and targets are cached in `~/.cache/sigstore-python/tuf/<url>/ ` -- this is not great for windows
- [x] logging in python-tuf is pretty verbose at INFO: we'll have to change that at some level
- [x] error handling is almost non-existent, but let's handle in later PR #357 . The situations I've identified are:
  * python-tuf RepositoryError (metadata from repo is not valid in some way)
  * python-tuf DownloadError (trouble downloading things)
  * python-tuf OSError (trouble storing things on disk)
  * discovery errors in our _tuf module: the repository content was not what we expected  
- [x] should remove the embedded key/cert material from the repo


jku:
Oh, a note on fulcio: I now feed all certs with usage "Fulcio" and status "Active" to Verifier with the assumption that
* these all form a chain
* cryptography copes with the (non-controlled) order of these certs

hopefully there are no problems with these assumptions

sigstore/_internal/rekor/client.py
```diff
@@ -411,14 +406,10 @@ def __del__(self) -> None:
...
-            DEFAULT_REKOR_URL, _DEFAULT_REKOR_ROOT_PUBKEY, CTKeyring.production()
-        )
-
-    @classmethod
-    def staging(cls) -> RekorClient:
-        return cls(STAGING_REKOR_URL, _STAGING_REKOR_ROOT_PUBKEY, CTKeyring.staging())
+    def with_updater(cls, updater: TrustUpdater) -> RekorClient:
+        rekor_key = updater.get_rekor_key()
+        ctfe_keys = updater.get_ctfe_keys()
+        return cls(DEFAULT_REKOR_URL, rekor_key, CTKeyring(ctfe_keys))
```
jku, line 412: This is a bug: should not use a DEFAULT_REKOR_URL here. Need to reconsider

woodruffw:
> Oh, a note on fulcio: I now feed all certs with usage "Fulcio" and status "Active" to Verifier with the assumption that
> 
> * these all form a chain
> * cryptography copes with the (non-controlled) order of these certs
> 
> hopefully there are no problems with these assumptions

These assumptions should be fine! We throw everything into an OpenSSL X.509 store, and I don't believe it imposes any order of insertion.

sigstore/_internal/rekor/client.py
woodruffw: Making sure I understand: the bug here is that this URL will be wrong with `with_updater` is used on `TrustUpdater.staging()`, correct?

sigstore/_internal/tuf.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def _get_dirs(url: str) -> Tuple[Path, Path]:
+    """Return metadata dir and target cache dir for URL"""
+    # NOTE: this is not great for windows: should maybe depend on appdirs?
+    # TODO: there should be URL normalization if URLs come from user
+    dir = parse.quote(url, safe="")
+    md_dir = Path.home() / ".local" / "share" / "sigstore-python" / "tuf" / dir
+    targets_dir = Path.home() / ".cache" / "sigstore-python" / "tuf" / dir
+    return md_dir, targets_dir
```
woodruffw, line 37: NTS: Might be worth using <https://pypi.org/project/pyxdg/> or similar here.

I agree that it's not great for Windows; maybe there's an even higher wrapper that handles both XDG and Windows-preferred paths.

woodruffw: (And macOS too, since `~/Library/...` is preferred over XDG-style paths.)

di: I agree that https://pypi.org/project/appdirs/ is probably a good solution for this.

woodruffw: SG! I'll do that in a bit.

sigstore/_internal/rekor/client.py
woodruffw: This is causing `sigstore verify --staging` to fail, so I'll look in to it next.

woodruffw: More generally, this suggests that we should really have a top-level configuration object, similar to a configurator, that unifies all of this state (since we should always use like-with-like, so there's no need for flexibility here).

woodruffw: 3a8f026 has the initial workaround.

woodruffw: Another option here is to let individual components manage their TUF state -- is there any risk in having `RekorClient.production()` create its own `TrustUpdater`, other than the performance hit with the current implementation?

Alternatively, maybe a `TrustUpdater` singleton.

woodruffw:
Online tests now pass, but the forced offline tests are still failing (as expected).

woodruffw:
> * [ ]  logging in python-tuf is pretty verbose at INFO: we'll have to change that at some level
> *

IMO this is worth an upstream PR -- looks like most of these should really be `DEBUG` level instead. I can look into that.

sigstore/_internal/rekor/client.py
jku: Thanks for investigating.

Making TrustUpdater aware of the sigstore mechanisms like RekorClient seems wrong.... although there are edge cases like CTKeyring thatare more data types that could indeed be return values for TrustUpdater.

I think the original idea is still salvageable -- I just removed the RekorClient.production()/staging() constructors too greedily. Creating a single TrustUpdater and supplying that "down the call stack" should work (there are only a few places where it's needed so it's not too much work).
 
I will try modifying the RekorClient constructors and will ask you to review if it seems to work.

sigstore/_verify/verifier.py
```diff
@@ -227,7 +240,7 @@ def verify(
             entry = materials.rekor_entry(self._rekor)
         except RekorEntryMissingError:
             return RekorEntryMissing(
-                signature=materials.signature,
+                signature=base64.b64encode(materials.signature).decode(),
```
jku, line 243: just checking: is this change supposed to be here?

jku:
out of the remaining issues, the lack of offline tests seems the most pressing. This happens because all uses of production or staging flows for sign() and verify() currently mean a minimum of two requests to  the TUF repository -- and tests use especially staging extensively.

There are potentially two separate avenues for solving this:
* in general (not just in the specific case of testing), sigstore-python would rather not make those requests if the local metadata is not expired: python-tuf does not support this as it would be against the specification... but I think adding that feature to python-tuf may still make sense. This is not going to be entirely trivial so I would not expect this to happen in the next week or two. I don't think working around python-tuf on this is a good idea.
* we could tweak the test environment so that the requests do not happen there: I have a few ideas which I plan to try out today

jku:
There weren't actually as many offline tests as I expected... but I made the mock fixture anyway as it's not that big -- of course it does mean I end up adding the staging key material in test assets.

There are still some failing tests, I'm looking at those after lunch: at least it looks line the pytest.mark.parametrize happens even if the test is marked as "online" and fails if the parametrized setup requires a connection

jku:
I think TrustUpdater should have a few tests:
* make sure the expected things are downloaded over network (based on initial cache contents): not more, not less. We can mock the network requests for staging. Testing this online for production might be useful but a little trickier (as we can't tell what should get downloaded as easily)
* make sure the `get_*()` return values are sensible: we could just hard  code the expected values for staging

jku:
The only clear blocker right now is the exceptions review: I could use some help with that -- or I'm fine with leaving that to someone else. My understanding of the possible errors are:
* from python-tuf: RepositoryError (metadata from repo is not valid in some way: many subclasses of this error are possible but they all mean update is not possible)
* from: python-tuf DownloadError (trouble downloading things)
* from python-tuf: OSError (trouble storing things on disk)
* discovery errors in sigstore_internal.tuf module (repository content is not what we expected)

Any suggestions from maintainers?

test/unit/conftest.py
```diff
@@ -121,3 +136,35 @@ def verify(self, cert):
...
+    success = defaultdict(int)
+    failure = defaultdict(int)
+
+    class MockFetcher(FetcherInterface):
+        def _fetch(self, url: str):
+            filename = os.path.basename(url)
+            filepath = _TUF_ASSETS / filename
+            if filepath.is_file():
+                success[filename] += 1
+                return open(filepath, "rb")
```
jku, line 157: I suppose this should actually return a function that yields the content, otherwise the test leaves the file unclosed

sigstore/_verify/verifier.py
woodruffw: Yep -- `RekorEntryMissing.signature` is meant to be a `str`, but we were populating it with the raw `bytes` from the materials by accident. I noticed that while (accidentally) trigging the error case by mixing up different Sigstore instances 🙂

woodruffw:
> The only clear blocker right now is the exceptions review: I could use some help with that -- or I'm fine with leaving that to someone else. My understanding of the possible errors are:
> 
> * from python-tuf: RepositoryError (metadata from repo is not valid in some way: many subclasses of this error are possible but they all mean we cannot download what we need)
> * from: python-tuf DownloadError (trouble downloading things)
> * from python-tuf: OSError (trouble storing things on disk)
> * discovery and parsing errors in sigstore_internal.tuf module (repository content is not what we expected)
> 
> Any suggestions from maintainers?

I'll give this some thought, but I'm _mostly_ okay with us punting on this for the initial PR -- `sigstore-python` needs some exception refactoring anyways, and we could probably do a better job with the overall refactor by batching any handling here into those changes.

test/unit/test_sign.py
```diff
@@ -21,20 +21,18 @@
...
 
-def test_signer_staging():
+def test_signer_staging(mock_staging_tuf):
     signer = Signer.staging()
     assert signer is not None
 
 
-@pytest.mark.online
-@pytest.mark.ambient_oidc
-@pytest.mark.parametrize("signer", [Signer.production(), Signer.staging()])
```
woodruffw, line 36: Is there a reason for removing this parametrization?

test/unit/conftest.py
woodruffw: That's worth a `NOTE` or `TODO` in the source IMO, but not a big deal otherwise -- I'm okay with cleanup at process destruction time for unit tests 🙂

woodruffw:
> There weren't actually as many offline tests as I expected... but I made the mock fixture anyway as it's not that big -- of course it does mean I end up adding the staging key material in test assets.

Thanks for doing this! IMO we should put `assets-staging-tuf/` under the pre-existing `assets/` dir, but otherwise this looks great.

test/unit/test_sign.py
jku: Yes, the parametrization setup happens even if --skip-online: and that setup currently does a network request... 

Now that I think about it, I guess the parameter could be the class, and we only instantiate inside the test. That might work?

woodruffw: That SGTM!

sigstore/_internal/tuf.py
```diff
@@ -0,0 +1,159 @@
...
+        updater.refresh()
+        return updater
+
+    def _get(self, usage: str) -> List[bytes]:
+        """Return all active targets with given usage"""
+        if not self._updater:
+            self._updater = self._setup()
+
+        data = []
+        assert self._updater._trusted_set.targets
```
woodruffw, line 118: `bandit` flagged this -- is there a reason for this `assert` beyond maybe an error handling case that we should use an exception for?

woodruffw:
This is looking pretty good to me! I think we can move it out of a draft for now; there are only a few small cleanup tasks that I think should get done before we merge.

sigstore/_cli.py
```diff
@@ -427,12 +412,21 @@ def _sign(args: argparse.Namespace) -> None:
     elif args.fulcio_url == DEFAULT_FULCIO_URL and args.rekor_url == DEFAULT_REKOR_URL:
         signer = Signer.production()
     else:
-        ct_keyring = CTKeyring([load_pem_public_key(args.ctfe_pem.read())])
+        # Assume "production" keys if none are given as arguments
+        updater = TrustUpdater.production()
+        if args.ctfe_pem is not None:
+            ctfe_keys = [args.ctfe_pem.read()]
+        else:
+            ctfe_keys = updater.get_ctfe_keys()
```
tetsuo-cpp, line 420: If we end up doing this for a custom Fulcio URL, it'd be good to log a warning as this is probably not what the user intended.

This would partially address #324.

I can follow up with another PR if you feel this is out of scope.

sigstore/_internal/tuf.py
```diff
@@ -0,0 +1,159 @@
...
+            self._metadata_dir.mkdir(parents=True, exist_ok=True)
+            root_json = read_embedded(fname)
+            with tuf_root.open("wb") as io:
+                io.write(root_json)
+
+        # intialize targets cache dir
+        # TODO: Pre-populate with any targets we ship with sources
+        self._targets_dir.mkdir(parents=True, exist_ok=True)
+
+        logger.debug("TUF metadata: %s", self._metadata_dir)
```
tetsuo-cpp, line 86: We've been using f-strings for this type of thing elsewhere in the codebase.

woodruffw: Yeah, let's normalize on f-strings here. @tetsuo-cpp feel free to push to this branch!

tetsuo-cpp: [d9aa72c](https://github.com/sigstore/sigstore-python/pull/351/commits/d9aa72c2e04477bf5e1d0b818f9ac31d4c91a651)

tetsuo-cpp:
I'm seeing some weird lint errors on this branch that I don't see in `main`, CI doesn't seem to have this issue as it's getting all the way to `bandit`: 
```
sigstore/_utils.py:161: error: Returning Any from function declared to return "bytes"  [no-any-return]
Found 1 error in 1 file (checked 24 source files)
```

Looks like that file hasn't even been touched.

jku:
> I'm seeing some weird lint errors on this branch that I don't see in `main`, CI doesn't seem to have this issue as it's getting all the way to `bandit`:
> 
> ```
> sigstore/_utils.py:161: error: Returning Any from function declared to return "bytes"  [no-any-return]
> Found 1 error in 1 file (checked 24 source files)
> ```
> 
> Looks like that file hasn't even been touched.

This is unrelated, it's not showing up in main because William added an ignore in 0647d75.

sigstore/_cli.py
jku: That sounds like a good idea but since there's even an issue already, let's leave it for a future PR

tetsuo-cpp:
> This is unrelated, it's not showing up in main because William added an ignore in [0647d75](https://github.com/sigstore/sigstore-python/commit/0647d759257d6974855547984ffef198f31b9803).

I assumed that was necessary somehow because we switched to Ruff, but your branch doesn't have that commit. Anyway, if you don't mind, I'll merge in `main` and get this branch up to date.

sigstore/_cli.py
tetsuo-cpp: Yep, sounds good.

sigstore/_internal/tuf.py
jku: A nicer solution would be great, the issue is this:
* `_trusted_set.targets` is correctly annotated `Optional[Metadata[Targets]]`
* the Optional is there because before the metadata is refreshed, `_trusted_set.targets` is None
* but at this point when metadata has been successfully refreshed we know that the actual annotation is `Metadata[Targets]`
* unfortunately mypy cannot understand this: the assert is there to inform mypy

So if you have other clever solutions, be my guest.

jku:
I pushed some tiny documentation tweaks but I would agree: this looks good to me apart from the final linting issue https://github.com/sigstore/sigstore-python/pull/351#discussion_r1054711928

sigstore/_internal/tuf.py
jku: if no smarter solutions appear, I would suggest a "# nosec" to disable bandit here: this feels like an appropriate use of assert to me.

tetsuo-cpp: Agreed, I think that's fine.

tetsuo-cpp (dismissed):
LGTM provided that the linting is sorted out. Nice work.

jku:
I added the "#nosec" to get tests passing: feel free to solve the issue in some other way if you want.

sigstore/_internal/tuf.py
woodruffw: I think we can use `typing.cast` for this! I'll push that up.

woodruffw: Ah, looks like `cast` won't quite work. But I'm going to replace this with a `# type: ignore[union-attr]`, since that silences the underlying MyPy problem rather than turning it into a Bandit problem that we then ignore 🙂

woodruffw (approved):
LGTM! :shipit:

sigstore/_internal/tuf.py
jku: that'll work but I'll note it's not just mypy: it's static typing in general for that code block. So e.g. editor predictions become unavailable as well when assert is removed, this is unfortunate.

I think the ultimate fix is to avoid Optional in cases like this. I will look at the python-tuf code and see what could be done.

sigstore/_internal/tuf.py
```diff
@@ -0,0 +1,160 @@
...
+        if not self._updater:
+            self._updater = self._setup()
+
+        data = []
+
+        # NOTE: _updater has been fully initialized at this point, but mypy can't see that.
+        targets = self._updater._trusted_set.targets.signed.targets  # type: ignore[union-attr]
+        for target_info in targets.values():
+            custom = target_info.unrecognized_fields["custom"]["sigstore"]
+            if custom["status"] == "Active" and custom["usage"] == usage:
```
asraa, line 123: note that on the verification path, Expired/Inactive targets can still be used to verify against. It might not be worth filtering these out, but rather returning a List of bytes that has this metadata attached

sigstore/_internal/tuf.py
```diff
@@ -0,0 +1,160 @@
...
+    def _get(self, usage: str) -> List[bytes]:
+        """Return all active targets with given usage"""
+        if not self._updater:
+            self._updater = self._setup()
+
+        data = []
+
+        # NOTE: _updater has been fully initialized at this point, but mypy can't see that.
+        targets = self._updater._trusted_set.targets.signed.targets  # type: ignore[union-attr]
+        for target_info in targets.values():
```
asraa, line 121: I actually would caution against using the custom metadata and use the path filters "rekor*", "fulcio*", "ctfe*", instead as the filtering mechanism. While we'll keep some of the metadata, I think we are aiming to change them (for example, status will be deprecated in favor of explicit NotBefore/NotAfter ranges)

asraa: That being said "usage" should match here and be stable

sigstore/_internal/tuf.py
jku: thanks -- I might add an argument like `active_only: bool` to the getters instead but this is a good point, not sure how I missed it.

I'll file a new issue on these, but maybe tomorrow

sigstore/_internal/tuf.py
jku: Yeah I knew the custom metadata was not a permanent solution but I'm not sure filtering by path is going to be either -- maybe makes sense to update once the target discovery and delegation plans are properly decided

asraa: > I'm not sure filtering by path is going to be either -- maybe makes sense to update once the target discovery and delegation plans are properly decided

That's fair! Yeah, in the end distributing as a single bundle would probably fix that and that likely will be the case next year. The migration to that will be slow anyway (as in, the existing targets will continue to be distributed like this).

sigstore/_internal/tuf.py
jku: following up in #375

PR #359, opened 2022-12-22
workflows: Pin actions/checkout to its latest release SHA

### LINKED ISSUE #358: actions/checkout seemingly not updated by dependabot
Most workflows use `actions/checkout@d171c3b028d844f2bf14e9fdec0c58114451e4bf`: dependabot does not seem to update it.

This is not a commit that exists in the checkout repository -- or rather, it must exist in the remote since the workflows run successfully but it's not part of the published git repository.

A wild guess: this is the SHA of a tag object (like "v2") instead of a commit. GitHub moves these tags around, meaning the tag "v2" still exists but the SHA is different, so the old SHA is now a dangling reference.

tetsuo-cpp:
Closes #358

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
           - "3.11"
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@d171c3b028d844f2bf14e9fdec0c58114451e4bf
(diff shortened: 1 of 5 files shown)


tetsuo-cpp:
I wasn't able to figure out what happened to the original SHA but it indeed doesn't exist in the commit history. In any case, the right move is to just pin it to the latest release.

di (approved):

PR #369, opened 2022-12-24
tuf: intrinsic types, use `urljoin`

woodruffw:
Just some small cleanup tasks.

### DIFF
--- a/sigstore/_internal/tuf.py
+++ b/sigstore/_internal/tuf.py
@@ -20,7 +20,6 @@ from __future__ import annotations
 
 import logging
 from pathlib import Path
-from typing import List, Optional, Tuple
 from urllib import parse
 
 import appdirs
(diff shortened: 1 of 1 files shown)


di (approved):

jku:
many of the annotation fixes are only available on 3.9 or 3.10. I'll file an issue on that

woodruffw:
> many of the annotation fixes are only available on 3.9 or 3.10.
> 
> Is the idea that it's fine for linting to only work on some newer Python version as long as runtime works on 3.7 that's mentioned in pyproject.toml?

Nope! Linting and the runtime should both function on 3.7 (or whatever our minimum Python happens to be at any point). In this case, "intrinsic" types (I don't know what to actually call them) do actually work with Python 3.7 thanks to postponed annotations with `from __future__ import annotations`.

Issue #322, opened 2022-11-29
Support `--certificate-github-workflow-` flags


tetsuo-cpp:
**Description**

I noticed that Cosign has a series of flags for checking the signing certificate extensions. From the help text:
```
--certificate-github-workflow-name string                                                  contains the workflow claim from the GitHub OIDC Identity token that contains the name of the executed workflow.
--certificate-github-workflow-ref string                                                   contains the ref claim from the GitHub OIDC Identity token that contains the git ref that the workflow run was based upon.
--certificate-github-workflow-repository string                                            contains the repository claim from the GitHub OIDC Identity token that contains the repository that the workflow run was based upon
--certificate-github-workflow-sha string                                                   contains the sha claim from the GitHub OIDC Identity token that contains the commit SHA that the workflow run was based upon.
--certificate-github-workflow-trigger string                                               contains the event_name claim from the GitHub OIDC Identity token that contains the name of the event that triggered the workflow run
```

This should be a good fit for @woodruffw's policy API.

woodruffw:
Yep! These should map 1-1 to their equivalent policies. 

Since `cosign` is currently in the process of reducing insecure/ineffective uses of their verification CLI, let's make sure we only allow these flags in ways that don't make policy verification effectively a no-op. For example, we should *probably* never allow `--certificate-github-workflow-trigger` on its own, since just about anybody can mint a cert with any particular workflow trigger embedded in its claims.

CC @znewman01 and @jku for thoughts (and particularly Zack to make sure we're aligned with `cosign` here).

jku:
I still think a usable CLI for this probably requires a service specific "endpoint" (whether it's a separate binary that just uses sigstore-python API or a mode in this cli app) that is somewhat opinionated about what it's doing. So I don't think adding options like this will make the generic CLI better for actually verifying GitHub releases.

A couple of reasons:
* A reasonable set of arguments for the verification ends up repeating many things and still requires the caller to construct the complex identity string
* like @woodruffw says, it's easy to end up not verifying something you actually should have verified
* what happens when sigstore becomes more popular -- do we add more `--service-specific-claim` flags as more services start attesting things?
 
A custom CLI endpoint on the other hand can A) make the inputs simpler and B) can require things that a "generic sigstore client" can not. So just as an example `github-release-verify <repo_name> <workflow_path> <tag> [<sha>]` should work and is probably simple enough to use without accidents. Now with the policy API that should be almost trivial to implement.

[...]

PR #379, opened 2023-01-03
_cli: add `sigstore verify identity`

woodruffw:
This adds `sigstore verify identity`, which is aliased (through some `argparse` hackery) back to `sigstore verify`.

This allows us to remain compatible with the existing CLI, while also giving us the flexibility we need to extend verification (e.g. `sigstore verify github`).

See #322.

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -267,12 +267,15 @@ def _parser() -> argparse.ArgumentParser:
 
     # `sigstore verify`
     verify = subcommands.add_parser(
-        "verify", formatter_class=argparse.ArgumentDefaultsHelpFormatter
+        "verify",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
     verify_subcommand = verify.add_subparsers(dest="verify_subcommand")
 
     # `sigstore verify identity`
-    verify_identity = verify_subcommand.add_parser("identity")
+    verify_identity = verify_subcommand.add_parser(
+        "identity", help="verify against a known identity and identity provider"
+    )
     input_options = verify_identity.add_argument_group("Verification inputs")
     input_options.add_argument(
         "--certificate",
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -100,12 +100,17 @@ def _set_default_verify_subparser(parser: argparse.ArgumentParser, name: str) ->
                 if sp_name in sys.argv[1:]:
                     subparser_found = True
         if not subparser_found:
-            # If `sigstore verify identity` wasn't passed explicitly, we need
-            # to insert the `identity` subcommand into the correct position
-            # within `sys.argv`. To do that, we get the index of the `verify`
-            # subcommand, and insert it directly after it.
-            verify_idx = sys.argv.index("verify")
-            sys.argv.insert(verify_idx + 1, name)
+            try:
+                # If `sigstore verify identity` wasn't passed explicitly, we need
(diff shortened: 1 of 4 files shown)


woodruffw:
NB: This PR doesn't add `sigstore verify github`, since I want to keep it small. I'll do that in a follow-up, once we're happy with the approach here 🙂

woodruffw:
Hmm, this doesn't quite work yet:

```console
$ sigstore verify README.mdusage: sigstore verify [-h] {identity} ...
sigstore verify: error: argument verify_subcommand: invalid choice: 'README.md' (choose from 'identity')
```

woodruffw:
Confirmed that each of the following work with these changes:

```console
$ sigstore sign --staging README.md --overwrite

$ sigstore verify README.md --cert-identity 'william@yossarian.net' --cert-oidc-issuer 'https://github.com/login/oauth' --staging

$ sigstore verify identity README.md --cert-identity 'william@yossarian.net' --cert-oidc-issuer 'https://github.com/login/oauth' --staging

**Comment by woodruffw** [comment]
Implementation details: we do a little bit of `argparse` patching via `_set_default_verify_subparser`, which does the following:

1. Nothing if the arguments look like help arguments (e.g. `--help`)
2. Scans the subparsers for the parser it was run against, looking for the default subparser (`identity` in our case)
    1. If `identity` is present, there's nothing to be done
    2. If `identity` is not present, it looks for the index of `verify` in `sys.argv` (as in `sigstore verify ...` and inserts it immediately afterwards, normalizing the command to `sigstore verify identity ...`.
        1. If `verify` itself is not present, there's nothing to be done (since we're in another `sigstore` top-level subcommand)

File: sigstore/_cli.py [hunk 1]
```diff
@@ -82,6 +82,37 @@ def _boolify_env(envvar: str) -> bool:
...
+    """
+    An argparse patch for configuring a default subparser for `sigstore verify`.
+
+    Adapted from <https://stackoverflow.com/a/26379693>
+    """
+    subparser_found = False
+    for arg in sys.argv[1:]:
+        if arg in ["-h", "--help"]:  # global help if no subparser
+            break
+    else:
```
tetsuo-cpp, line 95: Whoa... how did I not know that this exists.

sigstore/_cli.py
```diff
@@ -82,6 +82,37 @@ def _boolify_env(envvar: str) -> bool:
...
+                if sp_name in sys.argv[1:]:
+                    subparser_found = True
+        if not subparser_found:
+            try:
+                # If `sigstore verify identity` wasn't passed explicitly, we need
+                # to insert the `identity` subcommand into the correct position
+                # within `sys.argv`. To do that, we get the index of the `verify`
+                # subcommand, and insert it directly after it.
+                verify_idx = sys.argv.index("verify")
+                sys.argv.insert(verify_idx + 1, name)
```
tetsuo-cpp, line 109: Can we maybe log a warning here? I realise we don't want to make a breaking change but perhaps we can at least mark it was deprecated and slate it for future removal.

woodruffw: Yeah, a log sounds good. I'll do that in a moment.

woodruffw: https://github.com/sigstore/sigstore-python/pull/379/commits/088714b32e5d9d08a531d7c8b7ec73b7efd803ba

tetsuo-cpp (approved):
LGTM!

PR #386, opened 2023-01-05
_internal/tuf: Use expired Fulcio certificates too

### LINKED ISSUE #385: _internal/tuf: return also "expired" fulcio certificates
Splitting this from #375:
* we only use fulcio certs from TUF for verification currently
* verification should use "expired" certificates as well

I have a patch for this, will submit a PR.

jku:
Fixes #385

#### Summary

Currently only "active" Fulcio certificates are used when verifying. However, expired Fulcio certificates may have been active at signing time: We should include them in the "bundle of certificate chains" that is used during verify.

#### Release Note

* Included inactive Fulcio certificates in the certificate bundle that is used to verify signing certificate validity

### DIFF
--- a/sigstore/_internal/tuf.py
+++ b/sigstore/_internal/tuf.py
@@ -127,8 +127,8 @@ class TrustUpdater:
         updater.refresh()
         return updater
 
-    def _get(self, usage: str) -> list[bytes]:
-        """Return all active targets with given usage"""
+    def _get(self, usage: str, statuses: list[str]) -> list[bytes]:
+        """Return all targets with given usage and any of the statuses"""
         if not self._updater:
             self._updater = self._setup()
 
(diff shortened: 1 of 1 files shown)


jku:
Oh, forgot to talk about testing:
* I have manually verified that the third  (expired) certificate in the production instance metadata is being downloade and used
* I have not tried verifying a signature that uses that expired certificate because I have not found one I could use

di (approved):

woodruffw:
> * I have not tried verifying a signature that uses that expired certificate because I have not found one I could use

Some of our older `sigstore-python` releases should have a signing cert that's chained up to that older Fulcio cert. I'll check.

di:
I didn't look, but I bet you can find an old-enough cert for https://rekor.tlog.dev/?hash=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 (the shasum of an empty file)

PR #405, opened 2023-01-09
docs: Documentation tweaks

### LINKED ISSUE #404: Remove "This project is not ready for general-purpose use!" prior to 1.0 release.
**Description**
Remove [this line](https://github.com/sigstore/sigstore-python/blame/1a9612a24b0e747d4c59b420b20ee0bb210a2032/README.md#L12) prior to the 1.0 release.

(ref: https://sigstore.slack.com/archives/C024FPJKC6L/p1672741726076749)

woodruffw:
Closes #404.


di (dismissed):

tetsuo-cpp:
@woodruffw From our conversation on Slack:

> make sure we’re fully documenting all members of the sigstore.verify API in the generated HTML

[bdff195](https://github.com/sigstore/sigstore-python/pull/405/commits/bdff19507dfd339be0777e632b7696fa40d2ebe9)

I've added the `models` and `verifier` modules to the `verify` module's `__all__`, so everything is being documented now.

A lot of the types are re-exported in the top-level `verify` module as well. I think the choices are reasonable however, as the types that are re-exported in `verify` are more or less required to use the `verify` API. The parts that aren't re-exported are mainly the `policy` API which makes sense since there are a lot of types there and users aren't going to use most of them.

> experiment with the rendering a bit to see if we can make the code snippets a little nicer

The code snippets that show up when I click "View Source" are formatted nicely and are in a monospace font. And when I create a markdown code block with "```python", it looks similarly good.

I'm assuming you mean the function signatures? They seem to have syntax highlighting but they're in a variable width font. I personally don't see an issue with that and as far as I can remember, that's pretty normal. I believe Doxygen does that by default too.

As for the other two points, I'll follow up with some additional PRs.

tetsuo-cpp (approved):

tetsuo-cpp (approved):

PR #432, opened 2023-01-12
sigstore, test: relax timeout on TUF retrieval

### LINKED ISSUE #431: Relax TUF's timeouts/"slow downloads" check
Just filing this issue from discussion with @jku for tracking purposes.

We've seen some spurious CI failures due to `SlowRetrievalError`s coming from TUF -- the default timeout for stalled network activity seems to be 4s, which is probably a little too short for GitHub Actions (where traffic prioritization/shaping might mean longer delays). 

More generally, I think TUF's default here doesn't play a substantial role in either Sigstore or sigstore-python's security/threat model: it's hard to imagine (possibly due to a lack of imagination on my part) a situation with verification would "fail open" due to a timeout of 4s versus one of, say, 30 seconds. There should still be _some_ timeout that produces a hard error rather than a soft one (e.g. to prevent an attacker from masking repo updates by slowing them down), but 30s seems more reasonable for our use case.

(This needs to be more than just a unit test carveout, because we expect people to use sigstore-python in their CI pipelines.)

woodruffw:
Fixes #431.

### DIFF
--- a/sigstore/_internal/tuf.py
+++ b/sigstore/_internal/tuf.py
@@ -19,12 +19,14 @@ TUF functionality for `sigstore-python`.
 from __future__ import annotations
 
 import logging
+from functools import lru_cache
 from pathlib import Path
 from urllib import parse
 
 import appdirs
 from cryptography.x509 import Certificate, load_pem_x509_certificate
 from tuf.ngclient import Updater
+from tuf.ngclient._internal.requests_fetcher import RequestsFetcher
 
 from sigstore._utils import read_embedded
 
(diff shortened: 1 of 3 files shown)


woodruffw:
cc @jku as well 🙂

sigstore/_internal/tuf.py
```diff
@@ -19,12 +19,14 @@
...
 
 import logging
+from functools import lru_cache
 from pathlib import Path
 from urllib import parse
 
 import appdirs
 from cryptography.x509 import Certificate, load_pem_x509_certificate
 from tuf.ngclient import Updater
+from tuf.ngclient._internal.requests_fetcher import RequestsFetcher
```
woodruffw, line 29: NB: This makes use of an internal TUF API, although I think this is acceptable for now: this internal API shows up in public interfaces and is unlikely to change in the near future; we also may be able to get `python-tuf` to make it into a public API.

di (approved):
Gross, I love it. We should definitely try and upstream a more sane way to configure this.

woodruffw:
> Gross, I love it. We should definitely try and upstream a more sane way to configure this.

Yep! Opened https://github.com/theupdateframework/python-tuf/issues/2268 to track that.

sigstore/_internal/tuf.py
woodruffw: xref https://github.com/theupdateframework/python-tuf/issues/2268

PR #433, opened 2023-01-12
Makefile, ci: test Windows, accomodations

### LINKED ISSUE #417: Makefile: Windows compatibility
This is just a reminder to myself: Windows uses `{env}/Scripts` instead of `{env}/bin` for its virtual environment scripts/binaries, so we should tweak the `Makefile` to specialize for that.

woodruffw:
~~WIP, just seeing how annoying this will be.~~

Closes #417.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# These directories contain TUF and other assets that are either digested
+# or sized-checked so CRLF normalization breaks them.
+sigstore/_store/** binary
+test/unit/assets/** binary
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,16 +16,16 @@ jobs:
     strategy:
       matrix:
         conf:
-          - { py: "3.7", os: "ubuntu-latest" }
-          - { py: "3.8", os: "ubuntu-latest" }
-          - { py: "3.9", os: "ubuntu-latest" }
-          - { py: "3.10", os: "ubuntu-latest" }
-          - { py: "3.11", os: "ubuntu-latest" }
+          # - { py: "3.7", os: "ubuntu-latest" }
+          # - { py: "3.8", os: "ubuntu-latest" }
+          # - { py: "3.9", os: "ubuntu-latest" }
+          # - { py: "3.10", os: "ubuntu-latest" }
+          # - { py: "3.11", os: "ubuntu-latest" }
           # NOTE: We only test Windows and macOS on the latest Python;
           # these primarily exist to ensure that we don't accidentally
           # introduce Linux-isms into the development tooling.
           - { py: "3.11", os: "windows-latest" }
-          - { py: "3.11", os: "macos-latest" }
+          # - { py: "3.11", os: "macos-latest" }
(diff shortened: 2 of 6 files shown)


woodruffw:
Looks like some of the tests are failing on Windows; will investigate.

woodruffw:
This looks like some kind of CRLF weirdness, which doesn't make a lot of sense to me -- the `staging-tuf` test assets should all be LF-only and we should be reading them in binary mode.

woodruffw:
> This looks like some kind of CRLF weirdness, which doesn't make a lot of sense to me -- the `staging-tuf` test assets should all be LF-only and we should be reading them in binary mode.

Confirmed that it's this:

```
    def _fetch(self, url: str) -> Iterator[bytes]:
        filename = os.path.basename(url)
        filepath = _TUF_ASSETS / filename
        if filepath.is_file():
            success[filename] += 1
            content = filepath.read_bytes()
            if filename == "snapshot.json":
>               raise ValueError(len(content))
E               ValueError: 957
```

The actual length is `928`:

```console
$ wc -c test/unit/assets/staging-tuf/snapshot.json 
     928 test/unit/assets/staging-tuf/snapshot.json
```

...however, I still don't understand the "how", unless `read_bytes()` means the exact opposite of what I think it means (return the raw contents, without any platform line normalization).

woodruffw:
This should be good to go, although one of the repo admins is going to need to update the required checks to include the new ones/name changes...

tetsuo-cpp (dismissed):
LGTM!

tetsuo-cpp:
> This should be good to go, although one of the repo admins is going to need to update the required checks to include the new ones/name changes...

@di I think we need you to do this for us. I don't have the permissions for that.

tetsuo-cpp:
> This should be good to go, although one of the repo admins is going to need to update the required checks to include the new ones/name changes...

@di Could you please do this for us? I don't have the permissions for that.

di:
Can we use https://github.com/re-actors/alls-green instead? It's really a pain to make the list of required checks and keep it up to date.

woodruffw:
> Can we use https://github.com/re-actors/alls-green instead? It's really a pain to make the list of required checks and keep it up to date.

Works for me! I can create a PR for that in a little, unless @tetsuo-cpp gets to it first 🙂

tetsuo-cpp:
> Can we use https://github.com/re-actors/alls-green instead? It's really a pain to make the list of required checks and keep it up to date.

Sure, I can make a PR for that now.

woodruffw:
> > Can we use https://github.com/re-actors/alls-green instead? It's really a pain to make the list of required checks and keep it up to date.
> 
> 
> 
> Sure, I can make a PR for that now.
> 
> 
> 
> Edit: Done at #435.

Nice. Let's merge there and update here, and then @di should be able to rid us of the annoying check management once and for all 🙂

woodruffw:
I'll also use this branch to make the `check` names better.

woodruffw:
Looks like the staging instance has degraded a bit more over the past few days...

di (approved):

PR #435, opened 2023-01-13
workflows: Use `alls-green` to maintain required checks

tetsuo-cpp:


woodruffw (approved):

di:
I think we probably want to give these better names so they don't just show up as "check" for each workflow

woodruffw:
> I think we probably want to give these better names so they don't just show up as "check" for each workflow

Doing this in #433.

Issue #375, opened 2022-12-30
verification and signing need to support different sets of keys/certs


jku:
Current TrustUpdater implementation (the tuf component in sigstore-python) always returns keys/certs that have status "Active". Verification should also use status "Expired". I'm not 100% sure if this applies to all target types but I think all certs used in verification at least.

I will have a look at this myself.

jku:
I don't think I can figure this out on my own -- I don't believe the problems are TUF issue as such, I just don't understand how a sigstore client should do these things.

One issue is fulcio cert chains. Current set of certs is:
* fulcio.crt.pem: expired
* fulcio_intermediate_v1.crt.pem: active
* fulcio_v1.crt.pem: active

I assume client is supposed to form cert chains here but what's the logic: how do you select the certs that you put in one chain? I can do that for the active chain (just put all active keys in there) but what about the others?

The other issue is just the multi-key usage: I think this is just not implemented in sigstore-python yet? does cosign have examples already -- like using multiple rekor keys to verify a signature?

jku:
After looking at cosign:
* seems to theoretically support multiple rekor/ctfe keys and actually supports multiple fulcio cert chains... 
* but tuf lookup hard codes a specific key  for the first two so I think they always use a single key for rekor and ctfe
* tuf lookup also hard codes the three certificates for fulcio (that form two chains I believe)
* the fulcio certs are split into two cert pools: root and intermediate and apparently the go x509 module can just sort that out and create cert chains out of them -- I wonder if python cryptography can do that too? I still think it's weird the certificates are not made available as chains

woodruffw:
> EDIT: have not found a definitive answer to the last question but it does seem like we can just stuff all certs we have into a openssl X509Store and it will figure out the chains.

Yep! On the cert side at least, we can punt that complexity to `X509Store` 🙂 

We already support multiple CTFE keys in the form of the `CTKeyring`; we should probably make sure we're fully populating it with every CTFE key we find in the TUF repo.

woodruffw:
For Rekor, I don't think there's been a key rotation yet. But we'll probably want a similar approach to `CTKeyring` (and maybe even just to make that implementation generic).

jku:
So the case for fulcio certs seems clear. This should be easy to fix (and a bit harder to test):
* we should be using both all targets where `usage == "Fulcio" and status in ["Active, "Expired"]` (the bug is that we are currently only using "Active" status). This is the only combination of Fulcio certs we should ever need.
[...]

PR #458, opened 2023-01-17
keyring: CTFE & Rekor shared `Keyring` abstraction

jleightcap:
see #375.

open questions:
- [x] currently a `Keyring` is an abstraction over a `list[bytes]` with specific key types allowed by `isinstance`. should there be a fully abstract `class Keyring(ABC)` derived by distinct `CTKeyring` `RekorKeyring` classes?
- [x] how should the Rekor pubkey be derived from a keyring for for SET verification? more specifically, from `RekorClient.__init__ :: url: str, pubkey: bytes, ct_keyring: CTKeyring` to `RekorClient.__init__ :: url: str, rekor_keyring: Keyring, ct_keyring: Keyring`
- [x] where should `keyring.py` module live?


woodruffw:
> * should there be a fully abstract `class Keyring(ABC)` derived by distinct `CTKeyring` `RekorKeyring` classes?

I don't think we need that strong of an abstraction -- the `Keyring` itself is probably sufficiently generic, since all types of keys we'd like to manage have the same key ID generation mechanism and general format.

woodruffw:
> makes sense. my thinking in suggesting this is that for functions that operate over both/multiiple keyrings, a single `Keyring` type loses type safety at least in passing parameters.

Oh, that's a good point. IMO, a `NewType` here would be sufficient, then -- that'll keep it at the type/mypy level.

sigstore/_internal/ctfe.py
```diff
@@ -16,90 +16,8 @@
...
-    pass
-
-
-class CTKeyringLookupError(CTKeyringError):
-    """
-    A specialization of `CTKeyringError`, indicating that the specified
-    key ID wasn't found in the keyring.
-    """
 
+class CTKeyring(Keyring):
```
woodruffw, line 22: Let's do this with a `NewType` instead:

```python
CTKeyring = NewType("CTKeyring", Keyring)
```

sigstore/_internal/rekor/client.py
```diff
@@ -41,6 +40,10 @@
 STAGING_REKOR_URL = "https://rekor.sigstage.dev"
 
 
+class RekorKeyring(Keyring):
```
woodruffw, line 43: Same as `CTKeyring`: let's use a `NewType`.

woodruffw:
LGTM structurally; needs lint and two small nits.

woodruffw (approved):
LGTM, confirmed that this works as expected. Nice work @jleightcap!

PR #461, opened 2023-01-18
Only run conformance if safe-to-test label is applied

loosebazooka:
Signed-off-by: Appu <appu@google.com>

same with java https://github.com/sigstore/sigstore-java/pull/309


loosebazooka:
jk this isn't necessary.

PR #462, opened 2023-01-18
sigstore: improve JSON response handling

woodruffw:
WIP.

This ratchets down some of our internal handling of JSON responses, with the aim of providing better and more reliable error behavior to end users.


woodruffw:
This is just meant to be a start; there are a few other places where we use `json()` and pull apart the dictionary that could be replaced with a validating model.

Issue #251, opened 2022-10-14
Support generating and verifying with "Sigstore" bundles


woodruffw:
There are currently (at least) 3 different things called "bundles" in the Sigstore ecosystem:

* "cosign bundles", which `cosign --bundle` emits. These contain a Rekor entry, plus checksum, certificate, and signature needed to perform a normal verification step.
* "Rekor offline bundles", which contain *just* the Rekor entry and its SET. These need to be combined with separate inputs to perform a normal verification step.
* "sigstore bundles", which are currently being designed (https://github.com/sigstore/cosign/pull/2204). These will deprecate "cosign bundles" and perform the same function.

Once the "Sigstore bundle" format is stabilized, `sigstore-python` should both consume and emit it (by default, rather than emitting/loading separate files for each component).

This is a counterpart to #52 and #194, and will obviate/deprecate the work in #247 once finished.

Subtasks:

- [x] Generate Sigstore-style bundles while signing with `sigstore sign` (#465)
- [ ] Verify using Sigstore-style bundles with the `sigstore verify` subcommands

di:
Marking this blocked on https://github.com/sigstore/cosign/issues/2131

woodruffw:
The sigstore bundle spec is being developed in `sigstore/protobuf-specs`; tracking https://github.com/sigstore/protobuf-specs/issues/6

jku:
nothing against protobufs as such but it'll likely be yet another dependency just to read a single config file  :(

woodruffw:
> nothing against protobufs as such but it'll likely be yet another dependency just to read a single config file :(

Yeah, this isn't ideal (but also not the end of the world). My plan here was to look into generating a separate Python package from the protobuf specs, something like `sigstore-protos` or `sigstore-models`, so that we can avoid a direct protobuf dependency.

di:
Or we could have the repo that defines the specs generate/publish the protos package as well.

woodruffw:
Bumping this to post-stable, since the Sigstore bundle format itself isn't fully stabilized.

woodruffw:
Unblocking because a `0.1` release of the bundle format is now available; we should begin evaluating against it.

tetsuo-cpp:
I'm going to prepare a PR for verifying bundles.

woodruffw:
Done as of #478!

edgarrmondragon:
Not a big deal, but should the subtask

[...]

PR #465, opened 2023-01-20
Initial Sigstore bundle support

woodruffw:
This adds initial support for Sigstore bundles during signing, in the form of the `--bundle` flag. When passed, `sigstore sign` will generate a single `{input}.sigstore` instead of separate `.crt`, `.sig`, and `.rekor` files.

I haven't included `sigstore verify` support in the initial changeset, in an effort to keep the diff small. But adding it shouldn't be difficult, and will provide a good dogfood/smoke test for round-tripping through the bundle format.

See #251.

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -652,8 +652,8 @@ def _sign(args: argparse.Namespace) -> None:
             if not args.no_bundle:
                 bundle = file.parent / f"{file.name}.sigstore"
 
-        extants = []
         if not args.overwrite:
+            extants = []
             if sig and sig.exists():
                 extants.append(str(sig))
             if cert and cert.exists():
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -587,14 +587,14 @@ def _sign(args: argparse.Namespace) -> None:
 
     if args.bundle:
         logger.warning(
-            "--bundle support is experimental; this flag may change behaviour "
-            "between releases until stabilized or may be removed."
+            "--bundle support is experimental; the behaviour of this flag may change "
+            "between releases until stabilized."
         )
 
     if args.no_bundle:
         logger.warning(
-            "--no-bundle support is experimental; this flag may change behaviour "
-            "between releases until stabilized or may be removed."
+            "--no-bundle support is experimental; the behaviour of this flag may change "
+            "between releases until stabilized."
         )
 
     # `--no-default-files` has no effect on `--{signature,certificate,rekor-bundle,bundle}`,
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -187,11 +187,12 @@ def _add_shared_verify_input_options(group: argparse._ArgumentGroup) -> None:
     )
     group.add_argument(
         "--bundle",
-        action="store_true",
-        default=_boolify_env("SIGSTORE_BUNDLE"),
+        metavar="FILE",
+        type=Path,
+        default=os.getenv("SIGSTORE_BUNDLE"),
         help=(
-            "Verify from {input}.sigstore for each input; this option is experimental "
-            "and may change between releases until stabilized"
+            "The Sigstore bundle to verify with; not used with multiple inputs; this option is "
+            "experimental and may change between releases until stabilized"
         ),
     )
     group.add_argument(
@@ -351,10 +352,20 @@ def _parser() -> argparse.ArgumentParser:
     )
     output_options.add_argument(
         "--bundle",
-        action="store_true",
-        default=_boolify_env("SIGSTORE_BUNDLE"),
+        metavar="FILE",
+        type=Path,
+        default=os.getenv("SIGSTORE_BUNDLE"),
         help=(
-            "Emit a single {input}.sigstore file for each input; this option is experimental "
+            "Write a single Sigstore bundle to the given file; does not work with multiple input "
+            "files; this option is experimental and may change between releases until stabilized"
+        ),
+    )
+    output_options.add_argument(
+        "--no-bundle",
+        action="store_true",
+        default=False,
+        help=(
+            "Don't emit {input}.sigstore files for each input; this option is experimental "
             "and may change between releases until stabilized"
         ),
     )
@@ -574,6 +585,18 @@ def _sign(args: argparse.Namespace) -> None:
             "upcoming release of sigstore-python in favor of Sigstore-style bundles"
         )
 
+    if args.bundle:
+        logger.warning(
+            "--bundle support is experimental; this flag may change behaviour "
+            "between releases until stabilized or may be removed."
+        )
+
+    if args.no_bundle:
+        logger.warning(
+            "--no-bundle support is experimental; this flag may change behaviour "
+            "between releases until stabilized or may be removed."
+        )
+
     # `--no-default-files` has no effect on `--{signature,certificate,rekor-bundle,bundle}`,
     # but we forbid it because it indicates user confusion.
     if args.no_default_files and (
@@ -589,6 +612,10 @@ def _sign(args: argparse.Namespace) -> None:
     if args.rekor_bundle and args.bundle:
         args._parser.error("--rekor-bundle may not be combined with --bundle")
 
+    # Fail if `--bundle` and `--no-bundle` are both specified.
+    if args.bundle and args.no_bundle:
+        args._parser.error("--bundle may not be combined with --no-bundle")
+
     # Fail if `--signature` or `--certificate` is specified *and* we have more
     # than one input.
     if (args.signature or args.certificate or args.rekor_bundle) and len(
@@ -602,50 +629,52 @@ def _sign(args: argparse.Namespace) -> None:
     # Build up the map of inputs -> outputs ahead of any signing operations,
     # so that we can fail early if overwriting without `--overwrite`.
     output_map = {}
-    extants = []
     for file in args.files:
         if not file.is_file():
             args._parser.error(f"Input must be a file: {file}")
 
-        if args.bundle:
-            logger.warning(
-                "--bundle support is experimental; the behavior of this flag may change "
-                "between releases until stabilized."
-            )
+        sig, cert, rekor_bundle, bundle = (
+            args.signature,
+            args.certificate,
+            args.rekor_bundle,
+            args.bundle,
+        )
+        if (
+            not sig
+            and not cert
+            and not rekor_bundle
+            and not bundle
+            and not args.no_default_files
+        ):
+            sig = file.parent / f"{file.name}.sig"
+            cert = file.parent / f"{file.name}.crt"
+            rekor_bundle = file.parent / f"{file.name}.rekor"
+            if not args.no_bundle:
+                bundle = file.parent / f"{file.name}.sigstore"
 
-            bundle = file.parent / f"{file.name}.sigstore"
-
-            if bundle.exists():
+        extants = []
+        if not args.overwrite:
+            if sig and sig.exists():
+                extants.append(str(sig))
+            if cert and cert.exists():
+                extants.append(str(cert))
+            if rekor_bundle and rekor_bundle.exists():
+                extants.append(str(rekor_bundle))
+            if bundle and bundle.exists():
                 extants.append(str(bundle))
 
-            output_map[file] = {"bundle": bundle}
-        else:
-            sig, cert, rekor_bundle = (
-                args.signature,
-                args.certificate,
-                args.rekor_bundle,
-            )
-
-            if not sig and not cert and not rekor_bundle and not args.no_default_files:
-                sig = file.parent / f"{file.name}.sig"
-                cert = file.parent / f"{file.name}.crt"
-                rekor_bundle = file.parent / f"{file.name}.rekor"
-
-                if sig and sig.exists():
-                    extants.append(str(sig))
-                if cert and cert.exists():
-                    extants.append(str(cert))
-                if rekor_bundle and rekor_bundle.exists():
-                    extants.append(str(rekor_bundle))
-
-            output_map[file] = {"cert": cert, "sig": sig, "rekor_bundle": rekor_bundle}
-
-        if not args.overwrite and len(extants) > 0:
             if extants:
                 args._parser.error(
                     "Refusing to overwrite outputs without --overwrite: "
                     f"{', '.join(extants)}"
                 )
+
+        output_map[file] = {
+            "cert": cert,
+            "sig": sig,
+            "rekor_bundle": rekor_bundle,
+            "bundle": bundle,
+        }
 
     # Select the signer to use.
     if args.staging:
@@ -696,27 +725,27 @@ def _sign(args: argparse.Namespace) -> None:
         print(f"Transparency log entry created at index: {result.log_entry.log_index}")
 
         sig_output: TextIO
-        if "sig" in outputs:
+        if outputs["sig"] in outputs:
             sig_output = outputs["sig"].open("w")
         else:
             sig_output = sys.stdout
 
         print(result.b64_signature, file=sig_output)
-        if "sig" in outputs:
+        if outputs["sig"] is not None:
             print(f"Signature written to {outputs['sig']}")
 
-        if "cert" in outputs:
+        if outputs["cert"] is not None:
             with outputs["cert"].open(mode="w") as io:
                 print(result.cert_pem, file=io)
             print(f"Certificate written to {outputs['cert']}")
 
-        if "rekor_bundle" in outputs:
+        if outputs["rekor_bundle"] is not None:
             with outputs["rekor_bundle"].open(mode="w") as io:
                 rekor_bundle = RekorBundle.from_entry(result.log_entry)
                 print(rekor_bundle.json(by_alias=True), file=io)
             print(f"Rekor bundle written to {outputs['rekor_bundle']}")
 
-        if "bundle" in outputs:
+        if outputs["bundle"] is not None:
             with outputs["bundle"].open(mode="w") as io:
                 print(result._to_bundle().to_json(), file=io)
             print(f"Sigstore bundle written to {outputs['bundle']}")
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -725,7 +725,7 @@ def _sign(args: argparse.Namespace) -> None:
         print(f"Transparency log entry created at index: {result.log_entry.log_index}")
 
         sig_output: TextIO
-        if "sig" in outputs:
+        if outputs["sig"] is not None:
             sig_output = outputs["sig"].open("w")
         else:
             sig_output = sys.stdout
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -159,7 +159,7 @@ def _add_shared_instance_options(group: argparse._ArgumentGroup) -> None:
     )
 
 
-def _add_shared_input_options(group: argparse._ArgumentGroup) -> None:
+def _add_shared_verify_input_options(group: argparse._ArgumentGroup) -> None:
     """
     Common input options, shared between all `sigstore verify` subcommands.
     """
@@ -184,6 +184,15 @@ def _add_shared_input_options(group: argparse._ArgumentGroup) -> None:
         type=Path,
         default=os.getenv("SIGSTORE_REKOR_BUNDLE"),
         help="The offline Rekor bundle to verify with; not used with multiple inputs",
+    )
+    group.add_argument(
+        "--bundle",
+        action="store_true",
+        default=_boolify_env("SIGSTORE_BUNDLE"),
+        help=(
+            "Verify from {input}.sigstore for each input; this option is experimental "
+            "and may change between releases until stabilized"
+        ),
     )
     group.add_argument(
         "files",
@@ -341,6 +350,15 @@ def _parser() -> argparse.ArgumentParser:
         ),
     )
     output_options.add_argument(
+        "--bundle",
+        action="store_true",
+        default=_boolify_env("SIGSTORE_BUNDLE"),
+        help=(
+            "Emit a single {input}.sigstore file for each input; this option is experimental "
+            "and may change between releases until stabilized"
+        ),
+    )
+    output_options.add_argument(
         "--overwrite",
         action="store_true",
         default=_boolify_env("SIGSTORE_OVERWRITE"),
@@ -387,7 +405,7 @@ def _parser() -> argparse.ArgumentParser:
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
     input_options = verify_identity.add_argument_group("Verification inputs")
-    _add_shared_input_options(input_options)
+    _add_shared_verify_input_options(input_options)
 
     verification_options = verify_identity.add_argument_group("Verification options")
     _add_shared_verification_options(verification_options)
@@ -420,7 +438,7 @@ def _parser() -> argparse.ArgumentParser:
     )
 
     input_options = verify_github.add_argument_group("Verification inputs")
-    _add_shared_input_options(input_options)
+    _add_shared_verify_input_options(input_options)
 
     verification_options = verify_github.add_argument_group("Verification options")
     _add_shared_verification_options(verification_options)
@@ -556,15 +574,20 @@ def _sign(args: argparse.Namespace) -> None:
             "upcoming release of sigstore-python in favor of Sigstore-style bundles"
         )
 
-    # `--no-default-files` has no effect on `--{signature,certificate,rekor-bundle}`, but we
-    # forbid it because it indicates user confusion.
+    # `--no-default-files` has no effect on `--{signature,certificate,rekor-bundle,bundle}`,
+    # but we forbid it because it indicates user confusion.
     if args.no_default_files and (
-        args.signature or args.certificate or args.rekor_bundle
+        args.signature or args.certificate or args.rekor_bundle or args.bundle
     ):
         args._parser.error(
             "--no-default-files may not be combined with --signature, "
-            "--certificate, or --rekor-bundle",
+            "--certificate, --rekor-bundle, or --bundle",
         )
+
+    # Similarly forbid `--rekor-bundle` with `--bundle`, since it again indicates
+    # user confusion around outputs.
+    if args.rekor_bundle and args.bundle:
+        args._parser.error("--rekor-bundle may not be combined with --bundle")
 
     # Fail if `--signature` or `--certificate` is specified *and* we have more
     # than one input.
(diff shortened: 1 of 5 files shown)


woodruffw:
Example generated bundle:

```json
{
    "mediaType": "application/vnd.dev.sigstore.bundle+json;version=0.1",
    "verificationMaterial": {
        "x509CertificateChain": {
            "certificates": [
                {
                    "rawBytes": "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUN3akNDQWtlZ0F3SUJBZ0lVU0NpWUVuTzVLa0g0YWxaNTlDNmEyMUVwUENZd0NnWUlLb1pJemowRUF3TXcKTnpFVk1CTUdBMVVFQ2hNTWMybG5jM1J2Y21VdVpHVjJNUjR3SEFZRFZRUURFeFZ6YVdkemRHOXlaUzFwYm5SbApjbTFsWkdsaGRHVXdIaGNOTWpNd01USXdNVFkxTkRBeVdoY05Nak13TVRJd01UY3dOREF5V2pBQU1IWXdFQVlICktvWkl6ajBDQVFZRks0RUVBQ0lEWWdBRW5RRUZsMmUzdzdySjQ5eHdGbWVheWN3ZnZTajBMWXB1WFZ1d3R0aE4KaTF3TWY0ZGF0SGNSeFpMN3ZTaXp1VHY5M2twZEZ5UTBhOGJYeTBTYUpia2NBUFVEVDFTY05VVkZwOEVOOTdTTApBbmJ2dzJLRlk1YjF4QmtwR0VwajMwNmVvNElCU1RDQ0FVVXdEZ1lEVlIwUEFRSC9CQVFEQWdlQU1CTUdBMVVkCkpRUU1NQW9HQ0NzR0FRVUZCd01ETUIwR0ExVWREZ1FXQkJUdFpicTBIYmk5RWc3eThIaXZQKzFvV2NNNzh6QWYKQmdOVkhTTUVHREFXZ0JSeGhqQ21GSHhpYi9uMzF2UUZHbjlmLyt0dnJEQWpCZ05WSFJFQkFmOEVHVEFYZ1JWMwphV3hzYVdGdFFIbHZjM05oY21saGJpNXVaWFF3TEFZS0t3WUJCQUdEdnpBQkFRUWVhSFIwY0hNNkx5OW5hWFJvCmRXSXVZMjl0TDJ4dloybHVMMjloZFhSb01JR0tCZ29yQmdFRUFkWjVBZ1FDQkh3RWVnQjRBSFlBS3pDODNHaUkKeWVMaDJDWXBYblFmU0RreGxnTHluRFBMWGtOQS9yS3Nobm9BQUFHRjBCbitCZ0FBQkFNQVJ6QkZBaUJ6TENLbwpOUzlqS01iaHYwU2FteGdoWUJQK3RlbExGRzlYcDBBVXlQQmlpUUloQUxXTXpzU1pBaVdYZHN6S0xXQU15b01LClBNcjVOWndTM1Q4SmE0ZFhWbUJWTUFvR0NDcUdTTTQ5QkFNREEya0FNR1lDTVFDNGU1cktGcWRHbEZkTzJEemkKR3k3ajZJVkpZdENhQStraitUL0MwVWhwZ0xON2h2dGVXOXdkSUwrU1l0bjRvVG9DTVFDMmswNDVFdktBcTdHRwpEbngwaEhrTWJmS2ZKK0NCZjV1bjZSODkxOWpST1ZVdUdEeUlLUWhFSi9RQjBhaVV3bDg9Ci0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K"
                }
            ]
        },
        "tlogEntries": [
            {
                "logIndex": "2242969",
                "logId": {
                    "keyId": "0y8wo8MtY5wrdiIFohx7sHeI5oKDpK5vQhGHI6G+pJY="
                },
                "kindVersion": {
                    "kind": "hashedrekord",
                    "version": "0.0.1"
                },
                "integratedTime": "1674233642",
                "inclusionPromise": {
                    "signedEntryTimestamp": "MEUCIQD31VsczBNtApPFNW31nn20hzxULvVCtLTe7BnAjJCQiwIgN4RlYRkfRSdC7A8E6FufttvH53L5GCE+gmuh2TKVx+I="
                },
                "inclusionProof": {
                    "logIndex": "2228150",
                    "rootHash": "sxhOFoisbFXhJmKOSlimQWBpgkrDT5eCaSaUcydRGEg=",
                    "treeSize": "2228151",
                    "hashes": [
                        "TwqHCU5ELTQy/KKeu+0+Qlnjr0zzPziKXjE5TGXeRT0=",
                        "HtGiDzArXW02VJtZKJVYEIW1cyB7jR0d9c60io54x8s=",
                        "Wjgnj5IgRwHKGVnpu6DKjZWP2ZDRyIvXC2OKO437mQ8=",
                        "ZXWmnfXIqBNf4a2+MosY2T0zeCGACcqlYxrYXdY5INU=",
                        "rkq9/Kqf8fYE4vhLb5vzhhbcsI5kT2JNY06JZmLMlC0=",
                        "O8GhxttbaSwc3hOClnMY2tliEqlpXiUKhuUKPbPYZbA=",
                        "UaBSu2yb0gg+mg7gWVuQViCWJroP3wAo9sm+70DtT94=",
                        "hbzBh170Y7wu6HE4gFaJX1m3tcnffbIbneNVNitxEQ8=",
                        "7RquioMcFFcBwp8L9WgAYvNdmFIX2PLFRO3Ig3I4aG4=",
                        "sEdlyfBp6l2RMcB1BIr7PJyI8d6jWYNBbz4EO0ocBH0=",
                        "PbBHHJ1PoLFq1tjP/Z4nE7yCxtIxUhtqkSdpbkLmrE4=",
                        "d8Hx9AnJ483YnmugSNyUJyzV+dWJfK4wEcn9d9Q+4hM=",
                        "KCciPEwSUhXM04e8244YynDhwh+722/pabo6ZAtt0fc=",
                        "cfT2nw0gal6i8QktNkaFsq6w3Aeu62pnB6sDRzKkAZw=",
                        "VwBj5hN1tw74kRJeHAQaqdSWrXWk7Zb4c1PJfrpiKNw="
                    ]
                },
                "canonicalizedBody": "eyJib2R5IjoiZXlKaGNHbFdaWEp6YVc5dUlqb2lNQzR3TGpFaUxDSnJhVzVrSWpvaWFHRnphR1ZrY21WcmIzSmtJaXdpYzNCbFl5STZleUprWVhSaElqcDdJbWhoYzJnaU9uc2lZV3huYjNKcGRHaHRJam9pYzJoaE1qVTJJaXdpZG1Gc2RXVWlPaUl5TVdKbU9XRmxNR1F5WkRsaFpETmlOMlppTmpSbU1tUTJNV0prTWpFeU1XSXhZemMxWldWaU1UaGlNekJqWmpjeVptRm1NVEV3TW1aa01tRm1ZMlkzSW4xOUxDSnphV2R1WVhSMWNtVWlPbnNpWTI5dWRHVnVkQ0k2SWsxSFdVTk5VVU5IZEdnck1tWjRlRVkxZWpWNVIwVTVaSEZrVVhwWFFXTTJWMFoxVUhCV1ExcEljakF6UTJSM2F6UXpaV2RRVmpWRU9GaE9Wak51WjA5dlJqUkdWMWx2UTAxUlEwd3hkVFZTWVhGTUwwUXlSMGw0ZUdOQ2JHTkVjREJMZEhVcllUQTVhVlZMYlRCUlFVZDFVR1pLTUVZeGVraDFRMmxTVWxndk0zWXdLMWcxVTNOcFNWVTlJaXdpY0hWaWJHbGpTMlY1SWpwN0ltTnZiblJsYm5RaU9pSk1VekIwVEZNeFExSlZaRXBVYVVKRVVsWktWVk5WV2twUk1FWlZVbE13ZEV4VE1IUkRhekZLVTFWT00yRnJUa1JSVjNSc1dqQkdNMU5WU2tKYU1HeFdWVEJPY0ZkVlZuVlVlbFpNWVRCbk1GbFhlR0ZPVkd4RVRtMUZlVTFWVm5kVlJVNWFaREJPYmxkVmJFeGlNWEJLWlcxdmQxSlZSak5VV0dOTFZHNXdSbFpyTVVOVVZXUkNUVlpXUmxFeWFFNVVWMDE1WWtjMWFrMHhTakpaTWpGV1pGWndTRlpxU2s1VmFsSXpVMFZHV2xKR1dsSlZWVkpHWlVaYU5sbFdaR3RsYlZKSVQxaHNZVlY2Um5kWmJUVlRZa0Z3YW1KVVJuTlhhMlJ6WVVkU1NGWllaRWxoUjA1UFZGZHdUbVF3TVZWVFdHUk9Wa1pyZUZSclVrSmxWbVJ2V1RBMVRtRnJNVE5VVmxKS1pEQXhWVmt6WkU5U1JVWTFWakp3UWxGVk1VbFhXR1JHVVZac1NVTnJkSFpYYTJ3MllXcENSRkZXUmxwU2EzTXdVbFZXUWxFd2JFVlhWMlJDVWxjMVVsSlZXbk5OYlZWNlpIcGtlVk5xVVRWbFNHUkhZbGRXYUdWWFRqTmFibHBVWVdwQ1RWZFlRakZYUmxveFpETlNNR0ZGTkV0aFZFWXpWRmRaTUZwSFJqQlRSMDVUWlVad1RVNHpXbFJoV0hBeFZraFpOVTB5ZEhkYVJWbzFWVlJDYUU5SFNsbGxWRUpVV1ZWd2FXRXlUa0pWUmxaRlZrUkdWRmt3TlZaV2ExcDNUMFZXVDA5VVpGUlVRWEJDWW0xS01tUjZTa3hTYkdzeFdXcEdORkZ0ZEhkU01GWjNZV3BOZDA1dFZuWk9SV3hEVlRGU1JGRXdSbFpXV0dSRldqRnNSVlpzU1hkVlJVWlNVME01UTFGV1JrVlJWMlJzVVZVeFExUlZaRUpOVmxaclEydHdVbFZWTVU1UlZ6bElVVEJPZWxJd1JsSldWVnBEWkRBeFJWUlZTWGRTTUVWNFZsZFNSVm94UmxoUmEwcFZaRVp3YVdOVVFrbFpiV3MxVWxkak0yVlVhRWxoV0ZwUlMzcEdkbFl5VGs1T2VtZzJVVmRaUzFGdFpFOVdhMmhVVkZWV1NGSkZSbGhhTUVwVFpVZG9jVkV5TVVkVFNHaHdXV2s1ZFUxNlJqSlZWVnBJWW1wc2JVeDVkREJrYmtwRlVWZHdRMW93TlZkVFJrcEdVV3RHYlU5RlZraFdSVVpaV2pGS1YwMTNjR2hXTTJoNldWWmtSMlJHUmtsaVNGcHFUVEExYjFreU1YTmhSMHB3VGxoV1lWZEdSak5VUlVaYVV6QjBNMWRWU2tOUlZXUkZaRzV3UWxGclJsSlZWMVpvVTBaSmQxa3dhRTVPYTNnMVQxYzFhRmRHU25aRGJWSllVMWhXV2sxcWJEQlVSRW8wWkd4dmVXSklWazFOYW14dldrWm9VMkl3TVVwU01IUkRXakk1ZVZGdFpFWlNWVVpyVjJwV1Fsb3hSa1JSYTJnelVsZFdibEZxVWtKVFJteENVek53UkU5RVRraGhWV3RMWlZkV1RXRkVTa1JYV0VKWllteEdiVlV3VW5KbFIzaHVWRWhzZFZKR1FrMVhSM1JQVVZNNWVWTXpUbTlpYlRsQ1VWVkdTRkpxUWtOaWFYUkRXakJHUWxGclJrNVJWa28yVVd0YVFtRlZTalpVUlU1TVluZHdUMVY2YkhGVE1ERnBZVWhaZDFVeVJuUmxSMlJ2VjFWS1VVc3pVbXhpUlhoSFVucHNXV05FUWtKV1dHeFJVVzFzY0ZWVmJHOVJWWGhZVkZod2VsVXhjRUpoVm1SWldraE9ObE13ZUZoUlZURTFZakF4VEVOc1FrNWphbFpQVjI1a1ZFMHhVVFJUYlVVd1drWm9WMkpWU2xkVVZVWjJVakJPUkdOVlpGUlVWRkUxVVd0R1RsSkZSWGxoTUVaT1VqRnNSRlJXUmtST1IxVXhZMnQwUjJOWFVraGlSVnByVkhwS1JXVnRhMHRTTTJzellXcGFTbFpyY0Zwa1JVNW9VVk4wY21GcGRGVk1NRTEzVmxkb2Qxb3dlRTlPTW1neVpFZFdXRTlZWkd0VFZYZHlWVEZzTUdKcVVuWldSemxFVkZaR1JFMXRjM2RPUkZaR1pHdDBRbU5VWkVoU2QzQkZZbTVuZDJGRmFISlVWMHB0VXpKYVMwc3dUa05hYWxZeFltcGFVMDlFYTNoUFYzQlRWREZhVm1SVlpFVmxWV3hNVlZkb1JsTnBPVkpSYWtKb1lWWldNMkpFWnpsRGFUQjBURk13ZEZKVk5VVkpSVTVHVld4U1NsSnJiRVJSVmxKR1RGTXdkRXhUTUVzaWZYMTlmUT09IiwiaW50ZWdyYXRlZFRpbWUiOjE2NzQyMzM2NDIsImxvZ0lEIjoiZDMyZjMwYTNjMzJkNjM5YzJiNzYyMjA1YTIxYzdiYjA3Nzg4ZTY4MjgzYTRhZTZmNDIxMTg3MjNhMWJlYTQ5NiIsImxvZ0luZGV4IjoyMjQyOTY5fQ=="
            }
        ]
    },
    "messageSignature": {
        "messageDigest": {
            "algorithm": "SHA2_256",
            "digest": "Ib+a4NLZrTt/tk8tYb0hIbHHXusYswz3L68RAv0q/Pc="
        },
        "signature": "MGYCMQCGth+2fxxF5z5yGE9dqdQzWAc6WFuPpVCZHr03Cdwk43egPV5D8XNV3ngOoF4FWYoCMQCL1u5RaqL/D2GIxxcBlcDp0Ktu+a09iUKm0QAGuPfJ0F1zHuCiRRX/3v0+X5SsiIU="
    }
}
```

That bundle signs for `README.md` as of [a0377a4](https://github.com/sigstore/sigstore-python/pull/465/commits/a0377a4f4bfc04c96e49ed1950826b27cfdaca5e).

woodruffw:
cc @znewman01 and @kommendorkapten for visibility 🙂

sigstore/sign.py
```diff
@@ -188,3 +212,62 @@ class SigningResult(BaseModel):
...
+                ],
+            )
+
+        tlog_entry = TransparencyLogEntry(
+            log_index=self.log_entry.log_index,
+            log_id=LogId(key_id=bytes.fromhex(self.log_entry.log_id)),
+            # TODO: Maybe leave this field out? It appears to be optional
+            # and it's just hardcoded for us, since it's the only kind of Rekor
+            # entry we support.
+            kind_version=KindVersion(kind="hashedrekord", version="0.0.1"),
```
woodruffw, line 244: cc @kommendorkapten do you have opinions on this? `kind_version` isn't marked as mandatory and the information it provides is probably duplicated in the log entry body itself, so I wasn't sure if it makes sense to include it here.

sigstore/sign.py
```diff
@@ -188,3 +212,62 @@ class SigningResult(BaseModel):
...
+            integrated_time=self.log_entry.integrated_time,
+            inclusion_promise=InclusionPromise(
+                signed_entry_timestamp=base64.b64decode(
+                    self.log_entry.signed_entry_timestamp
+                )
+            ),
+            inclusion_proof=inclusion_proof,
+            # TODO: Is this correct?
+            canonicalized_body=self.log_entry.encode_canonical(),
```
woodruffw, line 253: Flagging for review: I _think_ this is right, but I'm not 100% sure -- our `encode_canonical` produces a bytestring of the canonicalized JSON for the log entry, which _should_ be the same thing that Rekor's REST API returns.

sigstore/sign.py
```diff
@@ -188,3 +212,62 @@ class SigningResult(BaseModel):
...
+    def _to_bundle(self) -> Bundle:
+        """
+        Creates a Sigstore bundle (as defined by Sigstore's protobuf specs)
+        from this `SigningResult`.
+        """
+
+        # TODO: Should we include our Fulcio intermediate in the chain?
+        chain = X509CertificateChain(
+            certificates=[X509Certificate(raw_bytes=self.cert_pem.encode())]
+        )
```
woodruffw, line 225: Flagging for review: I wasn't sure about this either. During discussions in protobuf-specs it sounds like the root cert isn't expected to be included, but that clients *may* include any intermediates in use.

woodruffw:
Currently evaluating against `sigstore-js`; blocked on resolution of https://github.com/sigstore/protobuf-specs/issues/62

sigstore/sign.py
woodruffw: `sigstore-js` doesn't seem to have problems verifying even without us embedding the intermediate, so I think this is probably fine for now.

sigstore/sign.py
```diff
@@ -188,3 +212,61 @@ class SigningResult(BaseModel):
     """
     A record of the Rekor log entry for the signing operation.
     """
+
+    def _to_bundle(self) -> Bundle:
```
woodruffw, line 216: This can wait until a follow-up, but the amount of round-tripping here (`str -> bytes`, base64 decoding, etc) suggests that the current `SigningResult` field types aren't ideal 🙂 

They're already part of the public API, however, so any internal changes to them will probably need backwards-compatible properties.

sigstore/_cli.py
```diff
@@ -340,6 +349,15 @@ def _parser() -> argparse.ArgumentParser:
             "multiple input files"
         ),
     )
+    output_options.add_argument(
```
znewman01, line 352: My instinct would be to make this take a path

(But totally up to you!)

sigstore/sign.py
znewman01: +1 to the sentiment, but I understand the need for compatibility :)

sigstore/sign.py
```diff
@@ -188,3 +212,61 @@ class SigningResult(BaseModel):
...
+        """
+        Creates a Sigstore bundle (as defined by Sigstore's protobuf specs)
+        from this `SigningResult`.
+        """
+
+        # TODO: Should we include our Fulcio intermediate in the chain?
```
znewman01, line 222: I think so

https://github.com/sigstore/protobuf-specs/blob/16541696de137c6281d66d075a4924d9bbd181ff/protos/sigstore_common.proto#L170-L173

sigstore/sign.py
znewman01: See also https://github.com/sigstore/protobuf-specs/issues/52

If I'm reading correctly we should keep it here because it may disappear from the log entry later.

znewman01 (dismissed):
Overall looks great

sigstore/sign.py
znewman01: Oh, resolve this comment we can follow up in the thread you opened already

sigstore/sign.py
znewman01: I think you're supposed to include it based on the docs?

https://github.com/sigstore/protobuf-specs/blob/16541696de137c6281d66d075a4924d9bbd181ff/protos/sigstore_common.proto#L170-L173

CC @haydentherapper  who will have a strong opinion here

woodruffw: Hmm, my reading of the language there:

```
        // The first certificate in the array must be the leaf
        // certificate used for signing. Any intermediate certificates
        // must be stored as offset 1 to n-1, and the root certificate at
        // position n.
```

Is that the certificate list MUST be nonempty, the first member MUST be the leaf, and *any if present* intermediates and root MUST be listed in chain order. In other words, just the leaf is fine, but @haydentherapper can confirm!

sigstore/_cli.py
woodruffw: Yeah, I need to think a bit about this -- `sigstore sign` can take multiple inputs, so `sigstore sign foo.txt bar.txt --bundle foo.txt` doesn't fit quite right :slightly_smiling_face: 

I'll probably do a variant of what I do for the other output flags: `--bundle` will actually enable the bundle output format, and then maybe `--sigstore-bundle <foo>` will actually write to `<fOo>` if exactly one input is provided. But that isn't great for UX either :upside_down_face:

sigstore/sign.py
woodruffw: Makes sense! In that case, am I right in thinking that `kind_version` should probably be marked as mandatory in the protobuf definitions?

sigstore/sign.py
```diff
@@ -188,3 +212,61 @@ class SigningResult(BaseModel):
...
+            canonicalized_body=base64.b64decode(self.log_entry.body),
+        )
+
+        material = VerificationMaterial(
+            x509_certificate_chain=chain,
+            tlog_entries=[tlog_entry],
+        )
+
+        bundle = Bundle(
+            media_type="application/vnd.dev.sigstore.bundle+json;version=0.1",
```
woodruffw, line 261: It'd be nice if there was some way to default to this `media_type`, but I have no idea if that's possible :slightly_smiling_face:

sigstore/_cli.py
znewman01: Ahh interesting...yeah, hard to find anything that makes sense here.

I wonder whether you need two verbs, a `sign-one` and a `sign-many`.

Happy to let it stand for now

sigstore/sign.py
znewman01: IIUC no :(

That said, I think that's sort-of by design? You should only set that `media_type` if you immediately plan to serialize to JSON

sigstore/sign.py
znewman01: Yeah, we could stand to have an additional clarifying comment for `verification_material`.

sigstore/sign.py
woodruffw: Aha, I missed that detail! That makes sense, then, and it's not a huge deal here.

sigstore/sign.py
kommendorkapten: Thanks for the link @znewman01 !

Yes, that should be mandatory in the spec. It's being added as of this: https://github.com/sigstore/protobuf-specs/pull/64/files

sigstore/sign.py
kommendorkapten: The intent was that the leaf certificate MUST be included. Any intermediate (if existing), the and the root are RECOMMENDED to be included. 

Regardless if they are included or not in the bundle. The intermediates and the root must of course be verified against the trust root, as this is the only way to securely verify that they are trusted. The reason for not only depend on the root certificate, but also the intermediate is that we don't ship any CRLs, if an intermediate is revoed, it's removed from the TUF root, and so a client MUST NOT trust it.

kommendorkapten: But agree, lets clarify this. The validation part is covered by the client spec, but I think the language is a bit unclear there too.

sigstore/sign.py
kommendorkapten: Yeah, it would be nice to have const values in the language bindings that would capture the current media type used.

woodruffw: 100% agreed -- I'm not sure how best to accomplish that without hacky patches to the codegen, though...

sigstore/sign.py
haydentherapper: Historical context from Web PKI: Root CAs that follow CA/B forum requirements (typically those that are a part of OS or Mozilla trust stores) have their signing keys offline, and have long lifetimes (10-20 years). Intermediate CAs are created with online signing keys and shorter lifetimes (~3 years) to issue end-entity (leaf) certificates. Because these signing keys are online, there's a higher chance of compromise. In the event of a compromise, a root CA maintains a CRL containing intermediate CA certs that clients are expected to check for intermediate CA trustworthiness.

TLS connections provide the full cert chain, because intermediates are meant only for chain building, to find a path from an end-entity certificate to a root. The root is provided I assume just for auditing purposes, but a verifier should never trust a root provided in a chain, it should come from a trust store. The intermediates build a chain from the leaf to the root, with each intermediate's status being checked against a CRL.

Back to the present: The Sigstore public good instance's root CA is online, because maintaining offline CAs is a pain, especially with this model of cross-company production ownership. We could have skipped using an intermediate and issued end-entity certs directly off of the root, but we chose to leverage an intermediate because it allowed us to convert the Sigstore root CA into a PKI root for all use-cases across Sigstore (currently just Fulcio, but it could include a timestamp authority too).

We didn't want to maintain CRLs because clients struggle with checking them. Either you need to always query an CRL online (no more offline verification then), or cache CRLs and pick some cadence to update. Size can also become an issue, but not worried about that here since root CA CRLs rarely grow large.

But we don't need CRLs for the public good instance, because we have TUF! We put the intermediate and the root in TUF, and so if either become untrusted, we simply remove them from the TUF trust root. Clients (Cosign for example) fetch the verification material from TUF, and for OCI, ignore any certs set in the `chain` annotation (if the TUF root does not contain intermediates, then Cosign does read only the intermediates, not the root, from the chain annotation)

**tl;dr - For cases of verification from the public good instance, it is sufficient to include only the leaf/end-entity certificate in the bundle**

But wait! But what about non-public good instance Sigstore, private Sigstore instances? If they aren't using TUF, this is an issue if they want to rely on a more standard-pki-like experience. Maybe they don't want to ship intermediates in their trust root and rely on CRLs (there was an issue in Cosign about using CRLs). For these use-cases, it would be preferable to include the full chain in the bundle (the root is never necessary, but like I said above, I think it's useful for auditing purposes).

So **second tl;dr - I would consider including the full chain, but if you don't want to support this now, I would at least document this as a limitation/isssue**

woodruffw: Thanks for the extremely detailed explanation @haydentherapper! 

Including the intermediate and root certs would require some modifications to the current `SigningResult` model, but not impractical ones. Given that, I'll add a comment here and do that in a follow-up (since I plan on more generally refactoring `SigningResult` per https://github.com/sigstore/sigstore-python/pull/465#discussion_r1082935187 anyways).

woodruffw: 56f0636..8651f95

sigstore/_cli.py
woodruffw: Hmm, here's another option: instead of allowing users to opt into Sigstore bundles with `--bundle`, maybe we enable it by default and require them to disable it with `--no-bundle`? This would come at the cost of slightly confusing files created by default (we'd essentially dupe the verification materials on disk), but we'd also end up getting the format upgrade for "free" since users wouldn't have to remember to update their CLI. 

`--bundle` itself could then be `--bundle OUTPUT`, similar to the existing options for that.

Thoughts @di?

woodruffw: (This would also maybe make deprecating the existing outputs a little easier: we could begin warning that `{input}.{crt,sig,rekor}` are deprecated, and then hide them behind a `--legacy-outputs` or similar flag in an upcoming major release.

sigstore/sign.py
kommendorkapten: Yes, would be messy. I would love for Protobufs to support that natively, so many use-cases I have seen where it would make our lives easier.

sigstore/_cli.py
woodruffw: From discussion with @tetsuo-cpp: I'm convinced that `--no-bundle` is a better route here, and that we should emit the bundle by default. From there, `--bundle` can have the same behavior as the other single-output options do (i.e. write to the given file, rather than `{input}.bundle`).

tetsuo-cpp: I have a commit that does this on the way.

sigstore/_cli.py
```diff
@@ -669,11 +739,16 @@ def _sign(args: argparse.Namespace) -> None:
                 print(result.cert_pem, file=io)
             print(f"Certificate written to {outputs['cert']}")
 
+        if outputs["rekor_bundle"] is not None:
```
tetsuo-cpp, line 742: I changed these back from:
```python
if "rekor_bundle" in outputs:
```
This is because when we don't write to an output, the entry still gets added to the map, the file is just set to `None`.

tetsuo-cpp (dismissed):
LGTM!

@woodruffw I pushed a few commits to make the CLI work as we described, so please take a look at that when you get a moment. Other than that, this looks great. 🎉

sigstore/_cli.py
di: I agree with the plan to use `--no-bundle` and have bundles generated by default.

sigstore/_cli.py
```diff
@@ -655,7 +725,7 @@ def _sign(args: argparse.Namespace) -> None:
         print(f"Transparency log entry created at index: {result.log_entry.log_index}")
 
         sig_output: TextIO
-        if outputs["sig"]:
+        if outputs["sig"] in outputs:
```
di, line 728: This is a bug: the value `outputs["sig"]` will never be `in outputs` (the set of keys in `outputs`), and thus it will never write the `.sig` file.

I think this was meant to be:

```suggestion
        if outputs["sig"] is not None:
```

(I'm also curious why this wasn't caught by a test.)

di (changes requested):
Overall LGTM, aside from the bug with writing `.sig` files.

Just a note: we probably don't want to make a release with this until verification is supported as well.

sigstore/_cli.py
woodruffw: Yeah, this is supposed to be `"sig" in outputs` since we don't actually know the `"sig"` key is even present; this looks like a bad bulk edit on my part.

(I think we didn't catch this because the CLI code still isn't well-covered. But this is a good incentive to begin to raise it 🙂)

woodruffw: d045d21

woodruffw:
> Just a note: we probably don't want to make a release with this until verification is supported as well.

Sounds good!

sigstore/_cli.py
woodruffw: Makes sense! I've re-fixed https://github.com/sigstore/sigstore-python/pull/465#discussion_r1086729138 based on that.

sigstore/_cli.py
woodruffw: (I looked at our tests again, and this would have been caught by the conformance suite once it merged on main -- this is making me think that we should probably also run the conformance suite on PRs.)

di (approved):

PR #471, opened 2023-01-25
pyproject: semver match for tuf dep

woodruffw:
Changes our `tuf` constraint from `>= 2.0.0` to `~= 2.0.0`, since `tuf` is semantically versioned and we don't expect to be compatible with arbitrary major version changes.

cc @jku 


tetsuo-cpp (approved):
LGTM.

Issue #249, opened 2022-10-11
Chore: use newtypes to make the API more misuse-resistant


woodruffw:
This issue is just a reminder for myself.

We have a few internal APIs that take sequences of `str` parameters, where each is a base64 (or not-base64) encoded signature, PEM-encoded certificate, etc. It's easy to potentially forget a layer of base64 encoding or decoding with these.

Some `typing.NewType` usage would help here:

* `b64str` for a string that's base64-encoded
* `pemcert` for a string that contains a PEM-encoded certificate (**not** wrapped in another layer of base64)

woodruffw:
Some students at Purdue have volunteered to take a stab at this, so I've removed the request for help.

woodruffw:
No problem! Feel free to ping me either here or on the Sigstore Slack if you have any other questions.

emilejbm:
Emile here! Thanks for all the help William.

emboman13:
We're currently working on https://github.com/emilejbm/sigstore-python. We've changed some files to use NewTypes specified in newtypes.py. Give us a shout if anything looks incorrect or if we've misinterpreted what you wanted us to do. Specifically, changes to sign.py and verify/models.py

woodruffw:
> Give us a shout if anything looks incorrect or if we've misinterpreted what you wanted us to do. Specifically, changes to sign.py and verify/models.py

Would you mind opening a draft PR for your changes? That would make it much easier to do an initial round of review 🙂

woodruffw:
#474.

PR #474, opened 2023-01-26
NewTypes for clearer encoding types 

emilejbm:
Issue #249

#### Summary
 Using newtype for certain parameters that were previously only str will make the API more misuse-resistant

#### Release Note
 NONE

#### Documentation
 NewTypes can be documented so that users can verify that return types are of expected types


find_keywords.sh
```diff
@@ -0,0 +1,3 @@
+#/bin/bash
+
+grep -nr "b64\|base64\|cert_pem\|pem" sigstore/* test/* > key_lines.txt
```
woodruffw, line 3: Nit: fine for now, but this should be dropped from the changeset once it's ready for a final review 🙂

sigstore/_utils.py
```diff
@@ -32,6 +32,8 @@
 else:
     from importlib import resources
 
+import newtype
```
woodruffw, line 35: Nonexistent module? I see you added `newtypes.py` so this should probably be `import newtypes` (but I'll leave a comment w/ thoughts about avoiding a new module).

sigstore/newtypes.py
```diff
@@ -0,0 +1,6 @@
+from typing import NewType
+
+Hexstr = NewType('hexstr', str)
+B64str = NewType('b64str', str)
+Pemcert = NewType("pemcert", str)
+KeyID = NewType('keyID', bytes)
```
woodruffw, line 6: The `NewType` API expects the LHS of the expression and the first argument to match, so these should be:

```suggestion
hexstr = NewType('hexstr', str)
b64str = NewType('b64str', str)
pemcert = NewType("pemcert", str)
keyid = NewType('keyid', bytes)
```

(I made these all lowercase because they're all `NewType`s of primitive types, but it might make sense to have them be `PEMCert` etc. instead. But we can think more about that in a later review.)

sigstore/newtypes.py
```diff
@@ -0,0 +1,6 @@
+from typing import NewType
```
woodruffw, line 1: My preference would be to avoid a new module here (especially a public one) -- IMO these newtypes should go in the existing `_utils` module, since that's where a lot of them get constructed anyways.

woodruffw:
Thanks for opening this! Looks like a good first start; I left a couple of design nits and suggested changes.

FYI: there are a couple of things that our QA tools would have caught -- if you run `make lint` locally, you should be able to follow the output there to get things fixed.

(You can also run `make reformat` to automatically run code formatting.)

sigstore/newtypes.py
omartounsi7: Should we then `import _utils` instead of `import newtypes` in every file that needs changes? I am wondering if there is a way of putting all the needed newtypes in a function and doing `from _utils import function` instead.

woodruffw: > Should we then import _utils instead of import newtypes in every file that needs changes? 

Yes, exactly (although you should use the fully qualified `sigstore._utils` path for consistency with the rest of the code).

> I am wondering if there is a way of putting all the needed newtypes in a function and doing from _utils import function instead.

This is probably overkill -- having them at the top-level of `_utils` is idiomatic, and we can selectively import them as needed.

sigstore/sign.py
```diff
@@ -73,6 +73,7 @@
 from sigstore._internal.tuf import TrustUpdater
 from sigstore._utils import sha256_streaming
 from sigstore.transparency import LogEntry
+import newtypes
```
woodruffw, line 76: Nit: remove entirely and replace with imports from `sigstore._utils`.

sigstore/verify/models.py
```diff
@@ -31,6 +31,8 @@
 from sigstore._utils import base64_encode_pem_cert, sha256_streaming
 from sigstore.transparency import LogEntry
 
+from sigstore import newtypes
```
woodruffw, line 34: Nit: same as https://github.com/sigstore/sigstore-python/pull/474/files#r1092359520

sigstore/sign.py
```diff
@@ -73,7 +73,10 @@
 from sigstore._internal.tuf import TrustUpdater
 from sigstore._utils import sha256_streaming
 from sigstore.transparency import LogEntry
-import newtypes 
+from sigstore._utils import hexstr
+from sigstore._utils import b64str
+from sigstore._utils import pemcert
+from sigstore._utils import keyid
```
woodruffw, line 79: These should be imported in a single line; `make reformat` should do this for you automatically.

woodruffw:
Looks like you might have done a faulty merge/rebase somewhere: your diff contains some unrelated changes that are already on `main`.

emboman13:
> Looks like you might have done a faulty merge/rebase somewhere: your diff contains some unrelated changes that are already on `main`.

Yea; ran a rebase to try and fix an issue w/ missing signatures on commits; managed to make a new one. What would you recommend doing?

emboman13:
I think running 'git reset --hard ORIG_HEAD' should undo the rebase?

woodruffw:
> I think running 'git reset --hard ORIG_HEAD' should undo the rebase?

Yes, something like that. You can `git reflog` to go though your local reference log and then `git reset --hard WHATEVER_REF` came right before your rebase.

(FWIW, don't worry about the signoffs on commits -- I can fix that up before merge.)

omartounsi7:
@woodruffw I am not sure what this does:

`base64.b64encode(something).decode()`

I am assuming it encodes something in base 64 then decodes it, so I did not put any b64str newtypes around it. Is this correct?

jku:
> I am not sure what this does:
> `base64.b64encode(something).decode()`

`b64encode()` encodes data in base64. But it operates in bytes while we often want a UTF-8 string: This is what `decode()` is for, it decodes bytes into a string.

woodruffw:
Yep, what @jku said. So in that instance, that should indeed be a `b64str`, I believe.

omartounsi7:
@woodruffw Hello William, we have pushed all of our changes. But there is a conflict with _internal/rekor/client.py which we don't know how to resolve. There is a class RekorBundle that exists on our branch but not on the main branch. Other than that, we think we are ready for a code review.

woodruffw:
> There is a class RekorBundle that exists on our branch but not on the main branch.

You can remove that class and anything related to it from your branch -- that's part of an old internal API that we've removed.

sigstore/_internal/ctfe.py
```diff
@@ -25,7 +25,7 @@
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import ec, rsa
 
-from sigstore._utils import key_id, load_pem_public_key
+from sigstore._utils import key_id, keyid, load_pem_public_key
```
woodruffw, line 28: Hmm, having both `key_id` (a function) and `keyid` (a type) is definitely a little confusing. 

Let's go with CamelCase for these types after all:

* `keyid -> KeyID`
* `hexstr -> HexStr`
* `b64str -> B64Str`
* `pemcert -> PEMCert`
* `dercert -> DERCert`

sigstore/_utils.py
```diff
@@ -32,8 +32,15 @@
 else:
     from importlib import resources
 
+
 PublicKey = Union[rsa.RSAPublicKey, ec.EllipticCurvePublicKey]
 
+hexstr = NewType("hexstr", str)
```
woodruffw, line 38: These won't be public APIs, but let's add docstrings for them anyways, e.g.:

```suggestion
hexstr = NewType("hexstr", str)
"""
A newtype for `str` objects that contain hexadecimal strings (e.g. `ffabcd00ff`).
"""
```

omartounsi7:
We've implemented the requested changes. Are the docstrings good?

sigstore/_utils.py
```diff
@@ -32,8 +32,30 @@
...
+"""
+A newtype for `str` objects that contain base64 encoded strings.
+"""
+PEMCert = NewType("PEMCert", str)
+"""
+A newtype for `str` objects that contain PEM-encoded certificates.
+"""
+DERCert = NewType("DERCert", bytes)
+"""
+A newtype for `str` objects that contain DER-encoded certificates.
```
woodruffw, line 52:
```suggestion
A newtype for `bytes` objects that contain DER-encoded certificates.
```

sigstore/_utils.py
```diff
@@ -32,8 +32,30 @@
...
+"""
+KeyID = NewType("KeyID", bytes)
+"""
+A newtype for `str` objects that contain a key id.
```
woodruffw, line 56:
```suggestion
A newtype for `bytes` objects that contain a key id.
```

woodruffw:
> We've implemented the requested changes. Are the docstrings good?

Two small errors that I've made suggestions for, otherwise looks good! Once you get those fixed feel free to remove the draft status and we'll do a final review.

woodruffw (approved):

woodruffw:
Thanks all! Great work.

PR #478, opened 2023-01-30
Support verifying Sigstore bundles

tetsuo-cpp:


sigstore/transparency.py
```diff
@@ -76,6 +76,13 @@ class LogEntry:
     The base64-encoded Signed Entry Timestamp (SET) for this log entry.
     """
 
+    # NOTE: After Rekor bundles (provided by `--rekor-bundle`) are removed, this will no longer be
+    # necessary.
+    _from_rekor_bundle: bool
```
tetsuo-cpp, line 81: This isn't great, but we'll be able to remove it soon when we get rid of `--rekor-bundle` so I'm not too worried about it.

The only problem I see is that even though this field is marked as internal, you still need to provide a value for it if you construct one yourself. But I don't think users of the API have a good reason to be doing that anyway.

di: This also outputs a `RuntimeWarning` from Pydantic at runtime:

> `RuntimeWarning: fields may not start with an underscore, ignoring "_from_rekor_bundle"`

woodruffw: Yeah, I think we can do a little better than this API (both the private fiel;d and the manual initialization). I'll look into it.

woodruffw: Hmm, maybe we can just remove `--rekor-bundle` outright? We never "actually" stabilized it (we've always emitted a warning said that it's temporary and will be removed in an upcoming release), and we never integrated it into other tooling (like the GH Action) or other expectations (e.g. CPython doesn't list it).

Thoughts @di?

di: Fine with me!

woodruffw: Cool! I'll just yank all of that out, then, and save us from even having to think about this 🙂

sigstore/verify/models.py
```diff
@@ -123,24 +135,31 @@ class VerificationMaterials:
...
     """
 
-    _offline_rekor_entry: LogEntry | None
+    offline: bool
+    """
+    Whether to do offline Rekor entry verification.
+
+    This is a slightly weaker verification verification mode, as it demonstrates
+    that an entry has been signed by the log but not necessarily included in it.
     """
```
woodruffw, line 153: Flagging: I'm not super happy with this setting being here, since it's conceptually a verification option rather than a material. 

The other possibility I can think of is a separate `VerificationOptions` model, but that might be overkill given that this is currently the only option we support. But maybe there will be more in the future?

woodruffw:
@tetsuo-cpp I'll kick this back over to you, but to summarize the changes I've made:

* I removed Rekor bundle support entirely, both in the internal APIs and in the CLI
* I added `VerificationMaterials.offline` and `VerificationMaterials.from_bundle(...)`, the latter being the expected API for converting a Sigstore bundle into its equivalent materials.
* I fixed up the unit tests for my changes and added some new assets/updated the fixtures to reflect them.

I think this is in a pretty good state overall, but IMO we should bring the test coverage up a bit -- it'd be good to get some coverage for the new `InvalidMaterials` exception and its raise-sites.

tetsuo-cpp:
> I think this is in a pretty good state overall, but IMO we should bring the test coverage up a bit -- it'd be good to get some coverage for the new `InvalidMaterials` exception and its raise-sites.

Done in [dc1cb3b](https://github.com/sigstore/sigstore-python/pull/478/commits/dc1cb3bb77e5c810689577ead0303ad28ebb124e).

sigstore/verify/models.py
tetsuo-cpp: Personally, I think it's ok.

If we go for a separate `VerificationOptions` model, we're never going to collapse that back into `VerificationMaterials` even if it remains as one option for a very long time because there's always going to be the possibility of more.

sigstore/_cli.py
```diff
@@ -214,10 +201,10 @@ def _add_shared_verification_options(group: argparse._ArgumentGroup) -> None:
         required=True,
     )
     group.add_argument(
-        "--require-rekor-offline",
+        "--offline",
         action="store_true",
-        default=_boolify_env("SIGSTORE_REQUIRE_REKOR_OFFLINE"),
-        help="Require offline Rekor verification with a bundle; implied by --rekor-bundle",
+        default=_boolify_env("SIGSTORE_OFFLINE"),
+        help="Perform offline verification; requires a Sigstore bundle",
```
woodruffw, line 207: I'm now realizing that this is slightly misleading, since `--offline` implies "fully" offline, whereas we still do some network requests for TUF. I think we should keep this flag, though, and work on removing those requests.

sigstore/verify/models.py
woodruffw: Yeah, that's a good point. I think what I'll do it make it "private" like `_rekor_entry` so that it isn't part of our public API commitment. It'll still end up being stabilized as part of the methods on `VerificationMaterials`, but that'll at least avoid the need to change this model's public fields if we ever do decide to refactor it.

woodruffw (approved):
LGTM! :shipit:

PR #484, opened 2023-01-31
Prep 1.1.0

woodruffw:
~~WIP; need to fix the conformance tests.~~

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -29,6 +29,6 @@ jobs:
       - name: install sigstore-python
         run: python -m pip install .
 
-      - uses: trailofbits/sigstore-conformance@0748d63c53810e36cc3f4bbe4114301080f0d844 # v0.0.3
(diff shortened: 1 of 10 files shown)


woodruffw:
Ugh, forgot -- this isn't actually running my conformance suite changes, since they need to be propagated to `main` first 🙃

.github/workflows/conformance.yml
```diff
@@ -29,6 +29,6 @@ jobs:
       - name: install sigstore-python
         run: python -m pip install .
 
-      - uses: trailofbits/sigstore-conformance@0748d63c53810e36cc3f4bbe4114301080f0d844 # v0.0.3
+      - uses: sigstore/sigstore-conformance@0748d63c53810e36cc3f4bbe4114301080f0d844 # v0.0.3
```
woodruffw, line 32: cc @di: I think this will fix the conformance suite, but I can't actually verify that without merging because of how `pull_request_target` workflows are updated.

di (approved):

PR #507, opened 2023-02-22
Use DEFAULT_AUDIENCE instead of hard-coding

di:
Caught this when breaking this out into a separate library.


woodruffw (approved):

woodruffw:
Nice catch 🙂

PR #527, opened 2023-03-06
Root hash signature verification

### LINKED ISSUE #248: Include Root Hash verification when verifying a Rekor Inclusion Proof
**Description**

Mentioned here:
```
 Not related to this PR, but something we need to add - We now return checkpoints (https://github.com/sigstore/rekor/blob/main/openapi.yaml#L608), signed root hashes. We need to add verification of these when verifying the inclusion proof, otherwise we're just verifying a list of hashes without any connection to the log.

```

When verifying inclusion proof, there needs to be a pin on the root hash: this root hash can ideally come from verifying consistency with a stored client state, or verifying the signature on the root hash (e.g. a Signed Root Hash with the rekor public key). Since sigstore clients don't store a rekor state, their best option is the latter.

Otherwise inclusion proofs don't mitigate attacks where you are served a valid inclusion proof on some random log, not Rekor.

cc @haydentherapper

jleightcap:
Closes #248.

### DIFF
--- a/sigstore/_internal/rekor/checkpoint.py
+++ b/sigstore/_internal/rekor/checkpoint.py
@@ -19,15 +19,11 @@ Rekor Checkpoint machinery.
 from __future__ import annotations
 
 import base64
-import hashlib
 import re
 from dataclasses import dataclass
+from typing import List
 
-from cryptography.exceptions import InvalidSignature
 from pydantic import BaseModel, Field, StrictStr
-
-from sigstore._internal.rekor import RekorClient
-from sigstore._utils import KeyID
 
 
 @dataclass(frozen=True)
@@ -59,7 +55,7 @@ class LogCheckpoint(BaseModel):
     origin: StrictStr
     log_size: int
     log_hash: StrictStr
-    other_content: list[str]
+    other_content: List[str]
 
     @classmethod
     def from_text(cls, text: str) -> LogCheckpoint:
@@ -165,26 +161,6 @@ class SignedNote:
             signatures.append(signature)
 
         return cls(note=header, signatures=signatures)
-
-    def verify(self, client: RekorClient) -> None:
-        """
-        Verify the SignedNote with using the given RekorClient by verifying each contained signature.
-        """
-
-        note = hashlib.sha256(self.note.encode("utf-8")).digest()
-
-        # Grab the singular Rekor root public key as the signing key.
-        key_id = list(client._rekor_keyring._keyring.keys())[0]
-
-        for signature in self.signatures:
-            try:
-                client._rekor_keyring.verify(
-                    key_id=KeyID(key_id),
-                    signature=base64.b64decode(signature.signature),
-                    data=note,
-                )
-            except InvalidSignature as inval_sig:
-                raise InvalidSignedNote("invalid signature") from inval_sig
 
 
 @dataclass(frozen=True)
--- a/test/unit/internal/rekor/test_client.py
+++ b/test/unit/internal/rekor/test_client.py
@@ -22,7 +22,7 @@ from sigstore.transparency import LogInclusionProof
 class TestRekorInclusionProof:
     def test_valid(self):
         proof = LogInclusionProof(
-            log_index=1, root_hash="abcd", tree_size=2, hashes=[], checkpoint=""
+            log_index=1, root_hash="abcd", tree_size=2, hashes=[], checkpoint=None
         )
         assert proof is not None
 
@@ -30,17 +30,23 @@ class TestRekorInclusionProof:
         with pytest.raises(
             ValidationError, match="Inclusion proof has invalid log index"
         ):
-            LogInclusionProof(log_index=-1, root_hash="abcd", tree_size=2, hashes=[])
+            LogInclusionProof(
+                log_index=-1, root_hash="abcd", tree_size=2, hashes=[], checkpoint=None
+            )
 
     def test_negative_tree_size(self):
         with pytest.raises(
             ValidationError, match="Inclusion proof has invalid tree size"
         ):
-            LogInclusionProof(log_index=1, root_hash="abcd", tree_size=-1, hashes=[])
+            LogInclusionProof(
+                log_index=1, root_hash="abcd", tree_size=-1, hashes=[], checkpoint=None
+            )
 
     def test_log_index_outside_tree_size(self):
         with pytest.raises(
             ValidationError,
             match="Inclusion proof has log index greater than or equal to tree size",
         ):
-            LogInclusionProof(log_index=2, root_hash="abcd", tree_size=1, hashes=[])
+            LogInclusionProof(
+                log_index=2, root_hash="abcd", tree_size=1, hashes=[], checkpoint=None
+            )


woodruffw:
> Why have the `sigstore_protobuf_specs.dev.sigstore.*.v1` types in addition to some manually-defined... very close types?

The types under `sigstore_protobuf_specs` are only meant for client interoperation -- they're pure models, without any associated functionality. We marshal from them early on in `sigstore-python`'s lifecycle and into less flexible models that reflect the subset of Sigstore functionality we actually intend to support 🙂

woodruffw:
> What's been throwing me off w.r.t. expanding the [`VerificationMaterial` checkpoint field](https://github.com/sigstore/rekor/blob/main/openapi.yaml#L610) is that `models.py` declares `VerificationMaterials` (with an _s_) that we seem to actually operate over in `verify`.

`VerificationMaterials` and `VerificationMaterial` are confusingly named, and are conceptually very closely related: the former is our restricted model with verification functionality, and the latter is the interoperation model.

In this context, I think what we need to do is add a new `checkpoint` attribute to our `LogEntry` model, pulling it from the bundle's inclusion proof. That `checkpoint` then needs to be verified per the steps in #248. The relevant site for that is probably `VerificationMaterials.from_bundle`.

woodruffw:
> Ah, OK -- so that extends to a lot of the extremely close types, same with `LogInclusionProof` which was more obviously useful in leveraging pydantic. My next step then is to make sure I'm threading through the protobuf field correctly...

Yep, exactly; there's a roughly 1:1 correspondence between many `sigstore` and `sigstore_protobuf_specs` types, with the former being what we want to marshal into (because we can do stronger correctness checks versus the protobuf definitions).

woodruffw:
It took a little bit of digging, but it looks like a checkpoint is a `SignedNote`, which has an unmarshalling implementation defined here:

https://github.com/sigstore/rekor/blob/4b1fa6661cc6dfbc844b4c6ed9b1f44e7c5ae1c0/pkg/util/signed_note.go#L129-L187

> And in terms of usage, should this string be parsed into a pydantic-verified class or something? 

Yes! This should be one or more Pydantic data models.

woodruffw:
Specifically, here's the `SignedNote` representation:

```go
type SignedNote struct {
	// Textual representation of a note to sign.
	Note string
	// Signatures are one or more signature lines covering the payload
	Signatures []note.Signature
}
```

and a `SignedCheckpoint` is:

```go
type SignedCheckpoint struct {
	Checkpoint
	SignedNote
}
```

where unmarshalling is defined as:

```go
func (r *SignedCheckpoint) UnmarshalText(data []byte) error {
	s := SignedNote{}
	if err := s.UnmarshalText([]byte(data)); err != nil {
		return fmt.Errorf("unmarshalling signed note: %w", err)
	}
	c := Checkpoint{}
	if err := c.UnmarshalCheckpoint([]byte(s.Note)); err != nil {
		return fmt.Errorf("unmarshalling checkpoint: %w", err)
	}
	*r = SignedCheckpoint{Checkpoint: c, SignedNote: s}
	return nil
}
```

and the underlying `Checkpoint` type:

```go
type Checkpoint struct {
	// Origin is the unique identifier/version string
	Origin string
	// Size is the number of entries in the log at this checkpoint.
	Size uint64
	// Hash is the hash which commits to the contents of the entire log.
	Hash []byte
	// OtherContent is any additional data to be included in the signed payload; each element is assumed to be one line
	OtherContent []string
}

// UnmarshalText parses the common formatted checkpoint data and stores the result
// in the Checkpoint.
//
// The supplied data is expected to begin with the following 3 lines of text,
// each followed by a newline:
// <ecosystem/version string>
// <decimal representation of log size>
// <base64 representation of root hash>
// <optional non-empty line of other content>...
// <optional non-empty line of other content>...
//
// This will discard any content found after the checkpoint (including signatures)
func (c *Checkpoint) UnmarshalCheckpoint(data []byte) error {
	l := bytes.Split(data, []byte("\n"))
	if len(l) < 4 {
		return errors.New("invalid checkpoint - too few newlines")
	}
	origin := string(l[0])
	if len(origin) == 0 {
		return errors.New("invalid checkpoint - empty ecosystem")
	}
	size, err := strconv.ParseUint(string(l[1]), 10, 64)
	if err != nil {
		return fmt.Errorf("invalid checkpoint - size invalid: %w", err)
	}
	h, err := base64.StdEncoding.DecodeString(string(l[2]))
	if err != nil {
		return fmt.Errorf("invalid checkpoint - invalid hash: %w", err)
	}
	*c = Checkpoint{
		Origin: origin,
		Size:   size,
		Hash:   h,
	}
	if len(l) >= 3 {
		for _, line := range l[3:] {
			if len(line) == 0 {
				break
			}
			c.OtherContent = append(c.OtherContent, string(line))
		}
	}
	return nil
}
```

test/unit/verify/test_models.py
```diff
@@ -25,6 +25,7 @@
 
 
 class TestVerificationMaterials:
+    @pytest.mark.xfail
```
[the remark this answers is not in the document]
woodruffw: > At the moment these tests load the commited bundle.txt.sigstore which does not contain the checkpoint field. This seems to point to backwards-compatibility concerns...

That's probably because it's not included in our own bundle generation code 🙂 -- it's an optional field in the protobuf definition, which means that it was probably an oversight on my part. 

If the current bundles are missing checkpoints, we should fix the bundle generation code and check in some new (additional) bundles that contain checkpoints on their tlog entries. The current ones should stay and be renamed to indicate that they lack checkpoints, since they serve as a good behavioral test for handling that case.

sigstore/sign.py
```diff
@@ -238,6 +238,9 @@ def _to_bundle(self) -> Bundle:
                 hashes=[
                     bytes.fromhex(h) for h in self.log_entry.inclusion_proof.hashes
                 ],
+                # FIXME(jl): checkpoint should be a serialzied field here.
```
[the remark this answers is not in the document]
woodruffw: Hmm, that's strange: it's defined here:

https://github.com/sigstore/protobuf-specs/blob/c8a23a43dc1c685d336d112864ba237a15061c53/protos/sigstore_rekor.proto#L64

The release we use should be built from a version that contains that change.

sigstore/_internal/merkle.py
```diff
@@ -133,3 +134,33 @@ def verify_merkle_inclusion(entry: LogEntry) -> None:
...
+    Verify the inclusion proof's checkpoint.
+    """
+
+    inclusion_proof = entry.inclusion_proof
+    if inclusion_proof is None:
+        raise InvalidInclusionProofError("Rekor entry has no inclusion proof")
+    if inclusion_proof.checkpoint is None:
+        return
+
+    # verififcaiton occurs in two stages:
```
woodruffw, line 150:
```suggestion
    # verification occurs in two stages:
```

sigstore/_internal/merkle.py
```diff
@@ -133,3 +134,33 @@ def verify_merkle_inclusion(entry: LogEntry) -> None:
...
+def verify_checkpoint(client: RekorClient, entry: LogEntry) -> None:
+    """
+    Verify the inclusion proof's checkpoint.
+    """
+
+    inclusion_proof = entry.inclusion_proof
+    if inclusion_proof is None:
+        raise InvalidInclusionProofError("Rekor entry has no inclusion proof")
+    if inclusion_proof.checkpoint is None:
+        return
```
woodruffw, line 148: Shouldn't this be an error case? We probably shouldn't silently accept inclusion proofs that don't contain a checkpoint.

sigstore/_internal/merkle.py
```diff
@@ -133,3 +134,33 @@ def verify_merkle_inclusion(entry: LogEntry) -> None:
...
+
+    # verififcaiton occurs in two stages:
+    # 1) verify the signature on the checkpoint
+    # 2) verify the root hash in the checkpoint matches the root hash from the inclusion proof.
+
+    signed_checkpoint = SignedCheckpoint.from_text(inclusion_proof.checkpoint)
+
+    # FIXME(jl): produces an invalid signature exception -- am I using the keyring correctly?
+    # signed_checkpoint.signed_note.verify(client)
```
woodruffw, line 157: I might be missing it, but where is this `verify` method defined? I don't see it declared on `SignedNote`.

woodruffw:
Superseded by #634.

PR #525, opened 2023-03-06
`TrustUpdater` error handling

tnytown:
As discussed on #357, `TrustUpdater` does not have a coherent error handling scheme. The following exceptions are exposed to users:

- `OSError`, when file read/write issues occur
- TUF's `DownloadError`, when TUF fails to download metadata / target files
- TUF's `RepositoryError`, when TUF metadata is invalid
- `Exception`, raised when we don't find the entries we expect in TUF metadata

This PR wraps TUF's error types with our own `TUFError`, which will help us provide useful diagnostics to users when a TUF-internal issue occurs. Additionally, `MetadataError` is raised when the `get` methods on `TrustUpdater` don't find what they need.

Also in this PR is the structure for a new global error handling system. The `Error` type implements utilities, namely `print_and_exit`, for providing diagnostics to users. `TUFError` extends from this type.

Further work might include adapting existing error types, e.g. `CertificateVerificationFailure`, to this model. Doing this might make it easier for API consumers to handle Sigstore exceptions and allow us to export all of our error types from one place.

Resolves #357.

PR #531, opened 2023-03-07
Move all error diagnostics to _errors

tnytown:
#### Summary

This is the work mentioned in #525. I grepped for `dedent` and moved all user-facing error messages to subclasses of `Error` w/ the `diagnostics` method.

Depends on #525, that should be merged first.

#### Release Note

(Same as #525)

* A new error handling system was added. Diagnostic messages are now printed for certain errors.


sigstore/_errors.py
```diff
@@ -0,0 +1,150 @@
...
+    """Raised when TUF metadata does not conform to the expected structure."""
+
+    def diagnostics(self) -> str:
+        return f"""{str(self)}."""
+
+
+class VerificationError(Error):
+    """Raised when the verifier returns a `VerificationFailure` result."""
+
+    # HACK(tnytown): Importing this at the top of the module causes circular import issues.
```
tnytown, line 95: This is suspicious, I would appreciate guidance on how we can make it better 😬

tnytown:
Sigh, ok. I will need to definitively figure out the circular import issue.

sigstore/_errors.py
tetsuo-cpp: I just pushed [fc70655](https://github.com/sigstore/sigstore-python/pull/531/commits/fc706557c74e866faf4dc514b5a1b079a7e09300). Feel free to revert it if you'd prefer it some other way.

I fixed the circular import by moving `VerificationError` to `_cli.py`. I think this makes sense because this is a specialised error type that's only used in that module (kinda like how we define `IdentityError` in `oidc.py` since we only use it there).

tnytown: The move looks good, thanks! I wonder if it will keep making sense to declare errors in a module separate from their throw sites. We might run into similar issues with new errors if we ever want to dispatch on the type of a member variable as I did here :(

tnytown:
newly failing because of a recently introduced lint [in bandit](https://bandit.readthedocs.io/en/1.7.5/plugins/b113_request_without_timeout.html).

tnytown:
Blocked on #541.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ All versions prior to 0.9.0 are untracked.
 
 * Replaced ambient credential detection logic with the `id` package
   ([#535](https://github.com/sigstore/sigstore-python/pull/535))
+* Revamped error diagnostics reporting. All errors with diagnostics now implement
+  `sigstore._errors.Error`.
```
woodruffw, line 23: The `CHANGELOG` reflects public changes, so maybe it makes sense to make `sigstore._errors` a public module? That or re-export `sigstore._error.Error` as `sigstore.Error`.

tnytown: `python-tuf` does this and it makes sense IMO, but we'd have to shuffle some things around to avoid import cycles.

woodruffw: > Would we like to move all errors we expose to API consumers to `sigstore._errors`?

If we do this, we have to make sure that their existing modules re-export them, as part of not breaking out API guarantees. 

Given that, let's limit this PR to just making `sigstore._errors` public as-is, and then open a follow-up issue for consolidating our error types.

woodruffw (approved):

PR #535, opened 2023-03-09
treewide: Replace ambient credential detection with `id`

### LINKED ISSUE #528: Replace our ambient credential detection with `id`
`id` is now on PyPI (https://pypi.org/project/id/), and contains almost exactly the same APIs we have internally for ambient credential detection. We should switch to it.

tetsuo-cpp:
Closes #528


woodruffw (approved):
LGTM! This always makes me happy:

Screenshot 2023-03-09 at 9 57 49 AM

PR #541, opened 2023-03-11
oidc: configure timeouts for `requests`

tnytown:
#### Summary

Bandit recently released v1.7.5, which includes [a new lint for `requests` usage without a timeout](https://bandit.readthedocs.io/en/1.7.5/plugins/b113_request_without_timeout.html). This changeset fixes our usage of `requests` in `oidc`.

### DIFF
--- a/sigstore/_errors.py
+++ b/sigstore/_errors.py
@@ -52,10 +52,24 @@ class NetworkError(Error):
 
     def diagnostics(self) -> str:
         """Returns diagnostics for the error."""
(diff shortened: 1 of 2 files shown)


woodruffw (commented):
Let's also handle the `Timeout` exception, now that we're explicitly handling the timeout case. 

(We should have been doing this before, but now's a good opportunity to address it.)

woodruffw (approved):
LGTM, thanks!

PR #542, opened 2023-03-12
tuf: use bundled trusted root if available

### LINKED ISSUE #488: TUF: Support "bundled" trust root
This is a tracking issue for our side of https://github.com/sigstore/root-signing/issues/539 and https://github.com/sigstore/root-signing/pull/584: once the TUF repo contains `trusted_root.json` (or whatever it ends up being called), we should use that instead of retrieving each target in the trustroot individually.

tnytown:
#### Summary

This changeset modifies the TUF component to fetch keys from the `trusted_root.json` target, if available. `trusted_root.json` is a rollup of Sigstore component public keys. Using this target reduces the number of network roundtrips necessary to fetch keys individually.

Resolves #488.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,8 @@ All versions prior to 0.9.0 are untracked.
   ([#535](https://github.com/sigstore/sigstore-python/pull/535))
 * Revamped error diagnostics reporting. All errors with diagnostics now implement
   `sigstore.errors.Error`.
-* TUF now fetches its trust materials from a single target, if it is available.
+* Trust root materials are now retrieved from a single trust bundle,
+  if it is available via TUF
   ([#542](https://github.com/sigstore/sigstore-python/pull/542))
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ All versions prior to 0.9.0 are untracked.
   ([#535](https://github.com/sigstore/sigstore-python/pull/535))
 * Revamped error diagnostics reporting. All errors with diagnostics now implement
   `sigstore.errors.Error`.
+* TUF now fetches its trust materials from a single target, if it is available.
+  ([#542](https://github.com/sigstore/sigstore-python/pull/542))
 
 ### Fixed
 
--- a/sigstore/_internal/keyring.py
+++ b/sigstore/_internal/keyring.py
@@ -68,7 +68,7 @@ class Keyring:
     def __init__(self, keys: List[bytes] = []):
         """
         Create a new `Keyring`, with `keys` as the initial set of signing
-        keys.
+        keys. These `keys` can be in either DER or PEM encoded.
         """
         self._keyring = {}
         for key_bytes in keys:
@@ -78,7 +78,7 @@ class Keyring:
                 key = load_pem_public_key(key_bytes)
             except UnexpectedKeyFormatError as e:
                 raise e
-            except InvalidKeyError as e:
+            except InvalidKeyError:
                 key = load_der_public_key(key_bytes)
 
             self._keyring[key_id(key)] = key
--- a/sigstore/_internal/keyring.py
+++ b/sigstore/_internal/keyring.py
@@ -68,7 +68,7 @@ class Keyring:
     def __init__(self, keys: List[bytes] = []):
         """
         Create a new `Keyring`, with `keys` as the initial set of signing
-        keys. These `keys` can be in either DER or PEM encoded.
+        keys. These `keys` can be in either DER format or PEM encoded.
         """
         self._keyring = {}
         for key_bytes in keys:
--- a/sigstore/_internal/keyring.py
+++ b/sigstore/_internal/keyring.py
@@ -70,9 +70,9 @@ class Keyring:
 
             try:
                 key = load_pem_public_key(key_bytes)
+            except UnexpectedKeyFormatError as e:
+                raise e
             except InvalidKeyError as e:
-                if isinstance(e, UnexpectedKeyFormatError):
-                    raise e
                 key = load_der_public_key(key_bytes)
 
             self._keyring[key_id(key)] = key
--- a/sigstore/_internal/keyring.py
+++ b/sigstore/_internal/keyring.py
@@ -24,7 +24,14 @@ from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import ec, rsa
 
-from sigstore._utils import KeyID, key_id, load_pem_public_key
+from sigstore._utils import (
+    InvalidKeyError,
+    KeyID,
+    UnexpectedKeyFormatError,
+    key_id,
+    load_der_public_key,
+    load_pem_public_key,
+)
 
 
 class KeyringError(Exception):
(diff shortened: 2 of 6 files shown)


woodruffw (commented):
@tetsuo-cpp can do a full review, but as a small nit: needs a `CHANGELOG` entry 🙂

tnytown:
I've left this one on draft because I haven't adjusted the tests to exercise the new codepaths. Most of the new logic seems trivial except for the stuff that ensures that a given target in a trusted root is active/expired. I'll work on adding some basic testing for that.

sigstore/_internal/keyring.py
```diff
@@ -59,7 +66,15 @@ def __init__(self, keys: List[bytes] = []):
         """
         self._keyring = {}
         for key_bytes in keys:
-            key = load_pem_public_key(key_bytes)
+            key = None
+
+            try:
+                key = load_pem_public_key(key_bytes)
+            except InvalidKeyError as e:
```
tetsuo-cpp, line 73: Maybe:
```python
except UnexpectedKeyFormatError as e:
    raise e
except InvalidKeyError as e:
    key = load_der_public_key(key_bytes)
```

sigstore/_internal/tuf.py
```diff
@@ -197,7 +254,21 @@ def get_rekor_keys(self) -> list[bytes]:
...
-        keys = self._get("Rekor", ["Active"])
+        keys = []
+
+        self._trusted_root = self._get_trusted_root()
+        if self._trusted_root:
+            for key in self._trusted_root.tlogs:
+                if not _is_timerange_valid(
+                    key.public_key.valid_for, allow_expired=False
+                ):
+                    continue
```
tetsuo-cpp, line 265: If we're ignoring keys because they're expired, is this something that we want to log about? I feel like it could be confusing for a user if they've been supplying a trusted root and the keys stop working suddenly without logging anything.

tnytown: Yeah, that might be weird from a UX perspective. If we do this, should we also log for this in `_get`, the non bundled case?

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ All versions prior to 0.9.0 are untracked.
   ([#535](https://github.com/sigstore/sigstore-python/pull/535))
 * Revamped error diagnostics reporting. All errors with diagnostics now implement
   `sigstore.errors.Error`.
+* TUF now fetches its trust materials from a single target, if it is available.
+  ([#542](https://github.com/sigstore/sigstore-python/pull/542))
```
woodruffw, line 25:
```suggestion
* Trust root materials are now retrieved from a single trust bundle,
  if it is available via TUF
  ([#542](https://github.com/sigstore/sigstore-python/pull/542))
```

sigstore/_internal/tuf.py
tetsuo-cpp: I think it's fine. If this happens will get a `CTFE keys not found in TUF metadata` error and will report the issue and it should be obvious what the problem is.

tetsuo-cpp (approved):
LGTM!

woodruffw:
```
sigstore/_internal/keyring.py:81:39: F841 [*] Local variable `e` is assigned to but never used
```

sigstore/_internal/keyring.py
```diff
@@ -65,7 +72,15 @@ def __init__(self, keys: List[bytes] = []):
...
         for key_bytes in keys:
-            key = load_pem_public_key(key_bytes)
+            key = None
+
+            try:
+                key = load_pem_public_key(key_bytes)
+            except UnexpectedKeyFormatError as e:
+                raise e
+            except InvalidKeyError as e:
+                key = load_der_public_key(key_bytes)
```
woodruffw, line 82: This has introduced some flexibility in the keyring API (keys can now be PEM or DER encoded, rather than just PEM), so we should document it in the `__init__`'s docstring above.

sigstore/_internal/tuf.py
```diff
@@ -207,9 +278,29 @@ def get_fulcio_certs(self) -> list[Certificate]:
...
+        certs = []
+
+        self._trusted_root = self._get_trusted_root()
         # Return expired certificates too: they are expired now but may have
         # been active when the certificate was used to sign.
-        certs = self._get("Fulcio", ["Active", "Expired"])
+        if self._trusted_root:
+            for ca in self._trusted_root.certificate_authorities:
+                if not _is_timerange_valid(ca.valid_for, allow_expired=True):
+                    continue
```
woodruffw, line 289: This is a no-op, right? If so maybe we can just omit it entirely, and perhaps just leave a comment to the effect of "we could check `_is_timerange_valid` here but don't bother to, since expired Fulcio chains are considered valid for verification purposes."

sigstore/_internal/tuf.py
```diff
@@ -66,6 +74,21 @@ def _get_dirs(url: str) -> tuple[Path, Path]:
     return (tuf_data_dir / repo_base), (tuf_cache_dir / repo_base)
 
 
+def _is_timerange_valid(period: TimeRange | None, allow_expired: bool) -> bool:
```
woodruffw, line 77: Extreme nitpick, forces all users of this function to specify the `allow_expired` kwarg explicitly.

```suggestion
def _is_timerange_valid(period: TimeRange | None, *, allow_expired: bool) -> bool:
```

sigstore/_internal/tuf.py
```diff
@@ -144,6 +168,25 @@ def _setup(self) -> Updater:
...
+        if path is None:
+            try:
+                path = self._updater.download_target(root_info)
+            except (
+                TUFExceptions.DownloadError,
+                TUFExceptions.RepositoryError,
+            ) as e:
+                raise TUFError("Failed to download trusted key bundle") from e
+
+        return TrustedRoot().from_json(Path(path).read_bytes())
```
woodruffw, line 188: We can skip it for now, but something to consider for a future refactor: it might be nicer/cleaner to use an `@lru_cache` decorated function here.

woodruffw (commented):
Small nits 🙂

sigstore/_internal/tuf.py
woodruffw: Got it, thanks!

sigstore/_internal/keyring.py
```diff
@@ -61,11 +68,19 @@ class Keyring:
     def __init__(self, keys: List[bytes] = []):
         """
         Create a new `Keyring`, with `keys` as the initial set of signing
-        keys.
+        keys. These `keys` can be in either DER or PEM encoded.
```
woodruffw, line 71:
```suggestion
        keys. These `keys` can be in either DER format or PEM encoded.
```

woodruffw (approved):
LGTM. Two thoughts:

1. Assuming the Sigstore TUF repo maintainers are okay with it, I think we can transition to codebase to assuming a trust root bundle (and failing if not present) after a migration period. We could retain an explicit opt-in flag (like `--tuf-unbundled-trust-root`), if necessary. CC @asraa for thoughts 🙂 

2. We should add some more unit tests for the bundled trust root. @tnytown could you make a follow-up issue for that?

asraa (commented):
This is amazing!! 

Yes - we are planning to have this be the stable trusted root target from now on. We will leave the existing targets in there for likely a year, and then deprecate them, so the current code with defaults for the trusted_root.json looks good to me.

Honestly, I wouldn't add an option to use the fallback old unbundled targets. IMO there's no need to, the behavior and targets are the same in both the bundled and unbundled so users wouldn't see any difference

woodruffw:
Makes sense, thanks @asraa!

@tnytown: In that case, let's plan to remove support for the old "standalone" targets in an upcoming release. This is entirely an implementation detail in a private API, so we can do it with a patch release.

PR #543, opened 2023-03-13
sct, test_sign: adjust exception messages

tnytown:
Should fix the [failing online tests](https://github.com/sigstore/sigstore-python/actions/runs/4406822705/jobs/7719602599#step:6:881).


woodruffw (commented):
Fix LGTM, but this made me realize that we're probably emitting bad diagnostics for a few other `InvalidSCTError` cases -- we probably need to degeneralize the diagnostics, since not all cases are key lookup failures.

(Sorry, should have caught this during the last PR...)

woodruffw (approved):
Merging to get the CI fixed.

PR #548, opened 2023-03-13
Pin requirements on releases with a workflow

### LINKED ISSUE #115: Tighter update loop for hashed requirements file
In #114 @woodruffw raised that this file may become stale on new releases.

To resolve this and provide a tighter feedback loop on a release -> update, we could

- a) add a pin to the `requirements.in` file
- b) introduce a workflow that ensures the compiled file is up to date, and 
- c) also introduce a dependabot-like workflow that runs on every new new release to create a PR to update the pin (something like https://github.com/di/pip-api/blob/master/.github/workflows/pip-feed.yml should work).

tnytown:
#### Summary

This changeset resolves #115. We introduce a new workflow, `pin-requirements.yml`, which:

- Fetches the latest version of `sigstore-python` from PyPI;
- Updates the version in `requirements.in` from this version;
- Updates the pins in `requirements.txt` with `pip-compile`;
- Opens a pull request to `main` with these updates.

Here's a preview of what this pull request would look like: https://github.com/tnytown/sigstore-python/pull/2

Potential issues:
- Do we want to use a specific Python version for `pip-compile`? Would this affect the resultant `requirements.txt`?
- This workflow should run automatically on new PyPI releases, but I didn't test this end-to-end as I would have had to spin up some infrastructure. The trigger from `release.yml` may or may not work, we'll have to see when a new release rolls around.

### DIFF
--- a/.github/workflows/pin-requirements.yml
+++ b/.github/workflows/pin-requirements.yml
@@ -21,8 +21,6 @@ permissions:
 jobs:
   update-pinned-requirements:
     runs-on: ubuntu-latest
-    env:
-      TMPDIR: ${{ runner.temp }}
 
     permissions:
       pull-requests: write # Pull Request creation.
@@ -47,14 +45,14 @@ jobs:
           # Do some JSON traversal: get the URL of the asset that has a `.whl` file extension.
           wheel_url=$(curl $release_url | jq -r '.assets[] | select(.name | test(".whl$")) | .browser_download_url')
 
-          curl -L "$TMPDIR/sigstore.whl" -L $wheel_url
+          curl -L "$RUNNER_TEMP/sigstore.whl" -L $wheel_url
 
       - name: Update requirements
         run: |
           cd install
 
           # Pin on the downloaded wheel, as PyPI might not have updated yet.
-          echo "$TMPDIR/sigstore.whl" >requirements.in
+          echo "$RUNNER_TEMP/sigstore.whl" >requirements.in
           pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in
 
           # Strip the "v" prefix from the version tag name.
--- a/.github/workflows/pin-requirements.yml
+++ b/.github/workflows/pin-requirements.yml
@@ -44,36 +44,36 @@ jobs:
 
       - name: Compute version from tag
         run: |
-          echo "SIGSTORE_RELEASE_VERSION=$(echo "${{ env.SIGSTORE_RELEASE_TAG }}" | sed 's/^v//')" >>$GITHUB_ENV
+          echo "SIGSTORE_RELEASE_VERSION=$(echo "${SIGSTORE_RELEASE_TAG}" | sed 's/^v//')" >> "${GITHUB_ENV}"
 
       - name: Download wheel from GitHub release
         run: |
-          wheel_name="sigstore-${{ env.SIGSTORE_RELEASE_VERSION }}-py3-none-any.whl"
-          wheel_url="https://github.com/sigstore/sigstore-python/releases/download/${{ env.SIGSTORE_RELEASE_TAG }}/$wheel_name"
-          wheel_path="$RUNNER_TEMP/$wheel_name"
+          wheel_name="sigstore-${SIGSTORE_RELEASE_VERSION}-py3-none-any.whl"
+          wheel_url="https://github.com/sigstore/sigstore-python/releases/download/${env.SIGSTORE_RELEASE_TAG}/${wheel_name}"
+          wheel_path="${RUNNER_TEMP}/${wheel_name}"
 
-          curl -o "$wheel_path" -L $wheel_url
-          echo "SIGSTORE_WHEEL_PATH=$wheel_path" >>$GITHUB_ENV
+          curl -L "${wheel_url}" -o "${wheel_path}" 
+          echo "SIGSTORE_WHEEL_PATH=${wheel_path}" >> "${GITHUB_ENV}"
       - name: Update requirements
         run: |
           cd install
 
           # Pin on the downloaded wheel, as PyPI might not have updated yet.
-          echo "${{ env.SIGSTORE_WHEEL_PATH }}" >requirements.in
+          echo "${SIGSTORE_WHEEL_PATH}" > requirements.in
           pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in
 
           # Replace requirements.in. People should be able to run the `pip-compile` invocation provided in `requirements.txt`.
-          echo "sigstore==${{ env.SIGSTORE_RELEASE_VERSION }}" >requirements.in
+          echo "sigstore==${SIGSTORE_RELEASE_VERSION}" > requirements.in
 
       - name: Open pull request
         id: pr
         uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04 # v4.2.3
         with:
           title: |
-            Update pinned requirements for ${{ env.SIGSTORE_RELEASE_TAG }}
+            Update pinned requirements for ${SIGSTORE_RELEASE_TAG}
           body: |
-            Pins dependencies for <https://github.com/sigstore/sigstore-python/releases/tag/${{ env.SIGSTORE_RELEASE_TAG }}>.
+            Pins dependencies for <https://github.com/sigstore/sigstore-python/releases/tag/${SIGSTORE_RELEASE_TAG}>.
           commit-message: "[BOT] install: update pinned requirements"
-          branch: "pin-requirements/sigstore/${{ env.SIGSTORE_RELEASE_TAG }}"
+          branch: "pin-requirements/sigstore/${SIGSTORE_RELEASE_TAG}"
           signoff: true
           delete-branch: true
--- a/.github/workflows/pin-requirements.yml
+++ b/.github/workflows/pin-requirements.yml
@@ -70,10 +70,10 @@ jobs:
         uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04 # v4.2.3
         with:
           title: |
-            Update pinned requirements for ${{ inputs.tag }}
+            Update pinned requirements for ${{ env.SIGSTORE_RELEASE_TAG }}
           body: |
-            Pins dependencies for <https://github.com/sigstore/sigstore-python/releases/tag/${{ inputs.tag }}>.
+            Pins dependencies for <https://github.com/sigstore/sigstore-python/releases/tag/${{ env.SIGSTORE_RELEASE_TAG }}>.
           commit-message: "[BOT] install: update pinned requirements"
-          branch: "pin-requirements/sigstore/${{ inputs.tag }}"
+          branch: "pin-requirements/sigstore/${{ env.SIGSTORE_RELEASE_TAG }}"
           signoff: true
           delete-branch: true
--- a/.github/workflows/pin-requirements.yml
+++ b/.github/workflows/pin-requirements.yml
@@ -1,0 +1,48 @@
+name: Pin Requirements
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  update-pinned-requirements:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+        with:
+          ref: main
+
+      - uses: actions/setup-python@d27e3f3d7c64b4bbf8e4abfb9b63b83e846e0435
+        with:
+          python-version: "3.x"
+          cache: "pip"
+          cache-dependency-path: pyproject.toml
+
+      - run: pip install pip-tools
+
+      - name: Determine latest sigstore-python PyPI release
+        run: |
+          version=$(curl -H "Accept: application/vnd.pypi.simple.v1+json" https://pypi.org/simple/sigstore/ | jq -r '.versions[-1]')
+          echo "sigstore_python_version=$version" >>$GITHUB_ENV
+      - name: Update requirements
+        run: |
+          cd install
+          echo "sigstore==${{ env.sigstore_python_version }}" >requirements.in
+          pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in
+      - name: Open pull request
+        id: pr
+        uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04 # v4.2.3
+        with:
+          title: Update pinned requirements for v${{ env.sigstore_python_version }}
+          body: Pins dependencies for [v${{ env.sigstore_python_version }}](https://github.com/sigstore/sigstore-python/releases/tag/v${{ env.sigstore_python_version }}).
+          commit-message: "install: update pinned requirements"
+          branch: "pin-requirements/sigstore/${{ env.sigstore_python_version }}"
+          signoff: true
+          delete-branch: true
+      - if: ${{ steps.pr.outputs.pull-request-number }}
+        run: |
+          echo "Created PR \#${{ steps.pr.outputs.pull-request-number }}"
--- a/.github/workflows/pin-requirements.yml
+++ b/.github/workflows/pin-requirements.yml
@@ -18,7 +18,8 @@ jobs:
 
       - uses: actions/setup-python@d27e3f3d7c64b4bbf8e4abfb9b63b83e846e0435
         with:
-          python-version: "3.x"
+          # NOTE: Update whenever sigstore-python's minimum Python is bumped.
+          python-version: "3.7"
           cache: "pip"
           cache-dependency-path: pyproject.toml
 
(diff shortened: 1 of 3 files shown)


woodruffw:
FWIW, this should also run on the same `release: published` type events that our release workflow (`release.yml`) currently trigger on; that way we don't have to manually trigger it each time we publish a release.

We should also go ahead and disable the equivalent `dependabot.yml` behavior that's currently running -- otherwise it'll conflict with this workflow :slightly_smiling_face: 

> Do we want to use a specific Python version for `pip-compile`? Would this affect the resultant `requirements.txt`?

We should use the minimum supported Python, i.e. 3.7 currently. That should always capture the maxset of possibly resolved dependencies.

woodruffw:
> FWIW, this should also run on the same `release: published` type events that our release workflow (`release.yml`) currently trigger on; that way we don't have to manually trigger it each time we publish a release.

Oh, never mind -- I see how you did this with `workflow_call`. That also works just fine!

.github/workflows/release.yml
```diff
@@ -147,3 +147,11 @@ jobs:
...
             built-packages/*
             smoketest-artifacts/*
+
+  pin-requirements:
+    permissions:
+      pull-requests: write
+      contents: write
+    needs: [release-pypi]
+    if: ${{ !github.event.release.prerelease }}
+    uses: ./.github/workflows/pin-requirements.yml
```
woodruffw, line 157: Nit: Let's add a comment explaining what's happening here :slightly_smiling_face:

woodruffw: (Also, thanks for correctly filtering out prereleases!)

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,48 @@
...
+          cache: "pip"
+          cache-dependency-path: pyproject.toml
+
+      - run: pip install pip-tools
+
+      - name: Determine latest sigstore-python PyPI release
+        run: |
+          version=$(curl -H "Accept: application/vnd.pypi.simple.v1+json" https://pypi.org/simple/sigstore/ | jq -r '.versions[-1]')
+          echo "sigstore_python_version=$version" >>$GITHUB_ENV
+      - name: Update requirements
```
woodruffw, line 31: Nit: introduce some whitespace between these job steps to make them a little more visually distinguishable.

woodruffw, line 31:
```suggestion

      - name: Update requirements
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          cd install
+          echo "sigstore==${{ env.sigstore_python_version }}" >requirements.in
+          pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in
+      - name: Open pull request
```
woodruffw, line 36:
```suggestion

      - name: Open pull request
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          title: Update pinned requirements for v${{ env.sigstore_python_version }}
+          body: Pins dependencies for [v${{ env.sigstore_python_version }}](https://github.com/sigstore/sigstore-python/releases/tag/v${{ env.sigstore_python_version }}).
+          commit-message: "install: update pinned requirements"
+          branch: "pin-requirements/sigstore/${{ env.sigstore_python_version }}"
+          signoff: true
+          delete-branch: true
+      - if: ${{ steps.pr.outputs.pull-request-number }}
+        run: |
+          echo "Created PR \#${{ steps.pr.outputs.pull-request-number }}"
```
woodruffw, line 48: Vestigial?

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          cd install
+          echo "sigstore==${{ env.sigstore_python_version }}" >requirements.in
+          pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in
+      - name: Open pull request
+        id: pr
+        uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04 # v4.2.3
+        with:
+          title: Update pinned requirements for v${{ env.sigstore_python_version }}
+          body: Pins dependencies for [v${{ env.sigstore_python_version }}](https://github.com/sigstore/sigstore-python/releases/tag/v${{ env.sigstore_python_version }}).
```
woodruffw, line 41: Nit: Use YAML's multi-line support for these, maybe. Also don't bother with the titled link; let's just do a direct `<...>`-style link to the release URL.

```suggestion
          title: |
            Update pinned requirements for v${{ env.sigstore_python_version }}
          body: |
            Pins dependencies for <url>.
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,48 @@
...
+          commit-message: "install: update pinned requirements"
```
woodruffw, line 42: Just for visual scanning:

```suggestion
          commit-message: "[BOT] install: update pinned requirements"
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,48 @@
+name: Pin Requirements
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+permissions:
+  pull-requests: write
+  contents: write
```
woodruffw, line 9: Let's move these permissions into the individual job that needs them (there's only one job in this case, so this is just for general best practice).

woodruffw: (Also, a short comment explaining why each permission is present would be good!)

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,48 @@
...
+  update-pinned-requirements:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+        with:
+          ref: main
+
+      - uses: actions/setup-python@d27e3f3d7c64b4bbf8e4abfb9b63b83e846e0435
+        with:
+          python-version: "3.x"
```
woodruffw, line 21: Per comment about minimum Python:

```suggestion
          # NOTE: Update whenever sigstore-python's minimum Python is bumped.
          python-version: "3.7"
```

di: This is currently set here, which ensures anytime we manually update this, we use the right version: https://github.com/sigstore/sigstore-python/blob/main/install/.python-version

I think we need to use something higher than 3.7 to capture all possible dependencies.

woodruffw: It looks like we can install with anything in 3.7 - 3.10; the big constraint is that we can't install 3.11, because we need the `importlib_resources` shim if the installing version is below 3.11:

https://github.com/sigstore/sigstore-python/blob/37483bcbc2b009f58d36103c368533dfe0fc5082/pyproject.toml#L32

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,48 @@
...
+          cache: "pip"
+          cache-dependency-path: pyproject.toml
+
+      - run: pip install pip-tools
+
+      - name: Determine latest sigstore-python PyPI release
+        run: |
+          version=$(curl -H "Accept: application/vnd.pypi.simple.v1+json" https://pypi.org/simple/sigstore/ | jq -r '.versions[-1]')
+          echo "sigstore_python_version=$version" >>$GITHUB_ENV
```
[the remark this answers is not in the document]
tnytown: Hmm, there might be some complexity here. When we run `pip-compile` on a branch's `pyproject.toml`, it doesn't include the hash of `sigstore` itself. Here are a few approaches that we can use to work around this:

- Rebuild the wheel in the`pin-requirements` workflow and pin with the built wheel
  - Is this approach reproducible? I tried my hand at building a wheel locally and seemed to generate two wheels with the same hash.
  - We'd have to duplicate some building logic from the `release` workflow
- Pass in the wheel from `release.yml` with a `workflow_call` input (what will we do with `workflow_dispatch`?)
- Don't check `sigstore`'s hash (probably not a good idea)

I'm leaning towards rebuilding the wheel in `pin-requirements`, since that still allows us to have a manual trigger with `workflow_dispatch`. I'm just not sure if the hash of the resultant wheel is guaranteed to be the same as the one we upload from the `release` workflow.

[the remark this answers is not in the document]
tnytown: Now that you mention CDN availability, the pins in the repo might be updated before new versions of `sigstore` are widely available. Are we comfortable with that propagation delay?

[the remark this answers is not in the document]
di: > Is this approach guaranteed to be reproducible? I tried my hand at building a wheel locally and it seemed to generate two wheels with the same hash.

We don't guarantee reproducibility, so while it might work now, it might not work consistently.

Can't we just take the artifacts off the GitHub release itself?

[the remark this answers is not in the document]
tnytown: > > Is this approach guaranteed to be reproducible? I tried my hand at building a wheel locally and it seemed to generate two wheels with the same hash.
> 
> We don't guarantee reproducibility, so while it might work now, it might not work consistently.
> 
> Can't we just take the artifacts off the GitHub release itself?

Hmm, we should be able to do this. Do we want to keep the `workflow_dispatch` trigger? We'd be able to trigger manual updates on tags with associated releases.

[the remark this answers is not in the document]
di: I'm not sure why it would generally be necessary, but might be good to keep around in case something fails. One of the inputs could just be the tag.

.github/workflows/pin-requirements.yml
tnytown: yep, fixed in latest :)

.github/workflows/pin-requirements.yml
```diff
@@ -2,51 +2,75 @@ name: Pin Requirements
...
       - run: pip install pip-tools
 
-      - name: Determine latest sigstore-python PyPI release
+      - name: Download wheel from GitHub release
         run: |
-          version=$(curl -H "Accept: application/vnd.pypi.simple.v1+json" https://pypi.org/simple/sigstore/ | jq -r '.versions[-1]')
-          echo "sigstore_python_version=$version" >>$GITHUB_ENV
+          release_url="https://api.github.com/repos/sigstore/sigstore-python/releases/tags/${{ inputs.tag }}"
+          # Do some JSON traversal: get the URL of the asset that has a `.whl` file extension.
+          wheel_url=$(curl $release_url | jq -r '.assets[] | select(.name | test(".whl$")) | .browser_download_url')
```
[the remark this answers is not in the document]
tnytown: 😭

.github/workflows/pin-requirements.yml
```diff
@@ -2,51 +2,75 @@ name: Pin Requirements
...
           cache-dependency-path: pyproject.toml
 
       - run: pip install pip-tools
 
-      - name: Determine latest sigstore-python PyPI release
+      - name: Download wheel from GitHub release
         run: |
-          version=$(curl -H "Accept: application/vnd.pypi.simple.v1+json" https://pypi.org/simple/sigstore/ | jq -r '.versions[-1]')
-          echo "sigstore_python_version=$version" >>$GITHUB_ENV
+          release_url="https://api.github.com/repos/sigstore/sigstore-python/releases/tags/${{ inputs.tag }}"
```
woodruffw, line 44: Nitpick: Propagate `inputs.tag` as a env var (`SIGSTORE_RELEASE_TAG`?) for use here; otherwise an attacker who somehow gets control of `inputs.tag` can do template injection.

.github/workflows/pin-requirements.yml
```diff
@@ -2,51 +2,75 @@ name: Pin Requirements
...
+          # Do some JSON traversal: get the URL of the asset that has a `.whl` file extension.
+          wheel_url=$(curl $release_url | jq -r '.assets[] | select(.name | test(".whl$")) | .browser_download_url')
+
+          curl -L "$TMPDIR/sigstore.whl" -L $wheel_url
```
woodruffw: Typo?

```suggestion
          curl -O "$TMPDIR/sigstore.whl" -L $wheel_url
```

tnytown: should be fixed in latest revision, https://github.com/sigstore/sigstore-python/pull/548/files#diff-be794bea173f00a948d2c994c80d73aa9c0655b3b866348068dbb313faa22221R50

woodruffw: Looks good (and thanks for catching my own typo).

.github/workflows/pin-requirements.yml
tnytown: Let me know if this is what you had in mind! https://github.com/sigstore/sigstore-python/pull/548/files/16185e3ef1306bbe0bb92bf8a8f1c83ffc3aa8a7..2be5cdcea3e7a5386b9149e84833a095245b100c

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          python-version-file: install/.python-version
+          cache: "pip"
+          cache-dependency-path: pyproject.toml
+
+      - run: pip install pip-tools
+
+      - name: Compute version from tag
+        run: |
+          echo "SIGSTORE_RELEASE_VERSION=$(echo "${{ env.SIGSTORE_RELEASE_TAG }}" | sed 's/^v//')" >>$GITHUB_ENV
```
woodruffw, line 47:
```suggestion
          echo "SIGSTORE_RELEASE_VERSION=$(echo "${SIGSTORE_RELEASE_TAG}" | sed 's/^v//')" >> "${GITHUB_ENV}"
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Download wheel from GitHub release
+        run: |
+          wheel_name="sigstore-${{ env.SIGSTORE_RELEASE_VERSION }}-py3-none-any.whl"
```
woodruffw, line 51:
```suggestion
          wheel_name="sigstore-${SIGSTORE_RELEASE_VERSION}-py3-none-any.whl"
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,79 @@
...
+          wheel_url="https://github.com/sigstore/sigstore-python/releases/download/${{ env.SIGSTORE_RELEASE_TAG }}/$wheel_name"
```
woodruffw, line 52:
```suggestion
          wheel_url="https://github.com/sigstore/sigstore-python/releases/download/${env.SIGSTORE_RELEASE_TAG}/${wheel_name}"
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,79 @@
...
+          wheel_path="$RUNNER_TEMP/$wheel_name"
```
woodruffw, line 53:
```suggestion
          wheel_path="${RUNNER_TEMP}/${wheel_name}"
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,79 @@
...
+
+          curl -o "$wheel_path" -L $wheel_url
```
woodruffw, line 55:
```suggestion
          curl -L "${wheel_url}" -o "${wheel_path}" 
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,79 @@
...
+          echo "SIGSTORE_WHEEL_PATH=$wheel_path" >>$GITHUB_ENV
```
woodruffw, line 56:
```suggestion
          echo "SIGSTORE_WHEEL_PATH=${wheel_path}" >> "${GITHUB_ENV}"
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Update requirements
+        run: |
+          cd install
+
+          # Pin on the downloaded wheel, as PyPI might not have updated yet.
+          echo "${{ env.SIGSTORE_WHEEL_PATH }}" >requirements.in
```
woodruffw, line 62:
```suggestion
          echo "${SIGSTORE_WHEEL_PATH}" > requirements.in
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,79 @@
...
+          pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in
+
+          # Replace requirements.in. People should be able to run the `pip-compile` invocation provided in `requirements.txt`.
+          echo "sigstore==${{ env.SIGSTORE_RELEASE_VERSION }}" >requirements.in
```
woodruffw, line 66:
```suggestion
          echo "sigstore==${SIGSTORE_RELEASE_VERSION}" > requirements.in
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Open pull request
+        id: pr
+        uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04 # v4.2.3
+        with:
+          title: |
+            Update pinned requirements for ${{ env.SIGSTORE_RELEASE_TAG }}
```
woodruffw, line 73:
```suggestion
            Update pinned requirements for ${SIGSTORE_RELEASE_TAG}
```

.github/workflows/pin-requirements.yml
```diff
@@ -0,0 +1,79 @@
...
+          body: |
+            Pins dependencies for <https://github.com/sigstore/sigstore-python/releases/tag/${{ env.SIGSTORE_RELEASE_TAG }}>.
+          commit-message: "[BOT] install: update pinned requirements"
+          branch: "pin-requirements/sigstore/${{ env.SIGSTORE_RELEASE_TAG }}"
```
woodruffw, line 77:
```suggestion
            Pins dependencies for <https://github.com/sigstore/sigstore-python/releases/tag/${SIGSTORE_RELEASE_TAG}>.
          commit-message: "[BOT] install: update pinned requirements"
          branch: "pin-requirements/sigstore/${SIGSTORE_RELEASE_TAG}"
```

woodruffw (commented):
A bunch of small nitpicks, otherwise LGTM!

.github/workflows/pin-requirements.yml
woodruffw: Yep, exactly what I was thinking! Thanks.

woodruffw (approved):
LGTM! Let's do a pre-release (and manually trigger it) once merged to test this.

woodruffw:
Thanks @tnytown! Feel free to do a pre-release to test this.

Issue #547, opened 2023-03-13
Handle different SCT error cases correctly


woodruffw:
As identified in #543.

PR #555, opened 2023-03-15
sct, keyring: specialize errors

tnytown:
#### Summary

See #547, #543.

In the SCT module, we now distinguish between the following kinds of verification errors:
- Invalid (potentially malicious) input signatures,
- Invalid key ID,
- Malformed input or API misuse.

### DIFF
--- a/sigstore/_internal/sct.py
+++ b/sigstore/_internal/sct.py
@@ -32,7 +32,11 @@ from cryptography.x509.certificate_transparency import (
 from cryptography.x509.oid import ExtendedKeyUsageOID
 
 from sigstore._internal.ctfe import CTKeyring
-from sigstore._internal.keyring import KeyringError, KeyringLookupError
+from sigstore._internal.keyring import (
+    KeyringError,
+    KeyringLookupError,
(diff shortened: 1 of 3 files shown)


woodruffw:
+@tetsuo-cpp for review 🙂

sigstore/_internal/sct.py
```diff
@@ -142,13 +146,34 @@ class InvalidSCTError(Error):
...
+
+            Additional context:
+
+            Message: {str(self)}
+            """
+            + ctx
+        )
+
+
+class InvalidSCTKeyError(Error):
```
tetsuo-cpp, line 163: Is there a reason we shouldn't subclass from `InvalidSCTError` here?

tetsuo-cpp (approved):
LGTM! I left one style related question but if you have a good reason for doing this, feel free to merge as is.

sigstore/_internal/sct.py
tnytown: Ah, good catch! I must've forgotten to.

tetsuo-cpp (approved):

PR #562, opened 2023-03-15
pin-requirements: fix wheel URL

woodruffw:
This should prevent the bogus `file://` requirement.

### DIFF
--- a/.github/workflows/pin-requirements.yml
+++ b/.github/workflows/pin-requirements.yml
@@ -46,22 +46,12 @@ jobs:
         run: |
           echo "SIGSTORE_RELEASE_VERSION=$(echo "${SIGSTORE_RELEASE_TAG}" | sed 's/^v//')" >> "${GITHUB_ENV}"
 
-      - name: Download wheel from GitHub release
-        run: |
-          wheel_name="sigstore-${SIGSTORE_RELEASE_VERSION}-py3-none-any.whl"
-          wheel_url="https://github.com/sigstore/sigstore-python/releases/download/${SIGSTORE_RELEASE_TAG}/${wheel_name}"
-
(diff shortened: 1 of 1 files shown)


.github/workflows/pin-requirements.yml
```diff
@@ -50,16 +50,14 @@ jobs:
...
-          echo "SIGSTORE_WHEEL_PATH=${wheel_path}" >> "${GITHUB_ENV}"
+          echo "SIGSTORE_WHEEL_URL=${wheel_url}" >> "${GITHUB_ENV}"
       - name: Update requirements
         run: |
           cd install
 
-          # Pin on the downloaded wheel, as PyPI might not have updated yet.
-          echo "${SIGSTORE_WHEEL_PATH}" > requirements.in
+          # Pin on the wheel's GitHub URL, as PyPI might not have updated yet.
+          echo "${SIGSTORE_WHEEL_URL}" > requirements.in
```
di, line 60: Ehh, I don't love this. Have we actually experienced PyPI being too slow here, or is that just speculation? I think I'd prefer to not include URLs in this file if possible.

woodruffw: > Ehh, I don't love this. Have we actually experienced PyPI being too slow here, or is that just speculation? I think I'd prefer to not include URLs in this file if possible.

That was speculation from me having problems with it on the GitHub Action. But that might have been sporadic and we can always re-run the action if necessary, so we can drop this.

woodruffw: https://github.com/sigstore/sigstore-python/pull/562/commits/483f4d4d10001a5ca8056a93829fbaaa6ef4d181

di (approved):
Let's see if it works!

tnytown (approved):

PR #571, opened 2023-03-17
sigstore, test: reflow, fix

woodruffw:
Fixes an online OIDC test regression.

### DIFF
--- a/test/unit/test_sign.py
+++ b/test/unit/test_sign.py
@@ -75,11 +75,11 @@ def test_sct_verify_keyring_lookup_error(signer, monkeypatch):
 
     with pytest.raises(
         InvalidSCTError,
-    ) as exc:
+    ) as excinfo:
         signer.sign(payload, token)
 
     # The exception subclass is the one we expect.
-    assert isinstance(exc, InvalidSCTKeyError)
+    assert isinstance(excinfo.value, InvalidSCTKeyError)
 
 
 @pytest.mark.online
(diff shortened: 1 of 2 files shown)


tnytown (dismissed):

test/unit/test_sign.py
```diff
@@ -75,10 +75,12 @@ def test_sct_verify_keyring_lookup_error(signer, monkeypatch):
 
     with pytest.raises(
         InvalidSCTError,
-        match="Invalid key ID in SCT: not found in current keyring",
-    ):
+    ) as excinfo:
         signer.sign(payload, token)
 
+    # The exception subclass is the one we expect.
+    assert isinstance(excinfo.value, InvalidSCTKeyError)
```
tnytown, line 82: can we check for this specific subclass with `pytest.raises`?

woodruffw: We can, but I actually left it that way on purpose -- checking for the parent class being raised means that we haven't broken our exception contract around what this API raises. The `isinstance` check is then just a stand-in for ensuring that we got the right sub-error 🙂

tnytown (approved):

test/unit/test_sign.py
```diff
@@ -75,10 +75,12 @@ def test_sct_verify_keyring_lookup_error(signer, monkeypatch):
 
     with pytest.raises(
         InvalidSCTError,
-        match="Invalid key ID in SCT: not found in current keyring",
```
di, line 78: Aren't we losing a bit of our test by removing this?

woodruffw: > Aren't we losing a bit of our test by removing this?

That's why I added the instance check that follows it -- we're checking that we don't change the top-level exception type with this `pytest.raises`, and then the following check makes sure that the specific exception type is the one reflecting a missing key ID.

(The `match` string no longer exists -- it's baked into the diagnostics now.)

di (approved):

PR #572, opened 2023-03-20
`pin-requirements` workflow self-test

### LINKED ISSUE #563: The `pin-requirements.yml` workflow should test itself
The workflow should ensure that the requirements file it generates is `pip install`-able.

This wouldn't have caught #562 since the `file://` would have been local to the runner, but it will still be a good resilience test to have.

tnytown:
#### Summary

We now invoke the `requirements` workflow from `pin-requirements` to test install-ability. If any install fails, the requirements pinning PR is not opened.

Fixes #563.

### DIFF
--- a/.github/workflows/pin-requirements.yml
+++ b/.github/workflows/pin-requirements.yml
@@ -65,7 +65,7 @@ jobs:
           git commit --all -s -m "[BOT] install: update pinned requirements"
           git push -f origin "main:${SIGSTORE_NEW_BRANCH}"
 
-  test-install:
+  test-requirements:
     needs: update-pinned-requirements
     uses: ./.github/workflows/requirements.yml
     with:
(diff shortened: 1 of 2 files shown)


.github/workflows/pin-requirements.yml
```diff
@@ -53,15 +60,44 @@ jobs:
           echo "sigstore==${SIGSTORE_RELEASE_VERSION}" > requirements.in
           pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in
 
+      - name: Commit changes and push to branch
+        run: |
+          git commit --all -s -m "[BOT] install: update pinned requirements"
+          git push -f origin "main:${SIGSTORE_NEW_BRANCH}"
+
+  test-install:
```
woodruffw, line 68:
```suggestion
  test-requirements:
```

woodruffw (approved):
LGTM!

PR #573, opened 2023-03-20
workflows/pin-requirements: fix job name

tnytown:
#572: see https://github.com/sigstore/sigstore-python/actions/runs/4472434361


woodruffw (approved):
LGTM, thanks for the quick fix!

PR #576, opened 2023-03-22
Adapt ambient OIDC tests to support interactive flow for local testing

### LINKED ISSUE #570: Reduce our dependency on ambient OIDC tests for error cases
Our ambient OIDC tests consistently catch regressions, but they only run on first-party merges and PRs. As such, they tend to catch regressions *after* they make it to `main`.

We should reduce our dependency on the ambient OIDC credential in some of our error testcases. In particular, things like SCT error case handling should be test-able offline, since we can store the SCT as a test asset.

tnytown:
#### Summary

Introduces a new Makefile target `test-oidc`. This runs ambient OIDC tests with tokens from `sigstore get-identity-token`.

Partially fixes #570.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,8 +80,14 @@ reformat: $(VENV)/pyvenv.cfg
 .PHONY: test
 test: $(VENV)/pyvenv.cfg
 	. $(VENV_BIN)/activate && \
-		pytest --cov=$(PY_MODULE) $(T) $(TEST_ARGS) && \
+		$(TEST_ENV) pytest --cov=$(PY_MODULE) $(T) $(TEST_ARGS) && \
 		python -m coverage report -m $(COV_ARGS)
+
+.PHONY: test-oidc
+test-oidc: TEST_ENV += \
+	SIGSTORE_IDENTITY_TOKEN_production=$$($(MAKE) -s run ARGS="get-identity-token") \
+	SIGSTORE_IDENTITY_TOKEN_staging=$$($(MAKE) -s run ARGS="--staging get-identity-token")
+test-oidc: test
 
 .PHONY: doc
 doc: $(VENV)/pyvenv.cfg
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -80,9 +80,15 @@ reformat: $(VENV)/pyvenv.cfg
...
 	. $(VENV_BIN)/activate && \
-		pytest --cov=$(PY_MODULE) $(T) $(TEST_ARGS) && \
+		$(TEST_ENV) pytest --cov=$(PY_MODULE) $(T) $(TEST_ARGS) && \
 		python -m coverage report -m $(COV_ARGS)
 
+.PHONY: test-oidc
+test-oidc: TEST_ENV += \
+	SIGSTORE_IDENTITY_TOKEN_production=$$($(MAKE) -s run ARGS="get-identity-token") \
+	SIGSTORE_IDENTITY_TOKEN_staging=$$($(MAKE) -s run ARGS="--staging get-identity-token")
+test-oidc: test
```
woodruffw, line 90: Nit: Let's name this `test-interactive`, to emphasize that it'll require user interaction.

sigstore/oidc.py
```diff
@@ -125,11 +126,12 @@ def identity_token(  # nosec: B107
...
             # Launch web browser
             if not force_oob and webbrowser.open(server.base_uri):
-                print("Waiting for browser interaction...")
+                print("Waiting for browser interaction...", file=sys.stderr)
             else:
                 server.enable_oob()
                 print(
-                    f"Go to the following link in a browser:\n\n\t{server.auth_endpoint}"
+                    f"Go to the following link in a browser:\n\n\t{server.auth_endpoint}",
+                    file=sys.stderr,
```
woodruffw, line 134: Nice!

woodruffw:
Let's also document this new `make` target in the `CONTRIBUTING.md`, as another test entrypoint.

woodruffw (approved):
LGTM. `get-identity-token` should also work automatically with ambient credentials, so it might be good to add another test job in CI that makes sure we don't accidentally regress the `make test-interactive` target.

woodruffw (approved):

PR #587, opened 2023-03-28
Aggregate offline and online test coverage.

jleightcap:
Use `pytest --cov-append` to not overwrite the `.coverage` artifact on each `make test` run. Aggregation avoids the problem of only the *last sequential test suite* ran being account for by getting last say on the `.coverage` contents.

Tracking against: https://github.com/sigstore/sigstore-python/actions/runs/4517097044#summary-12269011312

With aggregation: https://github.com/sigstore/sigstore-python/actions/runs/4547521585

Coverage increased by... a whole 1%

(note: thanks for the pointer, @0xalpharush!)


woodruffw (approved):

woodruffw:
Thanks @jleightcap!

PR #591, opened 2023-03-30
Unit testing for bundled trust root getters

### LINKED ISSUE #580: More unit testing for bundled trust root changes
**Description**

#542 added support for TUF "bundled" trust roots. We need tests to cover the new surface, namely:

- [x] the  `get` methods
    - [x] `get_rekor_keys`
    - [x] `get_fulcio_certs`
    - [x] `get_ctfe_keys`
- [ ] `_get_trusted_root`

tnytown:
This changeset exercises the bundled trusted root codepaths in `get_rekor_keys`, `get_ctfe_keys`, and `get_fulcio_certs` with new offline tests. We also yakshave the TUF module further.

Resolves #580.


test/unit/internal/test_tuf.py
```diff
@@ -120,6 +124,44 @@ def test_updater_staging_get(mock_staging_tuf, tuf_asset):
...
+            )
+            for k in keys
+        ]
+
+    updater = TrustUpdater.production()
+
+    # The test should use the TUF path, so we stub out the legacy getter here.
+    monkeypatch.setattr(updater, "_get", lambda usage, statuses: [])
+
+    assert _der_keys(updater.get_ctfe_keys()) == _pem_keys([_contents("ctfe_2022.pub")])
```
tetsuo-cpp, line 153: Just so I know what to expect. When there's a key rotation, we should expect these to begin failing right? And in that case, we should refresh the keys and certs under `assets/prod-tuf`.

tetsuo-cpp: Hmm, I'm not too bothered by it. If it becomes annoying then we can look into monkey patching the time, but for now I'd say we can switch to staging assets when it's ready and pull the trigger on this.

test/unit/conftest.py
```diff
@@ -174,8 +195,6 @@ def _fetch(self, url: str) -> Iterator[bytes]:
             if filepath.is_file():
                 success[filepath] += 1
                 return BytesIO(filepath.read_bytes())
-            else:
-                print(f"POO {filepath}")
```
tetsuo-cpp, line 178: 😂

tetsuo-cpp (approved):
LGTM!

PR #609, opened 2023-04-11
swap over to using CDN to fetch TUF roots

bobcallaway:
#### Summary
The TUF roots for both production and staging instances of sigstore are now addressable via a CDN endpoint; this change uses the CDN instead of calling GCS directly via its HTTP endpoint.


woodruffw (approved):
Thanks!

PR #611, opened 2023-04-13
tuf: embed trusted root target

### LINKED ISSUE #608: Embed TUF targets to minimize downloads
**Description**

For the Golang library, we embed a copy of the [target files](https://github.com/sigstore/sigstore/tree/main/pkg/tuf/repository/targets), and check if the target file matches the target metadata [here](https://github.com/sigstore/sigstore/blob/main/pkg/tuf/client.go#L522-L527) (a necessary check since the target file may have been updated and would need to be downloaded). Would it be possible to do the same for sigstore-python?

tnytown:
This changeset embeds `trusted_root.json` for staging and prod and adjusts tests to reflect this.

Additionally, we tweak the logistics around embedded assets.

Resolves #608.


sigstore/_internal/tuf.py
```diff
@@ -121,24 +121,35 @@ def __init__(self, url: str) -> None:
...
-                fname = "staging-root.json"
-            else:
-                raise Exception(f"TUF root not found in {tuf_root}")
-
-            self._metadata_dir.mkdir(parents=True, exist_ok=True)
-            root_json = read_embedded(fname)
-            with tuf_root.open("wb") as io:
-                io.write(root_json)
+
+        try:
```
tetsuo-cpp, line 136: Should this wrapped with a:
```python
if not tuf_root.exists():
```
and similar below?

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ All versions prior to 0.9.0 are untracked.
 
 ### Changed
 
+* A cached copy of the trust bundle is now included with the distribution.
```
tetsuo-cpp, line 26: Not a big deal but we've been putting line breaks between these dot points up until now.

woodruffw (approved):

woodruffw:
LGTM!

PR #614, opened 2023-04-14
_cli: emit only sigstore bundle by default

### LINKED ISSUE #592: CLI: stop emitting `.sig` and `.crt` signing outputs
We should eventually (in the next major version?) output only the `.sigstore` bundle as a signing artifact. Before that, we may want to consider adding a `--bundle-only` option, complementary to `--no-bundle`, that emulates this behavior.

Follow-up to #465.

tnytown:
- Adjust CLI to emit only `{input}.sigstore` by default. Signature and certificate outputs are still supported, but users will have to pass the respective flags to generate them.
- Remove `--no-bundle`, as its functionality now overlaps with `--no-default-files`.
- Fail on multiple input case when `--bundle` is specified.
- Adjust CLI help to reflect all of the above.

Would we also want to change the stability of the `--bundle` option within the release?

Resolves #592.


woodruffw (approved):
LGTM!

PR #625, opened 2023-04-23
_cli: allow named pipes as inputs

### LINKED ISSUE #616: [BUG] Named pipes do not work as `sigstore sign` inputs
...they also might not work for other subcommand inputs, but this is the first one I checked with.

Expected behavior:

```bash
# signs for the content 'hello' successfully
$ sigstore sign <(echo -n 'hello')
```

Actual behavior:

```bash
$ sigstore sign <(echo -n 'hello')

usage: sigstore [-h] [-V] [-v] [--staging] [--rekor-url URL]
                [--rekor-root-pubkey FILE]
                {sign,verify,get-identity-token} ...
sigstore: error: Input must be a file: /dev/fd/63
```

woodruffw:
Adds a `_path_is_filelike` helper to abstract the check away here.

Fixes #616.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ All versions prior to 0.9.0 are untracked.
   Sigstore bundles are now preferred.
   ([#614](https://github.com/sigstore/sigstore-python/pull/614))
 
+### Fixed
+
+* The `sigstore` CLI now handles named pipes correctly
+  ([#625](https://github.com/sigstore/sigstore-python/pull/625))
+
 ## [1.1.2]
 
 ### Fixed
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -65,6 +65,14 @@ package_logger = logging.getLogger("sigstore")
 package_logger.setLevel(os.environ.get("SIGSTORE_LOGLEVEL", "INFO").upper())
 
 
+def _path_is_filelike(path: Path) -> bool:
+    """
+    A `pathlib.Path` helper for checking whether a path is "file-like,"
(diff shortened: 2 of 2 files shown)


woodruffw:
Whoops, I underthought this:

```
    sys.exit(main())
             ^^^^^^
  File "/Users/william/devel/sigstore-python/sigstore/_cli.py", line 543, in main
    _sign(args)
  File "/Users/william/devel/sigstore-python/sigstore/_cli.py", line 691, in _sign
    with outputs["bundle"].open(mode="w") as io:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/william/.pyenv/versions/3.11.2/lib/python3.11/pathlib.py", line 1044, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/dev/fd/63.sigstore'
```

Will think some more about whether this use case actually makes sense.

tetsuo-cpp:
I think we can probably do without this. I don't see a use-case for this other than a "hello world" example.

For `pip-audit`, I think it makes sense since it's an shorthand way to audit a package that you might be curious about. But for `sigstore-python`, the only reason you would sign an artifact would be so it can be verified at a later point in time, in which case the artifact would need to be uploaded somewhere. So realistically, you're always going to have a file. I might just be lacking imagination though. 😆

woodruffw:
Closing because this was ill-conceived.

PR #626, opened 2023-04-23
tuf: remove non-trusted-root handling paths

### LINKED ISSUE #584: TUF: Remove support for non-bundled roots of trust
Per conversation in #542: the bundled trust root is now considered the stable interface for retrieving all of the various pieces of Sigstore's root of trust, so we can remove support for the "standalone" targets that it replaced.

We should do #580 first, to convince ourselves that we haven't introduced any bugs by switching over to the bundled trust root.

woodruffw:
Fixes #584.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,16 @@ All versions prior to 0.9.0 are untracked.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


tetsuo-cpp (approved):
LGTM!

PR #633, opened 2023-04-25
verify: fix timerange inclusion check

woodruffw:
Our previous check was incorrect: it would fail verification on an otherwise valid inclusion proof if the integrated time happened to be at the last moment of `not_valid_after`.

See: https://github.com/sigstore/protobuf-specs/pull/78


tetsuo-cpp (approved):
LGTM!

PR #634, opened 2023-04-26
Root hash signature verification v2

### LINKED ISSUE #248: Include Root Hash verification when verifying a Rekor Inclusion Proof

tnytown:
Now with Signed Note verification!

Thanks to @jleightcap for doing all of the work! I'm just here to push it over the finish line :)

Supersedes #527.
Fixes #248.


sigstore/_internal/merkle.py
```diff
@@ -133,3 +135,38 @@ def verify_merkle_inclusion(entry: LogEntry) -> None:
...
+    Verify the inclusion proof's checkpoint.
+    """
+
+    inclusion_proof = entry.inclusion_proof
+    if inclusion_proof is None:
+        raise InvalidInclusionProofError("Rekor entry has no inclusion proof")
+    if inclusion_proof.checkpoint is None:
+        return
+
+    # verififcaiton occurs in two stages:
```
woodruffw, line 151: Typo:

```suggestion
    # Verification occurs in two stages:
```

sigstore/_internal/rekor/checkpoint.py
```diff
@@ -0,0 +1,202 @@
...
+class RekorSignature:
+    """
+    Represents a `RekorSignature` containing:
+    - the name of the signature, e.g. "rekor.sigstage.dev"
+    - the signature hash
+    - the base64 signature
+    """
+
+    # FIXME(jl): this does not feel like a de novo definition...
+    # does this exist already in sigstore-python (or its depenencices)?
```
woodruffw, line 44: I don't believe this exists anywhere, so this comment is safe to remove.

OTOH, we might want to rename this to `RekorCheckpointSignature` or similar, since there are other signatures that happen in the context of Rekor.

sigstore/_internal/rekor/checkpoint.py
```diff
@@ -0,0 +1,202 @@
...
+            raise ValueError("Malformed Note: must contain at least one signature!")
+        if data[-1] != "\n":
+            raise ValueError("Malformed Note: data section must end with newline!")
+
+        sig_parser = re.compile(r"\u2014 (\S+) (\S+)\n")
+        signatures: list[RekorSignature] = []
+        for name, signature in re.findall(sig_parser, data):
+            signature_bytes: bytes = base64.b64decode(signature)
+            if len(signature_bytes) < 5:
+                raise ValueError("Malformed Note: signature contains too few bytes")
```
woodruffw, line 158: These `ValueError`s might benefit from being dedicated exception types.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ All versions prior to 0.9.0 are untracked.
 
 ### Changed
 
+* `sigstore verify` now verifies signature timestamps with hashes signed
+  by Rekor.
+  ([#634](https://github.com/sigstore/sigstore-python/pull/634))
```
woodruffw, line 15:
```suggestion
* `sigstore verify` now performs additional verification of Rekor's inclusion proofs by cross-checking them against signed checkpoints
  ([#634](https://github.com/sigstore/sigstore-python/pull/634))
```

sigstore/sign.py
```diff
@@ -238,6 +239,12 @@ def _to_bundle(self) -> Bundle:
                 hashes=[
                     bytes.fromhex(h) for h in self.log_entry.inclusion_proof.hashes
                 ],
+                checkpoint=Checkpoint(
+                    envelope=self.log_entry.inclusion_proof.checkpoint or ""
+                )
+                # FIXME(jl): checkpoint should be a serialzied field here.
+                # this causes tests depending on the commited `bundle.txt.sigstore` to fail;
+                # the bundle doesn't contain the `"checkpoint"` field.
```
woodruffw, line 247: Is this comment still relevant?

sigstore/sign.py
```diff
@@ -238,6 +239,12 @@ def _to_bundle(self) -> Bundle:
                 hashes=[
                     bytes.fromhex(h) for h in self.log_entry.inclusion_proof.hashes
                 ],
+                checkpoint=Checkpoint(
+                    envelope=self.log_entry.inclusion_proof.checkpoint or ""
+                )
```
woodruffw, line 244: This fallback shouldn't be necessary anymore, right?

```suggestion
                checkpoint=Checkpoint(
                    envelope=self.log_entry.inclusion_proof.checkpoint
                )
```

sigstore/verify/models.py
```diff
@@ -239,20 +258,32 @@ def from_bundle(
...
+            if not checkpoint.envelope:
+                raise InvalidMaterials("expected checkpoint in inclusion proof")
+
+            parsed_inclusion_proof = LogInclusionProof(
+                checkpoint=checkpoint.envelope,
+                hashes=[h.hex() for h in inclusion_proof.hashes],
+                log_index=inclusion_proof.log_index,
+                root_hash=inclusion_proof.root_hash.hex(),
+                tree_size=inclusion_proof.tree_size,
+            )
```
woodruffw, line 278:
```suggestion

        # NOTE: We only load the inclusion proof from the bundle if we know that
        # we're performing an offline verification, since that's the only case
        # in which we'll use it.
        inclusion_proof = tlog_entry.inclusion_proof
        parsed_inclusion_proof: LogInclusionProof | None = None
        if inclusion_proof and offline:
            checkpoint = inclusion_proof.checkpoint

            # If the bundle to be verified offline includes an inclusion proof,
            # we verify it, which requires a checkpoint.
            if not checkpoint.envelope:
                raise InvalidMaterials("expected checkpoint in inclusion proof")

            parsed_inclusion_proof = LogInclusionProof(
                checkpoint=checkpoint.envelope,
                hashes=[h.hex() for h in inclusion_proof.hashes],
                log_index=inclusion_proof.log_index,
                root_hash=inclusion_proof.root_hash.hex(),
                tree_size=inclusion_proof.tree_size,
            )
```

sigstore/sign.py
```diff
@@ -238,6 +239,12 @@ def _to_bundle(self) -> Bundle:
                 hashes=[
                     bytes.fromhex(h) for h in self.log_entry.inclusion_proof.hashes
                 ],
+                checkpoint=Checkpoint(
+                    envelope=self.log_entry.inclusion_proof.checkpoint
+                )
+                # FIXME(jl): checkpoint should be a serialzied field here.
+                # this causes tests depending on the commited `bundle.txt.sigstore` to fail;
+                # the bundle doesn't contain the `"checkpoint"` field.
```
woodruffw, line 247:
```suggestion
```

sigstore/_internal/rekor/checkpoint.py
```diff
@@ -0,0 +1,229 @@
...
+                self.log_hash,
+            ]
+            + self.other_content
+        )
+
+
+@dataclass(frozen=True)
+class SignedNote:
+    """
+    Represents a signed `Note` containing a note and its corresponding list of signatures.
```
woodruffw, line 111: There's no `Note` class, so maybe just:

```suggestion
    Represents a "signed note" containing a note and its corresponding list of signatures.
```

sigstore/_internal/rekor/checkpoint.py
```diff
@@ -0,0 +1,229 @@
...
+                sig_hash=struct.unpack(">4s", signature_bytes[0:4])[0],
+                signature=base64.b64encode(signature_bytes[4:]),
+            )
+            signatures.append(signature)
+
+        return cls(note=header, signatures=signatures)
+
+    def verify(self, client: RekorClient, key_id: KeyID) -> None:
+        """
+        Verify the SignedNote with using the given RekorClient by verifying each contained signature.
```
woodruffw, line 170:
```suggestion
        Verify the `SignedNote` with using the given RekorClient by verifying each contained signature.
```

sigstore/_internal/rekor/checkpoint.py
```diff
@@ -0,0 +1,229 @@
...
+                    key_id=key_id, signature=base64.b64decode(sig.signature), data=note
+                )
+            except KeyringSignatureError as sig_err:
+                raise CheckpointError("invalid signature") from sig_err
+
+
+@dataclass(frozen=True)
+class SignedCheckpoint:
+    """
+    Represents a *signed* `Checkpoint`: a `LogCheckpoint` and its corresponding *signed* `Note`.
```
woodruffw, line 190:
```suggestion
    Represents a *signed* `Checkpoint`: a `LogCheckpoint` and its corresponding `SignedNote`.
```

sigstore/_internal/rekor/checkpoint.py
```diff
@@ -0,0 +1,229 @@
...
+    Represents a signed `Note` containing a note and its corresponding list of signatures.
+    """
+
+    note: StrictStr = Field(..., alias="note")
+    signatures: list[RekorSignature] = Field(..., alias="signatures")
+
+    @classmethod
+    def from_text(cls, text: str) -> SignedNote:
+        """
+        Serialize from a bundled text 'note'.
```
woodruffw, line 120:
```suggestion
        Deserialize from a bundled text 'note'.
```

sigstore/verify/verifier.py
```diff
@@ -246,17 +250,31 @@ def verify(
...
                 return VerificationFailure(
                     reason=f"invalid Rekor inclusion proof: {exc}"
                 )
+
+            try:
+                verify_checkpoint(self._rekor, entry)
+            except CheckpointError as exc:
+                return VerificationFailure(reason=f"invalid Rekor root hash: {exc}")
+
+            logger.debug("Successfully verified inclusion proof...")
```
woodruffw, line 268: Spacing:

```suggestion
            logger.debug("Successfully verified inclusion proof...")

```

woodruffw (approved):
This LGTM, but I'd appreciate a review from @asraa or @haydentherapper if either of you has the time!

In particular: we now verify the signature that comes with the checkpoint and cross-check the signed data (i.e. the root hash) against the inclusion proof's. We do this in two circumstances:

1. Whenever we're doing online verification (the default)
2. During offline verification, *if* the input bundle contains an inclusion proof (including a checkpoint)

haydentherapper:
Ack, will take a look tomorrow!

haydentherapper:
Sorry, had a busy week! Will look first thing Monday.

woodruffw:
No problem at all! And please let us know if we can do anything to help with review as well 🙂

haydentherapper (commented):
Looks really great! Just a couple small questions.

di:
> Just a couple small questions.

@haydentherapper Doesn't look like your questions made it into your review, FYI.

sigstore/verify/models.py
```diff
@@ -240,19 +259,34 @@ def from_bundle(
             )
         tlog_entry = tlog_entries[0]
 
-        inclusion_proof = LogInclusionProof(
-            log_index=tlog_entry.inclusion_proof.log_index,
-            root_hash=tlog_entry.inclusion_proof.root_hash.hex(),
-            tree_size=tlog_entry.inclusion_proof.tree_size,
-            hashes=[h.hex() for h in tlog_entry.inclusion_proof.hashes],
-        )
+        # NOTE: We only load the inclusion proof from the bundle if we know that
```
haydentherapper, line 262: Nit, should this note say for both online and offline verification?

sigstore/verify/verifier.py
```diff
@@ -246,17 +250,32 @@ def verify(
 
         # 5) Verify the inclusion proof supplied by Rekor for this artifact.
         #
-        # We skip the inclusion proof only if explicitly requested.
-        if not materials._offline:
+        # The inclusion proof should always be present in the online case. In
+        # the offline case, if it is present, we verify it.
```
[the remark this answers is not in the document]
asraa: I think it may have been optional because it doesn't provide any value over the SET when the client doesn't have any local state.

sigstore/verify/models.py
woodruffw: Yep -- we always retrieve the inclusion proof from an online Rekor instance unless the user has explicitly told us to prefer offline sources with `--offline`.

We did that because it was the only way to get the inclusion proof before we had bundles, but maybe that's no longer necessary -- is there client guidance around when clients should prefer online lookups versus in-situ bundle materials?

CC @znewman01

sigstore/verify/verifier.py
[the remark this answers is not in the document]
woodruffw: Just to clarify: this means that transparency log entries in turn are mandatory within bundles, right? Since the inclusion proof appears within each of them 🙂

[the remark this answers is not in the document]
woodruffw: Opened https://github.com/sigstore/protobuf-specs/pull/79.

sigstore/verify/models.py
haydentherapper: Both online and offline give you the same guarantees, trusting what comes from Rekor. Online has a few negatives: Reliability (since Rekor must be online), and privacy (revealing what entries you care about to Rekor). I'd say to prefer the bundle when presented.

Long term, there's a few options for stronger guarantees of inclusion:
* Online querying of distributors (gatherers of witness statements) or witnesses themselves, but this comes with some privacy tradeoffs
* Offline bundles containing a set of witnessed checkpoints

The first is just changing client behavior, the second would be a bundle spec update.

In terms of the client, I'd prefer offline over online, mostly from a performance/reliability standpoint.

sigstore/verify/verifier.py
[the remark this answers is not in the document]
woodruffw: Ah, GitHub didn't show those last two comments for me...

In that case, does it make sense for the log entries to be mandatory (i.e., there should be at least one), but with optional interior inclusion proofs? If so, I can update the protobuf-specs PR.

sigstore/verify/models.py
woodruffw: Makes sense, thank you! I'll tweak to reflect that.

sigstore/verify/verifier.py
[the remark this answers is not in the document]
haydentherapper: I think required for all seems reasonable. We're probably going to need a new version for future bundle updates around witnessing anyways as I bet they'll be breaking.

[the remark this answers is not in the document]
haydentherapper: edit: I'm also good with having inclusion proofs be optional because of the above, and basically tacklign this later on (maybe in a non breaking way)

sigstore/verify/models.py
woodruffw: Sorry, one last question (since I'm getting myself twisted up with online/offline again): isn't an inclusion proof retrieved from the bundle no stronger than an offline SET verification?

In other words: an adversary who can issue valid-looking SETs can also produce an valid-looking (from an offline context) inclusion proofs, including signed checkpoints on those inclusion proofs. 

So, in an "offline" context, is there any value to doing anything other than the SET verification? Correspondingly, in an "online" context, doesn't the inclusion proof + checkpoint _need_ to be retrieved online to ensure that it hasn't been forged from a split view of the log?

haydentherapper: checkpoint-verify-kanon

checkpoint-verify-offline

woodruffw: Makes sense, thank you! I think that ties a bow around it for me; I can refactor this to do the following:

1. Offline mode: require SET, opportunistically verify the inclusion proof and checkpoint if present
2. Online mode: require SET, require inclusion proof and checkpoint (retrieving them from Rekor if not present in the bundle)

woodruffw (dismissed):

woodruffw (approved):

Issue #567, opened 2023-03-16
Signing should support issuer and identity arguments too 


jku:
**Issue**

Situation is this
* application that uses sigstore-python knows that it wants a signature from a specific certificate identity/issuer
* triggering the browser authentication via sigstore-python forces the user choose the issuer/identity
* user can easily select the wrong one
* it is difficult for the application to check if the token is for the correct identity (as the token is just a blob from application perspective)

**Proposal**

**EDIT: Please see comment below for maybe a better proposal**

~~Signing should support issuer and identity arguments just like verification does.~~

~~In practice:~~
~~* `Issuer.identity_token()` could accept optional arguments `identity` and `issuer`~~
~~* `Issuer.identity_token()` could check the token contents and fail if they do not match the arguments (note that issuer here refers to the issuer that will be used in the certificate so the check needs to be against the _federated_ issuer if there is one, not the oidc issuer)~~
~~* this could be extended to the UI as well: `sigstore sign` could have `--cert-oidc-issuer` and `--cert-identity` options~~

~~In future identity_token() could make things easier for user too:~~
~~* could use the OIDC optional argument `login_hint=identity` -- unfortunately this is not supported by dex (the sigstore oidc proxy) currently~~
~~* dex apparently has a non-standard `connector` which could be used to pre-select the issuer~~

jku:
> Issuer.identity_token() could accept optional arguments identity and issuer

Thinking about this further, the same issue exists in the ambient credential case... which the identity_token() change would not help with.

So either `id.detect_credentials()` should have a similar change or the optional arguments actually belong to `Signer.sign()` which could then check that the provided token matches optional identity arguments before requesting a signing certificate.

Modifying the Signer API has the downside that the potential future improvements to the OIDC proxy would still require separate changes in the token-fetching components... but conceptually it feels correct that Signer would check that the token is what it should be: it doesn't matter if in the future a "login_hint" is passed to the OIDC proxy or not, Signer should still check that the token is as expected.

jku:
I have an actual proposal now. Wrote a whole document to make it clearer to myself: https://docs.google.com/document/d/1VNCXhUW_DSm0jG8ZmzhY2AfZo8KUUTDv4KbvHqI1yH8/edit?usp=sharing

The summary is that I think `token` should be a type that exposes the certificate identity and issuer to the calling application. That would allow the application to make decisions based on the token identity like this:

```python
issuer = Issuer.production()
token = issuer.identity_token()
[...]

PR #635, opened 2023-04-26
sigstore: refactor, use IdentityToken everywhere

woodruffw:
Key changes:

* `sigstore._internal.oidc.Identity` is now stabilized as `sigstore.oidc.IdentityToken`, and provides attributes similar to the ones originally requested/proposed by @jku in #567
* The top level `Signer.sign(...)` API now takes an `IdentityToken` rather than a raw `str`
* The refactored `IdentityToken` class now uses the correct `IdentityError` exception type (the first party one, rather than `id.IdentityError`
* The (internal) Fulcio CSR API now takes an `IdentityToken` rather than a raw `str`
* The (internal) CLI helpers now use and return `IdentityToken` rather raw `str` objects
* The (internal) CLI helpers now include `_die`, which improves our type-checking of the CLI

See #567.

CC @jku


sigstore/oidc.py
```diff
@@ -56,6 +58,91 @@ class _OpenIDConfiguration(BaseModel):
...
+        Returns this `IdentityToken`'s underlying subject.
+        """
+        return self._identity
+
+    @property
+    def issuer(self) -> str:
+        """
+        Returns a URL identifying this `IdentityToken`'s issuer.
+        """
+        return self._issuer
```
jku, line 127: I believe this returns the OIDC "issuer" which in a typical sigstore interactive login case is `https://oauth2.sigstore.dev/auth`. I think I am interested in the "federated connector" (or whatever the correct term is): the "issuer" that ends up written into the signing certificate itself, e.g. `https://github.com/login/oauth`

jku (commented):
I think this looks great from my perspective, apart from one thing: I am interested in the _**certificate** identity issuer_ which is not necessarily "iss" in the JWT. 

So as an example, in this example JWT below the certificate identity details I'm looking for are `email` and `federated_claims.connector_id`. I assume in other cases `iss` is really the certificate issuer.
```
{
  'iss': 'https://oauth2.sigstore.dev/auth', 
  'sub': '...', 
  'aud': 'sigstore',
  'exp': 1682590933,
  'iat': 1682590873, 
  'nonce': '...', 
  'at_hash': '...',
  'email': 'jku@goto.fi',
  'email_verified': True,
  'federated_claims': {
    'connector_id': 'https://github.com/login/oauth',
    'user_id': '...'
  }
}
```

sigstore/oidc.py
woodruffw: I'll double-check on this!

woodruffw: Yep, confirmed.

@jku would a `federated_connector_id: str | None` work for you here? I assume we'll need to accommodate the `None` case for identity tokens that don't go through a second hoop like the Sigstore/Sigstage OAuth connector.

jku: So this means there are situations where the application would have to somehow choose when too look at `federated_connector_id` and when to look at `issuer`? It feels like this still requires the application to make decisions that require knowing sigstore implementation details, and that the sigstore library would be a better place for that logic: If it's possible I would rather have a `certificate_issuer` (or something that always contains the issuer that will be inserted into the signing certificate).

woodruffw: > So this means there are situations where the application would have to somehow choose when too look at `federated_connector_id` and when to look at `issuer`? 

Yes, I think so.

> It feels like this still requires the application to make decisions that require knowing sigstore implementation details, and that the sigstore library would be a better place for that logic: If it's possible I would rather have a `certificate_issuer` (or something that always contains the issuer that will be inserted into the signing certificate).

`certificate_issuer` sounds a little funny to me here (I could be underthinking it!) since the identity token at this point is conceptually untied to any certificate issuance -- Fulcio *itself* is the issuer and will issue a certificate with `federated_connector_id` as an *extension* value, but only if it happens to accept this particular OIDC token (which we can't guarantee a priori).

I'm not sure if there's a clean way for `sigstore-python` to handle this for users, unless we get additional guarantees from Fulcio about whether `federated_connector_id` is _always_ the value that ends up becoming the underlying issuer extension value. If they can guarantee that, then IMO an attribute like `federated_issuer` or similar would make sense here.

CC @haydentherapper and @asraa for that last point 🙂

woodruffw: (Separately, I think a user could still use this without knowing too much about Sigstore's internals: if a user would like to completely forbid an IdP at `hxxps://example.com`, then *neither* `issuer` nor `federated_connector_id` should ever match that value.)

tetsuo-cpp (approved):
LGTM! I think just the point about `federated_connector_id` needs to be addressed. But otherwise, the refactor looks great.

sigstore/oidc.py
haydentherapper: Very interesting points!

The federated connector ID is an implementation detail of Sigstore and one that I wouldn’t expect clients would need to be aware of. I would assume clients simply pass through the identity token they’re given and the receiving party (fulcio) will handle the claims how it chooses. 

If a client wanted to verify the identity token before use, then the token no longer is just a blob and you’d need to know the expected claims per issuer. This is hard though because there’s so many different permutations - a subject may be in sub, or email (or for CI, so many options!). Issuer is in iss, but should the client know that dex is a federated provider and therefore you need to also look into the federated claim? 

 My overall comment is “it depends” to whether the client should know about federation or not. The simplest view is that an ID token is just a blob that gets returned from the OIDC dance and passed to fulcio. If you do want to verify a token before use, maybe you push this to a policy later, and let users specify which claims they want to verify? Otherwise you could have a public good infrastructure default policy that’s aware of federated issuer, but there’s still the question of where identity comes from.

Related, how are you picking which subject to sign for the challenge? Or do you just use CSRs, which don’t require a particular subject signed over?

woodruffw: Thanks for the detailed response @haydentherapper!

Yeah, this is a hard position for the client to be in: @jku has presented a reasonable use case in which a user might want to sanity-check an OIDC identity before sending it along to Fulcio, but I don't think there's a straightforward way for us to consistently guarantee the presence of the fields that the user might expect (either because of Dex, or an unusual but valid claim set, etc.).

I _think_ the best we can do is exposing the `federated_issuer` when we happen to have it, but it's possibly that that's also a Dex implementation detail and I'm not realizing that. 


> Related, how are you picking which subject to sign for the challenge? Or do you just use CSRs, which don’t require a particular subject signed over?

Yep, we punt on this by using CSRs :slightly_smiling_face:

haydentherapper: I’d lean toward it being an implementation detail. To my knowledge there’s no IETF standard for this concept, so we are exposing a detail that’s only present for the public instance (or someone running their own dex instance). 
it’s simple enough to add though, so I’m supportive of that, and if we see no usage of it, we can always clean it up later.

woodruffw:
I've added `IdentityToken.federated_issuer` in [45ab029](https://github.com/sigstore/sigstore-python/pull/635/commits/45ab02948721e958cc659b3a0e70407b0d079415).

woodruffw:
This should be good to go -- CC @jku for final thoughts (the majority of the refactor here is independent, but the new `IdentityToken.federated_issuer` represents our attempt at exposing the state you'd like here)

sigstore/oidc.py
```diff
@@ -57,6 +58,107 @@ class _OpenIDConfiguration(BaseModel):
...
+        This is nominally an **implementation detail** of a Sigstore deployment,
+        but is exposed because some applications may find it useful to filter/exclude
+        identity tokens by their "original" issuer, even if the token has been
+        forwarded through a federated provider (e.g., a Dex instance).
+        """
+        federated_claims = self._unverified_claims.get("federated_claims")
+        if federated_claims is None:
+            return None
+
+        return str(federated_claims.get("connector_id"))
```
jku, line 143: An unlikely occurrence I suppose but if the key isn't found I think this returns the string `"None"`.

jku (commented):
For my use case this solves the easy technical problem (calling jwt.decode() was never a major issue) but does not solve the hard problem (_how do I know which sigstore signing identity is going to be used?_)

That said, I can try to solve that in the application itself.

sigstore/oidc.py
woodruffw: Yeah, that's bad. Will fix!

woodruffw: Refactored and made more sound with eb14330..e03db4f.

sigstore/oidc.py
woodruffw: Opened https://github.com/sigstore/fulcio/issues/1145 for this.

haydentherapper: Will respond there, thanks!

woodruffw:
To recap from https://github.com/sigstore/fulcio/issues/1145: it's fine for the clients here to special-case on the well-known issuer URLs for the Sigstore production and staging instances, meaning that we can improve the API here a bit without risking exposing too many internal service-side details.

@jku Would an `IdentityToken.effective_certificate_subject: str` (naming pending) work well for your use case? That would be `iss` in non-federated cases, and `federated_claims.connector_id` in Dex-federated cases.

woodruffw:
[45774b6](https://github.com/sigstore/sigstore-python/pull/635/commits/45774b64287e4fa4f790c5ef29a9143304b33df6) implements the idea in https://github.com/sigstore/sigstore-python/pull/635#issuecomment-1536484895: `IdentityToken.expected_certificate_subject` now returns the federated issuer, falling back on the `iss` claim if the federated issuer is not present.

This relies on an implementation detail of Dex, but that detail is borne solely by the client codebase (and not the end user).

jku:
> @jku Would an `IdentityToken.effective_certificate_subject: str` (naming pending) work well for your use case? That would be `iss` in non-federated cases, and `federated_claims.connector_id` in Dex-federated cases.

Yes, it looks like that and `identity` are exactly what I wanted. Thanks!

With regards to naming, it would be good if Fulcio, or Sigstore as a whole system, had an agreed name for this concept so that different clients can use the same name.

sigstore/oidc.py
```diff
@@ -57,6 +58,145 @@ class _OpenIDConfiguration(BaseModel):
...
+
+    @property
+    def issuer(self) -> str:
+        """
+        Returns a URL identifying this `IdentityToken`'s issuer.
+        """
+        return self._issuer
+
+    @property
+    def expected_certificate_subject(self) -> str:
```
jku, line 162: This was `effective_certificate_subject` in a comment. I now you mentioned name is WIP, just pointing it out.

woodruffw: > This was `effective_certificate_subject` in a comment. I know you mentioned name is WIP, just pointing it out.

Yeah, this was a shave on my part: "effective" implies something that's already occurred (to me), whereas "expected" implies some behavior that we expect from the Sigstore services (i.e., this is the value we *expect* them to use for the issuer extension).

Happy to rename if you prefer the other naming, however!

woodruffw:
> With regards to naming, it would be good if Fulcio, or Sigstore as a whole system, had an agreed name for this concept so that different clients can use the same name.

Yep, strongly agreed. I wonder if that's something that should go in the client spec, or maybe somewhere else? CC @znewman01

znewman01:
Likely the Fulcio spec—CC @haydentherapper

haydentherapper:
Ack, added a TODO to the spec doc to include this somewhere

jku (approved):
Looks like a really good improvement to me, thanks for the patient tweaking.

Left one log content comment and have two other notes but none of these should IMO block this PR:
* The attribute name does not feel great but your reasoning about it makes sense and I don't have better suggestions. It does feel like `identity` is a lot like `expected_certificate_subject` but only one of them has "expected" in the name...
* IdentityTokens `identity` and `expected_certificate_subject` attributes also operate slightly differently: one is pre-computed and the other is not. I don't think this matters much

woodruffw:
> * The attribute name does not feel great but your reasoning about it makes sense and I don't have better suggestions. It does feel like `identity` is a lot like `expected_certificate_subject` but only one of them has "expected" in the name...

Yeah, fully agreed -- I'm vacillating between `expected` and `effective`, but I think both are confusing/non-ideal in ways that the other doesn't address 🙂

sigstore/sign.py
```diff
@@ -126,17 +126,16 @@ def sign(
 
         logger.debug("Retrieving signed certificate...")
 
-        oidc_identity = Identity(identity_token)
-        logger.debug(f"cert-identity: {oidc_identity.proof}")
-        logger.debug(f"cert-oidc-issuer: {oidc_identity.issuer}")
+        logger.debug(f"cert-identity: {identity.identity}")
+        logger.debug(f"cert-oidc-issuer: {identity.issuer}")
```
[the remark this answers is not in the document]
woodruffw: Tweaked to reduce the number of independent `debug` calls + emit all three properties here, with more reasonable names 🙂

tetsuo-cpp (approved):
The newest changes LGTM!

tetsuo-cpp (approved):

tetsuo-cpp (approved):

PR #637, opened 2023-04-27
workflows: bump sigstore-conformance

woodruffw:
Now with no labelling required!

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -6,7 +6,6 @@ on:
       - main
   workflow_dispatch:
   pull_request_target:
-    types: [labeled]
 
 permissions:  # added using https://github.com/step-security/secure-workflows
   contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/conformance.yml
```diff
@@ -6,7 +6,6 @@ on:
       - main
   workflow_dispatch:
   pull_request_target:
```
woodruffw, line 8: We can get rid of `pull_request_target` entirely, right?

tetsuo-cpp:
I think the labelling fix won't work until it's in `main`. But this should be good to go.

.github/workflows/conformance.yml
woodruffw: (i.e. it should just be `pull_request`)

tetsuo-cpp: I think we can make it just `pull_request`.

tetsuo-cpp: [d735a8d](https://github.com/sigstore/sigstore-python/pull/637/commits/d735a8d2b632c63da610aac3f80b9ffa8e9df6b8)

tetsuo-cpp (approved):
LGTM!

PR #641, opened 2023-05-01
sigstore: fix `detect_credential` signature

woodruffw:
This API accidentally gained a parameter in a point release, which is both a semver breakage and strictly unnecessary (since the parameter is an invariant).


woodruffw:
CC @lukpueh

di (approved):
Good catch.

woodruffw:
> Good catch.

All credit goes to @lukpueh!

Issue #514, opened 2023-03-01
Support signing multiple artifacts with a single certificate


mayaCostantini:
**Description**

For signing a large number of input files, supporting signing multiple artifacts with a single certificate (and eventually asynchronous signing) in `sigstore-python` could allow saving time and API calls to Fulcio and OIDC providers. Instead of creating a new key pair and requesting a different signing certificate from Fulcio when signing every artifact, those two first steps could be shared for a group of artifacts, and the rest of the steps executed sequentially or asynchronously.

This would result in the following flow when multiple artifacts are passed to sigstore to be signed:

- Generate a single key pair
- Record the current timestamp
- Get an identity token from an OIDC provider
- Request a single signing certificate from Fulcio using the id token
- Verify the certificate and sign off the artifacts (asynchronously)
- When signing, verify that the current certificate is not expired by comparing it to the recorded timestamp. If the validity time of the certificate is exceeded, request a new certificate.
- Send (asynchronous) requests to Rekor to create a signature entry for every artifact

This feature could be configurable to allow users to turn it off and request one certificate per artifact if needed (even though I am not able to think of any reason for having a different certificate for each artifact).

If the asynchronous part turns out to be too complex and involves rewriting too many parts of the library, we could simply use the current sequential signing flow for all artifacts, keeping the shared steps (key generation and certificate request) as the only optimizations.

cc @lukehinds @znewman01

woodruffw:
I might be wrong, but I believe we already do this: our `Signer` instance does the OIDC and CSR flows once per instance, and allows multiple artifacts to be signed against that instance.

woodruffw:
Whoops, I did indeed get this slightly wrong: `sigstore sign` only does the OIDC identity token flow once per invocation, but the Signer itself *always* performs key generation + CSR for each individual signing operation. So we do indeed spend quite a bit of time on network trips that could be flattened down to a single key and cert.

mayaCostantini:
@woodruffw I would be interested on working on this issue if no one else is assigned to it.

woodruffw:
> @woodruffw I would be interested on working on this issue if no one else is assigned to it.

Yeah, please do! I'd be open to this being a new kwarg on the `Signer` constructor, which would then perform the separate "single" flow you've described.

mayaCostantini:
Sounds good, thanks for the suggestions :slightly_smiling_face:

mayaCostantini:
[...]

PR #645, opened 2023-05-03
Add option to sign multiple artifacts with the same key and certificate

mayaCostantini:
#### Summary
Resolves https://github.com/sigstore/sigstore-python/issues/514

Add a `--single-cert` option to the`sign` subcommand to sign multiple artifacts by generating one private key and requesting a single certificate from Fulcio.

#### Release Note

* Add a `--single-cert` option to sign multiple artifacts with a single key and certificate

#### Documentation

Added corresponding documentation in `README.md` (`sign` command output).


sigstore/_cli.py
```diff
@@ -329,6 +329,12 @@ def _parser() -> argparse.ArgumentParser:
         default=_boolify_env("SIGSTORE_OVERWRITE"),
         help="Overwrite preexisting signature and certificate outputs, if present",
     )
+    output_options.add_argument(
```
mayaCostantini, line 332: I am unsure if this belongs to `output_options`, I can change that if necessary

mayaCostantini:
Converting to draft for fixing CI

woodruffw:
Thanks @mayaCostantini! I'll have some time to review this later today.

jku:
Based on Williams useful explanation in https://github.com/sigstore/sigstore-python/issues/514#issuecomment-1533246769, should we consider:
1. Making multiple signatures with single CSR (so `single_certificate=True`) the default
2. Even not having the option at all -- this could be added later if someone actually needs it

jku:
> As for 2., we could leave the parameter in the Signer constructor with True as the default, but we could hide it from the CLI to avoid confusion. What do you think?

I was thinking that for the API user the Signer argument is not really necessary as the caller can just create several Signers if they want several CSRs to be used (while still reusing the same identity token). 

However, looking at the API again I'm reminded the token is not a constructor argument to Signer, but an argument to `sign()`... so one Signer can be used to sign with multiple identities -- which means if Signer wants to store the signing certificate for re-signing, it would have to cache  _a certificate per each identity token_.

I wonder if it it would be cleaner to have another signer class instead (like I think William may have suggested at one point), one that would take the token as a constructor argument and would cache the CSR?

sigstore/_cli.py
woodruffw: `output_options` works for me! It could also go under a new `signing_options` or similar, but no strong opinion there.

woodruffw:
@mayaCostantini given that this changes the internal state of the `Signer` quite a bit, what do you think about one of these two designs:

* Breaking this out into a `SingleKeySigner`/`CachingSigner`, per https://github.com/sigstore/sigstore-python/issues/514#issuecomment-1453795673
* Flipping the behavior and making this the default, with `Signer(caching=false)` or similar

I'm okay with either of these (with a slight preference for the second), and IMO either would result in a cleaner diff/easier to maintain internal invariants.

If we go the second route, we'll probably want the instance to be "self-maintaining" -- accessing the signing key and materials should _always_ produce materials that are valid at the current time, so both should be properties that transparently refresh as needed under the hood (except for in the `caching=False` case, where we should _always_ produce a new key and certificate).

woodruffw:
> However, looking at the API again I'm reminded the token is not a constructor argument to Signer, but an argument to `sign()`... so one Signer can be used to sign with multiple identities -- which means if Signer wants to store the signing certificate for re-signing, it would have to cache _a certificate per each identity token_ and only reuse the cert if the tokens match.

Oh, this is a good point -- I forgot that I had designed it this way...

Yeah, a more general refactor is possibly in order here: `sign(...)` should(?) conceptually only take the input being signed over, with the `Signer` maintaining all other state. 

There's some weird impedance mismatch there, though -- constructing a `Signer` with an internal identity token means that the token will expire at some point; at that point the `Signer` instance will enter an invalid state, *even if* we refresh the CSR and short-lived key. That might be pretty confusing for end users, especially since different IdPs might have different expiration policies and timespans.

jku:
> Yeah, a more general refactor is possibly in order here: sign(...) should(?) conceptually only take the input being signed over, with the Signer maintaining all other state.

I think I agree with this -- it makes sense for the signer to be tied to a single identity -- but also with your other observations in this comment...  Maybe the easy/correct solution is a new signer-class  that implements this slightly different concept? Naming is hard, so I have not come up with a good name for this SingleCertificateSigner.  If it becomes clear that the new class provides everything users need, then the old one can be deprecated and removed later.

woodruffw:
> If it becomes clear that the new class provides everything users need, then the old one can be deprecated and removed later.

These changes will land with 2.0, so I'm personally okay with a decent bit of breakage here! So long as we can come up with a suitable API 🙂 

Just to write the constraints down in a single place:

1. The `Signer` instance itself _should_ (ideally) never become "invalid", at least not over reasonable object lifetimes (minutes-to-hours, not weeks-to-years);
2. A `Signer` _should_ be bound to a single identity: multiple calls to `sign(...)` with different identity tokens are confusing (if not currently dangerous, since the underlying key material is always refreshed);
3. A `Signer` _should_, by default, reuse its ephemeral key material (and certificate) for as many signing operations as possible.

One idea:

1. `Signer` is renamed to `SigningContext`
2. `SigningContext.sign` is renamed to `SigningContext.with_signer(identity_token=...)`, which is a *context manager* for a new `Signer` class
3. That *new* `Signer` class handles the invariants above: it can internally cache keys, and ensures that the identity token is bound in rather than passed in.

As an example:

```python
# access the production Sigstore instances
production = SigningContext.production()

with production.with_signer(identity_token=...) as signer:
    result = signer.sign(...)
```

Thoughts? I'm still not 100% happy with that (maybe it shouldn't be a context manager?), but it's food for thought 🙂

mayaCostantini:
Thanks for your thoughts on this. I personally like the idea of a context manager; I find it good to make API users understand that resources associated to signing (key and certificate) are ephemeral and should remain within a certain scope to avoid mixing up identities.

Does something like the following implement what you were thinking of @woodruffw ?

```
from contextlib import contextmanager

# New `Signer` class
class Signer:
    def __init__(self, identity_token: str, cache: bool=True):
        self._identity_token: str = identity_token
        self._signing_certificate: Optional[FulcioCertificateSigningResponse] = None
        self._private_key: Optional[ec.EllipticCurvePrivateKey] = None
        self.cache : bool = cache
        
    @property
    def private_key(self) -> ec.EllipticCurvePrivateKey:
        # Get or generate a private key
        ...
        
     @property
    def signing_cert(
        self,
    ) -> Callable[[str, ec.EllipticCurvePrivateKey], FulcioCertificateSigningResponse]:
        # Get or request a signing certificate for Fulcio with the provided identity token
        ...
    
    def sign(self, input_: IO[bytes]) -> SigningResult:
        # Sign the artifact
        ...


class SigningContext:
    def __init__(self, *, fulcio: FulcioClient, rekor: RekorClient):
        ...
    def production(cls) -> SigningContext:
        ...
    def staging(cls) -> SigningContext:
        ...
    @contextmanager
    def with_signer(self, identity_token: str):
        signer = Signer(identity_token: str, cache=True)
        try:
            yield signer
        finally:
            del signer
```

woodruffw:
> Does something like the following implement what you were thinking of @woodruffw ?

Yep, looks close to what I was thinking!

We should think a bit more about the lifetimes here (maybe it makes sense for the new `Signer` to explicitly raise when the identity token is expired?), but I think this is a good starting point 🙂

woodruffw:
> Sounds good! Should I proceed with the implementation?

Works for me! I assume you'll need an `IdentityToken.expired` or similar API, but we can address that when it comes up 🙂

mayaCostantini:
@woodruffw @jku feel free to take a look at https://github.com/sigstore/sigstore-python/pull/645/commits/ff30b0955ce6bc665684ec010e97d08e3dafe14d 
If this looks good to you, I can commit the updated tests and docs as well.

sigstore/sign.py
```diff
@@ -82,74 +89,106 @@ class Signer:
...
+
+        return self._private_key
+
+    @property
+    def signing_cert(
+        self,
+    ) -> Callable[
+        [str, ec.EllipticCurvePrivateKey, FulcioClient],
+        FulcioCertificateSigningResponse,
+    ]:
```
woodruffw, line 123: I had to re-read this a few times to figure out what was happening -- it's a `@property` that returns a nested function, so subsequent callsites look like instance methods when they really aren't 🙂 

That's a little confusing IMO: this should probably either be a normal instance method (no `@property`) or a `@property` that returns `FulcioCertificateSigningResponse | None`, with `None` indicating the exceptional case (`cache and expired`). 

Thoughts?

sigstore/sign.py
```diff
@@ -82,74 +89,106 @@ class Signer:
...
+        self._private_key: Optional[ec.EllipticCurvePrivateKey] = None
+
+    @property
+    def private_key(self) -> ec.EllipticCurvePrivateKey:
+        """Get or generate a signing key."""
+        if not self.cache or self._private_key is None:
+            logger.debug("Generating ephemeral keys...")
+            self._private_key = ec.generate_private_key(ec.SECP384R1())
+
+        return self._private_key
```
woodruffw, line 115: Small thing, but let's consider making the public API surface as small as possible to start with: `Signer` instances should only expose the `sign(...)` operation to begin with, not the helper properties that enable that operation.

Given that, this should probably be `def _private_key(self)` and the corresponding attribute should be `self.__private_key` (and similarly for `signing_cert`).

sigstore/sign.py
```diff
@@ -192,6 +231,75 @@ def sign(
...
+        rekor = RekorClient.staging(updater)
+        return cls(
+            fulcio=FulcioClient.staging(),
+            rekor=rekor,
+        )
+
+    @contextmanager
+    def with_signer(
+        self, identity_token: str, cache: bool = True
+    ) -> Generator[Signer, None, None]:
```
woodruffw, line 284: Nitpick: `mypy` lets us get away with this 🙂 

```suggestion
    ) -> Iterator[Signer]:
```

sigstore/sign.py
```diff
@@ -192,6 +231,75 @@ def sign(
...
+        `cache` determines whether the signing certificate and ephemeral private key
+        generated by the `Signer` instance should be reused (until the certificate expires)
+        to sign different artifacts.
+        Default is `True`.
+        """
+        signer = Signer(identity_token, cache)
+        try:
+            yield signer
+        finally:
+            del signer
```
woodruffw, line 300: I've got mixed feelings on `del` in Python 3: it can lead people into a false sense of security/"finality" that isn't actually guaranteed (e.g. anything taking a reference on any `Signer` instance state will prevent a full object deletion). 

OTOH, there's nothing unsafe about it either -- if we _do_ keep this, we might as well go 100% and also define `Signer.__del__` to include explicit destruction of the secret key and other materials. 

CC @jku for opinions.

woodruffw: Or another idea: rather than `del`, we could create a "poison" style API, e.g.:

```
signer._poison()

# any attempts to sign or use the signer object after this point raise an `InvalidObjectError` or similar
```

This has the benefit of correctly "enforcing" the lifetime of the `signer` binding in a way that context managers can't, at the cost of a decent bit of complexity (which I'm not sure it's worth).

sigstore/sign.py
```diff
@@ -82,74 +89,106 @@ class Signer:
...
-        `rekor` is a `RekorClient` capable of connecting to a Rekor instance
-        and creating transparency log entries.
+        `cache` determines whether the signing certificate and ephemeral private key
+        should be reused (until the certificate expires) to sign different artifacts.
+        Default is `True`.
         """
-        self._fulcio = fulcio
-        self._rekor = rekor
+        self._identity_token: str = identity_token
+        self.cache: bool = cache
```
woodruffw, line 104: One thought: rather than retaining `cache` as an attribute, we could do:

```python
self.__cached_private_key = None
if cache:
    self._cached_private_key = keygen(...)
```

This eliminates the dual state checks below (we can just check `_cached_private_key`).

sigstore/sign.py
mayaCostantini: +1, a simple method should be enough here

sigstore/sign.py
mayaCostantini: Nice! Will replace it

sigstore/sign.py
mayaCostantini: The first approach with defining a `__del__` method to get rid of the signer's attributes seems like a simpler option to me. Even if this does not happen with `del`, it makes sense that the attributes get deleted when leaving the signing context scope.

mayaCostantini:
Thanks for the review, I tried to address your comments in https://github.com/sigstore/sigstore-python/pull/645/commits/9788a150d501abfb5b750535396404346f6448cf

However I am still not sure if the approach is right regarding removing the `cache` attribute, please let me know what you think about it.

sigstore/sign.py
```diff
@@ -89,86 +89,105 @@ class Signer:
...
         `cache` determines whether the signing certificate and ephemeral private key
         should be reused (until the certificate expires) to sign different artifacts.
         Default is `True`.
         """
         self._identity_token: str = identity_token
-        self.cache: bool = cache
-        self._signing_certificate: Optional[FulcioCertificateSigningResponse] = None
-        self._private_key: Optional[ec.EllipticCurvePrivateKey] = None
+        self.__cached_private_key: Optional[ec.EllipticCurvePrivateKey] = None
+        self.__private_key: Optional[ec.EllipticCurvePrivateKey] = None
```
woodruffw, line 110: I don't think we need this, right? If the user hasn't explicitly requested caching then we should never store the ephemeral key at all, and always return a new one.

sigstore/sign.py
```diff
@@ -89,86 +89,105 @@ class Signer:
...
 
     @property
-    def private_key(self) -> ec.EllipticCurvePrivateKey:
+    def _private_key(self) -> ec.EllipticCurvePrivateKey:
         """Get or generate a signing key."""
-        if not self.cache or self._private_key is None:
+        if self.__cached_private_key is None:
             logger.debug("Generating ephemeral keys...")
-            self._private_key = ec.generate_private_key(ec.SECP384R1())
+            self.__private_key = ec.generate_private_key(ec.SECP384R1())
```
woodruffw, line 137: Per above: rather than storing into `__private_key`, I think we can return directly here.

sigstore/sign.py
```diff
@@ -293,7 +308,7 @@ def with_signer(
         to sign different artifacts.
         Default is `True`.
         """
-        signer = Signer(identity_token, cache)
+        signer = Signer(identity_token, self._fulcio, cache)
```
woodruffw, line 311: We can make this a little more resilient to API changes by pushing the whole `SigningContext` in, i.e.:

```suggestion
        signer = Signer(identity_token, self, cache)
```

(and then using `SigningContext._fulcio` wherever we need it.)

sigstore/sign.py
```diff
@@ -89,86 +89,105 @@ class Signer:
...
-    @property
-    def signing_cert(
+    def _signing_cert(
         self,
-    ) -> Callable[
-        [str, ec.EllipticCurvePrivateKey, FulcioClient],
-        FulcioCertificateSigningResponse,
-    ]:
+        private_key: ec.EllipticCurvePrivateKey,
+        fulcio: FulcioClient,
```
woodruffw, line 146: Rather than passing in here, maybe this makes sense as an instance attribute on the `Signer`, e.g. access through `self._signing_context._fulcio`?

sigstore/sign.py
```diff
@@ -30,7 +30,7 @@
 
 with artifact.open("rb") as a:
     signing_ctx = SigningContext.production()
-    with signing_ctx.with_signer(token, cache=True) as signer:
+    with signing_ctx.with_signer(token, fulcio=singing_ctx._fulcio, cache=True) as signer:
```
woodruffw, line 33: We probably don't want to make the user pass the Fulcio instance back in like this 🙂 -- I left a suggestion below for making the `SigningContext` into an instance attribute on the `Signer`.

```suggestion
    with signing_ctx.with_signer(token, fulcio=singing_ctx._fulcio, cache=True) as signer:
```

sigstore/sign.py
```diff
@@ -89,86 +89,105 @@ class Signer:
...
         self._identity_token: str = identity_token
-        self.cache: bool = cache
-        self._signing_certificate: Optional[FulcioCertificateSigningResponse] = None
-        self._private_key: Optional[ec.EllipticCurvePrivateKey] = None
+        self.__cached_private_key: Optional[ec.EllipticCurvePrivateKey] = None
+        self.__private_key: Optional[ec.EllipticCurvePrivateKey] = None
+        self.__cached_signing_certificate: Optional[
+            FulcioCertificateSigningResponse
+        ] = None
+        self.__signing_certificate: Optional[FulcioCertificateSigningResponse] = None
```
woodruffw, line 114: Same thought about this: I don't think we need this attribute at all, since in the non-caching case we'll always regenerate it.

woodruffw:
Thank you! And thanks for bearing through my reviews 🙂; I'll do another round today but this is looking really good to me!

sigstore/sign.py
jku: This _might_ be a little unexpected for contextmanager users: I think usually the context object ( signer in this case) still exists after the with-clause, it's just closed or inactive. In this case I think the whole object just disappears  ... 

I think it's fine though. Maybe worth a docstring mention?

jku (dismissed):
I've not tried this myself but it looks very good.

The only nitpick I might have is that the method name is a bit repetitive:
`with ctx.with_signer() as signer:`
that could be just 
`with ctx.signer() as signer:`
but I can also see the argument that the chosen name more easily identifies the method  as a context manager... so no strong opinions here.

woodruffw:
> the method name is a bit repetitive

Agreed -- `with_signer` was just the first thing I thought of, I think `with ctx.signer()` is equally descriptive given that the `with` syntax itself introduces the context manager!

sigstore/_internal/oidc/__init__.py
```diff
@@ -69,3 +76,12 @@ def __init__(self, identity_token: str) -> None:
...
                 raise IdentityError("Identity token missing `sub` claim")
+
+    def is_expired(self) -> bool:
+        """Verify if the identity token for this `Identity` is expired."""
+        if self.exp_timestamp:
+            now: float = datetime.now(timezone.utc).timestamp()
+            token_timestamp: float = self.exp_timestamp
+            return now > token_timestamp
+        else:
+            raise ValueError("Identity token does not have an expiration timestamp")
```
woodruffw, line 87: Let's push this error state up to the constructor, and raise `IdentityError` (since it seems like a valid `Identity` _needs_ this claim.)

sigstore/_internal/oidc/__init__.py
```diff
@@ -40,6 +46,7 @@ def __init__(self, identity_token: str) -> None:
         """
         identity_jwt = jwt.decode(identity_token, options={"verify_signature": False})
 
+        self.exp_timestamp = identity_jwt.get("exp")
```
woodruffw, line 49: We also need to retrieve the `nbf` claim here and check it below 🙂 

(Maybe we can push all of this into the `jwt.decode(...)` call instead? It should allow us to verify claims even through we're not verifying the signature.)

sigstore/sign.py
```diff
@@ -82,74 +89,111 @@ class Signer:
...
-        rekor = RekorClient.staging(updater)
-        return cls(fulcio=FulcioClient.staging(), rekor=rekor)
+        del self.__cached_signing_certificate
+        del self.__cached_private_key
+
+    @property
+    def _private_key(self) -> ec.EllipticCurvePrivateKey:
+        """Get or generate a signing key."""
+        if self.__cached_private_key is None:
+            logger.debug("Generating ephemeral keys...")
```
woodruffw, line 133: Extreme nitpick: 

```suggestion
            logger.debug("no cached key; generating ephemeral key")
```

sigstore/sign.py
woodruffw: Yeah, that's a good point -- I did a little more reading, and it seems like defining `__del__` isn't considered idiomatic in Python 3 anymore (both because of end user expectations, and because it's misleading -- it doesn't actually guarantee that the secret materials here have been freed).

Given that, I think we can drop `del` and `__del__` entirely and emphasize in documentation that the `Signer` object is only valid in its created context.

woodruffw:
I think this is pretty much ready to go -- I pushed a few small cleanups, including bringing the tests back up. My plan is to get #635 merged first since it makes the `Identity` (now `IdentityToken`) API public; we can then deconflict here.

mayaCostantini:
Thanks for the changes @woodruffw! Let me know if anything more is needed on my side

jku:
check-readme error on CI: README has an extra line compared to output if I'm reading that right
```
# sigstore --help
# sigstore sign --help
make: *** [Makefile:123: check-readme] Error 1
48d47
< 
##[error]Process completed with exit code 2.
```

woodruffw:
>  Let me know if anything more is needed on my side

Nope, nothing more needed! I caused a big mess with the other merge, so I'll deconflict here and try to have something in a mergeable state today.

Thanks a ton for going through the review cycles -- I'm very excited about landing this!

> check-readme error on CI: README has an extra line (line 181?) compared to output if I'm reading that right

Yeah, I'll look into that. Probably because the CLI flags have been updated.

woodruffw (dismissed):
LGTM, great work @mayaCostantini!

I'd like @di to also give this a review before merge.

woodruffw (approved):

mayaCostantini:
Thanks for the thorough reviews and help @woodruffw! I am excited as well to see this in v2.0 :slightly_smiling_face:

jku (dismissed):
This looks good to me.

As a nit-pick, if we don't have a good use case for the CLI flag, maybe it just does not need to exist?
* API users can always handle their signers like they want
* CLI users can invoke sign multiple times if they really need individual certs (admittedly this approach currently has other downsides like repeated TUF initializations)

woodruffw:
> As a nit-pick, if we don't have a good use case for the CLI flag, maybe it just does not need to exist?

Agreed, removing.

woodruffw (approved):

sigstore/oidc.py
```diff
@@ -77,22 +92,28 @@ def __init__(self, raw_token: str) -> None:
...
+                    "require": ["aud", "iat", "exp", "iss"],
+                },
+                audience=DEFAULT_AUDIENCE,
             )
-        except jwt.InvalidTokenError as exc:
-            raise IdentityError("invalid identity token") from exc
+        except Exception as exc:
+            raise IdentityError(
+                "Identity token is malformed or missing claims"
+            ) from exc
```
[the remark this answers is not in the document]
woodruffw: The actual traceback _should_ still contain that specificity, because of the `from exc` -- the traceback chain will contain the underlying `pwjwt` exception, which in turn (should) mention the missing or invalid claim.

(I can make that more explicit by embedding `{exc}` rather than chaining with it, if you'd like.)

sigstore/oidc.py
```diff
@@ -77,22 +92,28 @@ def __init__(self, raw_token: str) -> None:
...
+                    "verify_aud": True,
+                    "verify_iat": True,
+                    "verify_exp": True,
+                    "require": ["aud", "iat", "exp", "iss"],
+                },
+                audience=DEFAULT_AUDIENCE,
             )
-        except jwt.InvalidTokenError as exc:
-            raise IdentityError("invalid identity token") from exc
+        except Exception as exc:
```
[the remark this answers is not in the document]
woodruffw: This is unfortunately similar to https://github.com/pypi/warehouse/blob/beaa920c0c805c35066dab7239be9daf84ebb1ba/warehouse/oidc/services.py#L242-L276 -- we can't enforce pyjwt's own exception hygiene, so I figured it'd be best to assume it can raise anything rather than play whac-a-mole over time.

That being said, we can catch the more specific known exceptions and leave an `Exception` fallback, if you'd like 🙂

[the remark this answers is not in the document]
di: Ah, right, forgot about that. This should be fine then.

sigstore/oidc.py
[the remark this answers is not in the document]
di: I don't love depending on the exception from `pyjwt`, but let's see what happens.

di (approved):

woodruffw:
Thanks a ton @mayaCostantini! Excited to land this.

mayaCostantini:
Thanks a lot to you @woodruffw !

PR #662, opened 2023-05-31
sign: switch to P-256

woodruffw:
This is faster than P-384, is well-supported, and is well within security margins.

Longer term I'd like to switch to ed25519, but I'm having some trouble getting Rekor to accept `hashedrekord` entries for ed25519 signatures.

cc @alex

### DIFF
--- a/sigstore/sign.py
+++ b/sigstore/sign.py
@@ -122,7 +122,7 @@ class Signer:
         """Public API for signing blobs"""
         input_digest = sha256_streaming(input_)
 
-        private_key = ec.generate_private_key(ec.SECP384R1())
+        private_key = ec.generate_private_key(ec.SECP256R1())
 
         logger.debug(
             f"Performing CSR: identity={identity.identity} "


haydentherapper (approved):
Rekor doesn’t support ed25519 for hashedrekord because it requires the preimage, not the hash. See https://github.com/sigstore/rekor/issues/1325 for a fix, haven’t had a chance to double check that’ll work.

woodruffw:
> Rekor doesn’t support ed25519 for hashedrekord because it requires the preimage, not the hash. See [sigstore/rekor#1325](https://github.com/sigstore/rekor/issues/1325) for a fix, haven’t had a chance to double check that’ll work.

Cool, thanks! I think the Python client will be blocked on ed25519-ph anyways due to Cryptography not supporting it (in part because OpenSSL in turn doesn't).

di (dismissed):

di (approved):

Issue #665, opened 2023-06-02
Consolidate the sign API


laurentsimon:
**Description**

The existing `sign()` takes a limited number of inputs, and it may be useful to have an `options` input to add more flexible options. Example:
- `SIGSTORE_OAUTH_FORCE_OOB` is currently an env variable which needs to be set prior to calling the API. Should it not be an input to the signing function instead?
- Signing currently support `hashedrekord`, but this may change in the future to support intoto types

I wonder whether it'd make sense to consolidate the API.

laurentsimon:
> > * `SIGSTORE_OAUTH_FORCE_OOB` is currently an env variable which needs to be set prior to calling the API. Should it not be an input to the signing function instead?
> 
> I might be misunderstanding what you mean, but I don't think this is true -- the current `Signer.sign()` API takes an identity token that's been pre-retrieved, so that environment variable shouldn't affect it. 

The example shows using `Issuer.get_identity_token()` in https://github.com/sigstore/sigstore-python/blob/main/sigstore/sign.py#L26, which uses the env variable https://github.com/sigstore/sigstore-python/blob/main/sigstore/oidc.py#L264. If the token is passed as a string by a user, you're right it works. But if doing the OAuth flow, how should the caller create the token?

woodruffw:
Okay, I understand what you mean now -- you're talking about the `identity_token()` API, not the signing API. I don't think of the former as a part of the signing API, but it's reasonable to consider it part of it.

I agree that should become a parameter on `identity_token()`, rather than an environment variable. I'd be happy to accept a PR for it, otherwise I'll do it when I have a free moment 🙂

laurentsimon:
I sent https://github.com/sigstore/sigstore-python/pull/667

woodruffw:
Done with #667 I believe. Please ping if I got this wrong!

PR #667, opened 2023-06-02
feat: Add `--oauth-force-oob` CLI option

laurentsimon:
Adresses comment https://github.com/sigstore/sigstore-python/issues/665#issuecomment-1574033667

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -972,7 +972,7 @@ def _get_identity(args: argparse.Namespace) -> Optional[IdentityToken]:
     token = issuer.identity_token(
         client_id=args.oidc_client_id,
         client_secret=args.oidc_client_secret,
-        force_oob=args.oath_force_oob,
+        force_oob=args.oauth_force_oob,
     )
 
     return token
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -222,10 +222,10 @@ def _add_shared_oidc_options(
         help="The OpenID Connect issuer to use (conflicts with --staging)",
     )
     group.add_argument(
-        "--oidc-disable-default-browser",
+        "--oauth-force-oob",
         action="store_true",
-        default=_boolify_env("SIGSTORE_OIDC_DISABLE_DEFAULT_BROWSER"),
-        help="Do not start the default web browser to complete the OAuth flow. Instead display a web link to the user.",
+        default=_boolify_env("SIGSTORE_OAUTH_FORCE_OOB"),
+        help="Force an out-of-band OAuth flow and do not automatically start the default web browser",
     )
 
 
@@ -972,7 +972,7 @@ def _get_identity(args: argparse.Namespace) -> Optional[IdentityToken]:
     token = issuer.identity_token(
         client_id=args.oidc_client_id,
         client_secret=args.oidc_client_secret,
-        force_oob=args.oidc_disable_default_browser,
+        force_oob=args.oath_force_oob,
     )
 
     return token
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -221,6 +221,12 @@ def _add_shared_oidc_options(
         default=os.getenv("SIGSTORE_OIDC_ISSUER", DEFAULT_OAUTH_ISSUER_URL),
         help="The OpenID Connect issuer to use (conflicts with --staging)",
     )
+    group.add_argument(
+        "--oidc-disable-default-browser",
+        action="store_true",
+        default=_boolify_env("SIGSTORE_OIDC_DISABLE_DEFAULT_BROWSER"),
+        help="Do not start the default web browser to complete the OAuth flow. Instead display a web link to the user.",
+    )
 
 
 def _parser() -> argparse.ArgumentParser:
@@ -964,7 +970,9 @@ def _get_identity(args: argparse.Namespace) -> Optional[IdentityToken]:
         args.oidc_client_secret = ""  # nosec: B105
 
     token = issuer.identity_token(
-        client_id=args.oidc_client_id, client_secret=args.oidc_client_secret
+        client_id=args.oidc_client_id,
+        client_secret=args.oidc_client_secret,
+        force_oob=args.oidc_disable_default_browser,
     )
 
     return token
--- a/sigstore/_utils.py
+++ b/sigstore/_utils.py
@@ -178,7 +178,7 @@ def read_embedded(name: str, prefix: str) -> bytes:
     Read a resource embedded in this distribution of sigstore-python,
     returning its contents as bytes.
     """
-    return resources.files("sigstore._store").joinpath(prefix, name).read_bytes()  # type: ignore
+    return resources.files("sigstore._store").joinpath(prefix, name).read_bytes()
 
 
 def cert_is_ca(cert: Certificate) -> bool:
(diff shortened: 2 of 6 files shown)


sigstore/_utils.py
```diff
@@ -178,7 +178,7 @@ def read_embedded(name: str, prefix: str) -> bytes:
     Read a resource embedded in this distribution of sigstore-python,
     returning its contents as bytes.
     """
-    return resources.files("sigstore._store").joinpath(prefix, name).read_bytes()  # type: ignore
+    return resources.files("sigstore._store").joinpath(prefix, name).read_bytes()
```
laurentsimon, line 181: this was failing the `make lint`, so I removed it

woodruffw: > this was failing the make lint, so I removed it

Sorry, that's my fault -- the CI needs it, because we lint in CI against our minimum supported Python. So please add that back in 🙂

sigstore/_cli.py
```diff
@@ -221,6 +221,12 @@ def _add_shared_oidc_options(
         default=os.getenv("SIGSTORE_OIDC_ISSUER", DEFAULT_OAUTH_ISSUER_URL),
         help="The OpenID Connect issuer to use (conflicts with --staging)",
     )
+    group.add_argument(
+        "--oidc-disable-default-browser",
```
woodruffw, line 225: I think this flag name might be a little ambiguous: what it's doing is forcing the OOB flow, which is conceptually untied to the "default" browser. It's also doing an OAuth2 flow, so I think we should emphasize that rather than OIDC (OIDC refers to the "shape" of the claims that come out of the flow).

What do you think about `--oauth-no-browser` or `--oauth-force-oob`?

woodruffw: (I have a slight preference for the latter, since it'll be consistent with the environment suggestion below. But either is fine IMO; your call.)

sigstore/oidc.py
```diff
@@ -247,7 +246,10 @@ def staging(cls) -> Issuer:
...
 
     def identity_token(  # nosec: B107
-        self, client_id: str = "sigstore", client_secret: str = ""
+        self,
+        client_id: str = "sigstore",
+        client_secret: str = "",
+        force_oob: bool = False,
     ) -> IdentityToken:
         """
         Retrieves and returns an `IdentityToken` from the current `Issuer`, via OAuth.
```
woodruffw, line 255: Let's update the docstring here as well, to emphasize that `force_oob` modifies the browser-opening behavior.

woodruffw (commented):
Thanks for adding this! A couple of nitpicks 🙂

woodruffw:
JFYI: This will also require a `CHANGELOG.md` entry and updates to the `--help` outputs in the `README.md`. But we can do that as the last step.

sigstore/_cli.py
laurentsimon: I renamed `--oauth-force-oob`, even though I think prepending `--oidc-xxx` wold be more consistent with other commands if it's under the oidc group. Lmk

sigstore/oidc.py
laurentsimon: PTAL

sigstore/_cli.py
woodruffw: Yeah, it's a little inconsistent, but IMO it's worth it for the compatibility/symmetry with the existing flag. 

There might be a better section to put this under; I'll take a look.

sigstore/oidc.py
woodruffw: Looks good, thank you!

woodruffw:
LGTM behaviorally!

`check-readme` is unhappy, but I'll fix that:

```
# sigstore --help
# sigstore sign --help
4,5c4,5
<                      [--no-default-files] [--signature FILE]
make: *** [Makefile:[12](https://github.com/sigstore/sigstore-python/actions/runs/5158329157/jobs/9292054589?pr=667#step:5:13)3: check-readme] Error 1
<                      [--certificate FILE] [--bundle FILE]
---
>                      [--oauth-force-oob] [--no-default-files]
>                      [--signature FILE] [--certificate FILE] [--bundle FILE]
30c30,32
<   --oauth-force-oob     Force an out-of-band OAuth flow and do not automatically start the default web browser (default: False)
---
>   --oauth-force-oob     Force an out-of-band OAuth flow and do not
>                         automatically start the default web browser (default:
>                         False)
```

woodruffw (approved):

woodruffw (approved):
LGTM, thanks @laurentsimon! 

I need one of the other maintainers to approve as well, but this will be included in the 2.0 series.

di (approved):

di (dismissed):

tetsuo-cpp (dismissed):

woodruffw:
I have no clue why the macOS and Windows CIs are getting stuck here. Maybe some kind of bad runner state? Nothing in this PR itself should vary between platforms.

woodruffw:
Weirdly, both seem to hang at around the same test:

Windows:

```
2023-06-08T20:15:13.5528997Z test/unit/internal/oidc/test_issuer.py::test_fail_init_url PASSED        [ 75%]
2023-06-08T20:15:13.6085787Z test/unit/internal/oidc/test_issuer.py::test_init_url PASSED             [ 75%]
2023-06-08T20:22:23.5095829Z ##[error]The operation was canceled.
2023-06-08T20:22:23.6691779Z Post job cleanup.
2023-06-08T20:22:24.0268157Z [command]"C:\Program Files\Git\bin\git.exe" version
```

macOS:

```
2023-06-08T20:15:08.1732720Z test/unit/internal/oidc/test_issuer.py::test_fail_init_url PASSED        [ 75%]
2023-06-08T20:15:08.1998280Z test/unit/internal/oidc/test_issuer.py::test_init_url PASSED             [ 75%]
2023-06-08T20:22:31.5320710Z ##[error]The operation was canceled.
2023-06-08T20:22:31.5554660Z Post job cleanup.
2023-06-08T20:22:32.1636730Z [command]/usr/local/bin/git version
2023-06-08T20:22:32.2323580Z git version 2.40.1
```

I'll do some more digging later.

di:
I wonder if these are waiting for user input... perhaps we need to add a reasonable timeout around any flows that wait for user input?

woodruffw:
> I wonder if these are waiting for user input... perhaps we need to add a reasonable timeout around any flows that wait for user input?

Yeah, that was my guess, although it's a mystery to me why they'd block on these platforms and not on the Linux CI. 

I think a timeout around user input makes sense, in places we can enforce it (e.g. `stdin` and not the browser) -- I'll look into that as well.

woodruffw:
This was indeed a silly blocking thing; should be fixed with d6052c9.

di:
How was this passing on linux? E.g. https://github.com/sigstore/sigstore-python/actions/runs/5204665487/jobs/9412771815

woodruffw:
I think it probably passed because we're not being specific enough about _which_ `IdentityError` is being caught in the test.

Looking at it more closely, I'm not 100% sure what branch this test was originally intended to cover -- @jleightcap do you remember what your intent with this was?

woodruffw:
It looks like it was probably meant to catch this:

https://github.com/sigstore/sigstore-python/blob/d61025510f6497bf82fce89cd2aec27383c64cfe/sigstore/oidc.py#L361-L364

I'll add a proper exception message there and assert on it. My guess is that the Linux CI was passing because `webbrowser.open()` failed on it and pushed it into the OOB flow, even without `force_oob`.

woodruffw:
Tests are passing with the new match, so that confirms that the Linux CI was coincidentally entering the OOB flow (whereas Windows and macOS weren't, probably because both have a browser installed).

di (approved):

PR #669, opened 2023-06-06
workflows: debug staging-tests

woodruffw:
WIP.

See #668.

### DIFF
--- a/.github/workflows/staging-tests.yml
+++ b/.github/workflows/staging-tests.yml
@@ -11,6 +11,7 @@ on:
   push:
     branches:
       - main
+      - ww/debug-staging
   schedule:
     - cron: '0 */8 * * *'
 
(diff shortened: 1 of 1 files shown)


woodruffw:
Well, that's interesting:

```
  File "/home/runner/work/sigstore-python/sigstore-python/staging-env/lib/python3.11/site-packages/jwt/api_jwt.py", line 162, in decode_complete
    self._validate_claims(
  File "/home/runner/work/sigstore-python/sigstore-python/staging-env/lib/python3.11/site-packages/jwt/api_jwt.py", line 242, in _validate_claims
    self._validate_iat(payload, now, leeway)
  File "/home/runner/work/sigstore-python/sigstore-python/staging-env/lib/python3.11/site-packages/jwt/api_jwt.py", line 276, in _validate_iat
    raise ImmatureSignatureError("The token is not yet valid (iat)")
jwt.exceptions.ImmatureSignatureError: The token is not yet valid (iat)
```

This is also failing in the ordinary CI but wasn't failing when this was merged, so it's possible that GitHub's own OIDC IdP is regressing here.

woodruffw:
I've added a 5 second leeway window on the validity period claims. This shouldn't be strictly necessary but the observed behavior points to a consistent clock skew between GitHub's own IdP and the Actions runners environment.

Adding this leeway doesn't change the security boundary: the token still isn't verified in any sense, since Fulcio is responsible for verifying it.

di (approved):

Issue #658, opened 2023-05-25
Fail to sign artifacts when `ExtendedKeyUsage` claim is absent from issuer certificate


mayaCostantini:
**Description**

Signing an artifact against a private instance of Fulcio deployed with a password-protected private key backend raises an `ExtensionNotFound` error. After investigation, it seems like when verifying the SCT, the [`_is_preissuer`](https://github.com/sigstore/sigstore-python/blob/b65f64140a334e2de468259dab8f058dc9fa0c3a/sigstore/_internal/sct.py#L124) internal method fails if it cannot find the `ExtendedKeyUsage` extension in the issuer certificate.

However I did not observe this behavior when signing with Cosign against the same instances.
The issue is also absent when signing with `sigstore-python` against another private instance of Sigstore deployed with a Google CA backend, which seems to provide the correct extensions.

**Version**

`sigstore-python v1.1.1`

woodruffw:
Hmm, could you say more about the expected behavior here? From this:

> Signing an artifact against a private instance of Fulcio deployed with a password-protected private key backend raises an `ExtensionNotFound` error. After investigation, it seems like when verifying the SCT, the [`_is_preissuer`](https://github.com/sigstore/sigstore-python/blob/b65f64140a334e2de468259dab8f058dc9fa0c3a/sigstore/_internal/sct.py#L124) internal method fails if it cannot find the `ExtendedKeyUsage` extension in the issuer certificate.

-- it sounds like your backing CA isn't using a Precertificate Signing Certificate per RFC 6962. 

My read of the RFC is that the CA is required to do one of two things:

* Either use a Precertificate Signing Certificate (which **MUST** have an EKU extension);
* Or sign the TBSCert with the CA certificate that will sign the final certificate.

We haven't seen the latter in practice with Sigstore, but maybe that's what's happening here 🙂

ccordoui:
Yes indeed, we are hitting the second case with the TBS that need to be signed by the same Authority than the final one.

As such the CA doesn't need to have the CT EKU, in fact it does not need to have an EKU at all.
That being said, [the first line](https://github.com/sigstore/sigstore-python/blob/v1.1.1/sigstore/_internal/sct.py#L125) of `_is_preissuer`: 
```python
ext_key_usage = issuer.extensions.get_extension_for_class(ExtendedKeyUsage)
```
[...]

PR #674, opened 2023-06-08
Handle the case of missing EKU in _is_preissuer

ccordoui:
#### Summary
Previously, TBSCertificate signed with the same CA than the final certificate was failing if no `ExtendedKeyUsage` were present

#### Release Note
Fix a case where signing an artifact against a private sigstore instance was failing if no `ExtendedKeyUsage` was present in the Sigining CA

#### Documentation
NONE

Resolves: sigstore/sigstore-python#658

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,7 +55,7 @@ All versions prior to 0.9.0 are untracked.
   reducing the number of cryptographic operations and network roundtrips
   required when signing more than one input
   ([#645](https://github.com/sigstore/sigstore-python/pull/645))
-  
+
 * `sigstore sign` now uses an ephemeral P-256 keypair, rather than P-384
   ([#662](https://github.com/sigstore/sigstore-python/pull/662))
 
@@ -68,6 +68,11 @@ All versions prior to 0.9.0 are untracked.
 * Removed an unnecessary and backwards-incompatible parameter from the
   `sigstore.oidc.detect_credential` API
   ([#641](https://github.com/sigstore/sigstore-python/pull/641))
+
+* Fixed a case where `sigstore sign` (and `sigstore verify`) could fail while
+  using a private instance due to a missing due to a missing `ExtendedKeyUsage`
+  in the CA. We now enforce the fact that the TBSPrecertificate signer must be
+  a valid CA ([#658](https://github.com/sigstore/sigstore-python/pull/658))
 
 ## [1.1.2]
 
--- a/sigstore/_utils.py
+++ b/sigstore/_utils.py
@@ -226,28 +226,24 @@ def cert_is_ca(cert: Certificate) -> bool:
         # No BasicConstrains means that this can't possibly be a CA.
         return False
 
-    digital_signature = False
     key_cert_sign = False
     try:
         key_usage = cert.extensions.get_extension_for_oid(ExtensionOID.KEY_USAGE)
         key_cert_sign = key_usage.value.key_cert_sign  # type: ignore
-        digital_signature = key_usage.value.digital_signature  # type: ignore
     except ExtensionNotFound:
         raise InvalidCertError("invalid X.509 certificate: missing KeyUsage")
 
-    # If all three states are set, this is a CA.
-    if ca and key_cert_sign and digital_signature:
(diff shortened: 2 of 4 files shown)


woodruffw:
Thanks @ccordoui! For maximum pedantry: we probably also need the `CA:true` and similar RFC 5280 profile checks for the CA, right?

(Given that we don't build the chain anyways, I'm okay with skipping this for the purposes of this PR. But I want to make sure I'm not missing anything.)

woodruffw (commented):
LGTM; needs a "fixed" `CHANGELOG.md` entry 🙂

ccordoui:
> we probably also need the CA:true and similar RFC 5280 profile checks for the CA, right?

Yes, will do.

> LGTM; needs a "fixed" CHANGELOG.md entry 🙂

I'll add that too.

Thanks @woodruffw!

woodruffw:
No problem, thanks to you as well! You may be able to reuse one of the cert utility functions that I added under `utils` 🙂

ccordoui:
I have to check why I have an expired client certificate using the signing unit tests, but outside of that, it seems that using `cert_is_ca` from `utils` like I did is breaking the `sigstore sign` 😞:

```python
    def cert_is_ca(cert: Certificate) -> bool:
>       if cert.version != Version.v3:

E       AttributeError: '_EllipticCurvePublicKey' object has no attribute 'version'
```

woodruffw:
> I couldn't find a RFC specifying that we need the `digitalSignature` KU to be a CA, what did I miss?

Nothing 🙂 -- I added that check based on observed behaviors for the Sigstore CAs, but it's not actually mandated in RFC 5280 for CAs.

I don't think there's any harm in removing that check on `digitalSignature` entirely, but I'll check the current PKI profile draft for Sigstore and confirm.

woodruffw (approved):

CHANGELOG.md
```diff
@@ -69,6 +69,12 @@ All versions prior to 0.9.0 are untracked.
   `sigstore.oidc.detect_credential` API
   ([#641](https://github.com/sigstore/sigstore-python/pull/641))
 
+* Fixed a case where `sigstore sign` could fail using a private instance
+  (and `sigstore verrify`) due to a missing due to a missing `ExtendedKeyUsage`
+  in the CA
+  We now enforce the fact that the TBSPrecertificate signer must be a valid CA
+  ([#658](https://github.com/sigstore/sigstore-python/pull/658))
```
woodruffw, line 76:
```suggestion
* Fixed a case where `sigstore sign` (and `sigstore verify`) could fail while
  using a private instance due to a missing due to a missing `ExtendedKeyUsage`
  in the CA. We now enforce the fact that the TBSPrecertificate signer must be
  a valid CA ([#658](https://github.com/sigstore/sigstore-python/pull/658))
```

sigstore/_utils.py
```diff
@@ -226,28 +226,24 @@ def cert_is_ca(cert: Certificate) -> bool:
...
-    digital_signature = False
     key_cert_sign = False
     try:
         key_usage = cert.extensions.get_extension_for_oid(ExtensionOID.KEY_USAGE)
         key_cert_sign = key_usage.value.key_cert_sign  # type: ignore
-        digital_signature = key_usage.value.digital_signature  # type: ignore
     except ExtensionNotFound:
         raise InvalidCertError("invalid X.509 certificate: missing KeyUsage")
 
     # If all three states are set, this is a CA.
```
woodruffw, line 236:
```suggestion
    # If both states are set, this is a CA.
```

woodruffw (approved):

tetsuo-cpp (approved):

woodruffw:
@ccordoui mind updating the branch here? If I do it, we'll need to do another hop of reviewers because of the merge restrictions 🙃

ccordoui:
> @ccordoui mind updating the branch here? If I do it, we'll need to do another hop of reviewers because of the merge restrictions 🙃

A rebase will do?

woodruffw:
> @woodruffw a rebase will do?

Yes, that should be fine -- anything that updates the `HEAD` to make the ownership not mine.

PR #682, opened 2023-06-20
Fix missing SigningContext import in sign example

mayaCostantini:
#### Summary

A really small fix to include the `SigningContext` import in the example given in `sign.py`.

#### Release Note

NONE

### DIFF
--- a/sigstore/sign.py
+++ b/sigstore/sign.py
@@ -20,7 +20,7 @@ Example:
 ```python
 from pathlib import Path
 
-from sigstore.sign import Signer, SigningContext
+from sigstore.sign import SigningContext
 from sigstore.oidc import Issuer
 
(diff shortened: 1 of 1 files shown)


sigstore/sign.py
```diff
@@ -20,7 +20,7 @@
 ```python
 from pathlib import Path
 
-from sigstore.sign import Signer
+from sigstore.sign import Signer, SigningContext
```
woodruffw, line 23: Nit: that means this import is no longer necessary, I believe 🙂 

```suggestion
from sigstore.sign import SigningContext
```

woodruffw (approved):
LGTM, thank you!

woodruffw (approved):

tetsuo-cpp (approved):

PR #684, opened 2023-06-22
Bump `sigstore-conformance` to 0.0.5

tetsuo-cpp:


.github/workflows/conformance.yml
```diff
@@ -25,6 +25,6 @@ jobs:
       - name: install sigstore-python
         run: python -m pip install .
 
-      - uses: sigstore/sigstore-conformance@064fb32a890c30235f305281f3509c5e65e6f9e5 # v0.0.4
+      - uses: sigstore/sigstore-conformance@fc986104ac997147aa9739a06aa34bc1df909207
```
tetsuo-cpp, line 28: Bump this again and add a version comment once we cut a release.

woodruffw:
Hmm, the test failures here don't make a ton of sense -- it looks like the Rekor API is no longer returning valid JSON on HTTP errors, which in turn is surfacing because some endpoints are failing when we aren't expecting them to.

In particular, it looks like the Rekor entry lookup API is failing with HTTP 429, indicating rateliming. Maybe the staging instance is ratelimiting us too conservatively? CC @asraa and @haydentherapper (sorry if you're the wrong people to ping on this!)

tetsuo-cpp:
Yeah, all of our PRs are hitting this.

haydentherapper:
fyi @bobcallaway 

We're in the process of moving over to a new load balancer, we've just started with staging, and there's a few changes with rate limiting and responses for errors. Do you know what your average QPS is for these tests? You might be hitting a rate limit because all GHAs share a set of IPs.

re: errors and json, we'll have to look into this.

Thanks for the ping, we'll get this ironed out.

woodruffw:
> Do you know what your average QPS is for these tests? You might be hitting a rate limit because all GHAs share a set of IPs.

I don't have a firm number, but I'd expect it to be (well) under 10/second. Shared IPs seems likely a likely culprit :slightly_smiling_face:

woodruffw (approved):

bobcallaway:
> > Do you know what your average QPS is for these tests? You might be hitting a rate limit because all GHAs share a set of IPs.
> 
> I don't have a firm number, but I'd expect it to be (well) under 10/second. Shared IPs seems likely a likely culprit slightly_smiling_face

I'm bumping up the limit now, hopefully will unblock this.

PR #685, opened 2023-06-22
sigstore: 2.0.0rc1

tetsuo-cpp:


woodruffw:
@tetsuo-cpp are we trying to land something specific with this? This point release will include a bunch of breaking changes, so if we just need a new release then IMO a `2.0.0rc0` or similar would be appropriate 🙂

tetsuo-cpp:
> @tetsuo-cpp are we trying to land something specific with this? This point release will include a bunch of breaking changes, so if we just need a new release then IMO a `2.0.0rc0` or similar would be appropriate 🙂

We've previously started from rc1 I think so I've chosen `2.0.0rc1`.

woodruffw (approved):

PR #694, opened 2023-06-28
don't parse error messages as json

bobcallaway:
#### Summary
I don't think we need to pass a parsed JSON object up to the caller in the context of `RekorClientError`

@woodruffw thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,10 @@ All versions prior to 0.9.0 are untracked.
 * `sigstore sign` now uses an ephemeral P-256 keypair, rather than P-384
   ([#662](https://github.com/sigstore/sigstore-python/pull/662))
 
+* API change: `RekorClientError` does not try to always parse response
+  content as JSON
+  ([#694](https://github.com/sigstore/sigstore-python/pull/694))
+
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


woodruffw (dismissed):
Agreed!

Does the Rekor REST API specify anywhere that it _may_ return JSON vs. plain text on error codes? I don't think it's a significant issue, but if we ever elaborate on this exception type it'd be nice to know what we can rely on/assume.

bobcallaway:
> Nonblocking, but mind adding a small CHANGELOG entry as well? Something under the "Fixed" category to the effect of "Fixed an assumption about Rekor error responses that was causing crashes"
> 
> If not, I can add it later.

I'll add it, no problem.

bobcallaway:
> Agreed!
> 
> Does the Rekor REST API specify anywhere that it _may_ return JSON vs. plain text on error codes? I don't think it's a significant issue, but if we ever elaborate on this exception type it'd be nice to know what we can rely on/assume.

not really? The `Content-Type` on the response is `text/html; charset=UTF-8` so while I don't think we're out of HTTP spec, I'm not sure where we'd declare it (perhaps in the `openapi.yaml` file?)

di (approved):

Issue #700, opened 2023-07-06
Certificate timestamp issue when checking for validity datetime


perone:
The field `not_valid_after` in the certificate is not tz-aware, therefore when a check is done using this timestamp it will assume that it is localtime, changing the timestamp and causing the sign operation to always trigger a `ExpiredCertificate`.

Reproduces in version: `main` (2.0.0rc1)

woodruffw:
Thanks for catching this @perone, and for using our release candidate! I'll take a look at #701 now.

PR #701, opened 2023-07-06
Add timezone (utc) info into the cert not_valid_after field

perone:
#### Summary
This commit fixes the issue described in #700 where the certificate `not_valid_after` datetime is not tz-aware, making the call to `timestamp()` to assume local time, triggering the `ExpiredCertificate` exception all the time.

Resolves: #700 

#### Release Note
- Fixes a issue in the certificate expiration timestamp that caused exception during signing process.

### DIFF
--- a/sigstore/sign.py
+++ b/sigstore/sign.py
@@ -138,11 +138,8 @@ class Signer:
         # If it exists, verify if the current certificate is expired
         if self.__cached_signing_certificate:
             not_valid_after = self.__cached_signing_certificate.cert.not_valid_after
-            not_valid_after_timestamp = not_valid_after.replace(
-                tzinfo=timezone.utc
-            ).timestamp()
-
-            if datetime.now(timezone.utc).timestamp() > not_valid_after_timestamp:
+            not_valid_after_tzutc = not_valid_after.replace(tzinfo=timezone.utc)
+            if datetime.now(timezone.utc) > not_valid_after_tzutc:
                 raise ExpiredCertificate
             return self.__cached_signing_certificate
 
--- a/sigstore/sign.py
+++ b/sigstore/sign.py
@@ -137,10 +137,12 @@ class Signer:
         """Get or request a signing certificate from Fulcio."""
         # If it exists, verify if the current certificate is expired
         if self.__cached_signing_certificate:
-            if (
-                datetime.now(timezone.utc).timestamp()
-                > self.__cached_signing_certificate.cert.not_valid_after.timestamp()
-            ):
+            not_valid_after = self.__cached_signing_certificate.cert.not_valid_after
+            not_valid_after_timestamp = not_valid_after.replace(
+                tzinfo=timezone.utc
+            ).timestamp()
+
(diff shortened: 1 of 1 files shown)


sigstore/sign.py
```diff
@@ -137,10 +137,12 @@ def _signing_cert(
...
-            if (
-                datetime.now(timezone.utc).timestamp()
-                > self.__cached_signing_certificate.cert.not_valid_after.timestamp()
-            ):
+            not_valid_after = self.__cached_signing_certificate.cert.not_valid_after
+            not_valid_after_timestamp = not_valid_after.replace(
+                tzinfo=timezone.utc
+            ).timestamp()
+
+            if datetime.now(timezone.utc).timestamp() > not_valid_after_timestamp:
```
woodruffw, line 145: Hmm, I think we can simplify this even further by just comparing the `datetime` objects directly: `cryptography` guarantees that `Certificate` times are always UTC, so this should work:

```python
if datetime.now(timezone.utc) > self.__cached_signing_certificate.cert.not_valid_after:
    ...
```

That way we don't have to do any timestamp conversions at all 🙂 

Could you confirm that this works for you, @perone?

woodruffw: (NB: This might fail because of an aware vs. non-aware comparison, in which case we can use `datetime.utcnow()` instead of `datetime.now(timezone.utc)`).

perone: Thanks for reviewing @woodruffw. I would avoid the `utcnow()` [[1]](https://www.aaronoellis.com/articles/python-datetime-utcnow-considered-harmful) [[2]](https://blog.ganssle.io/articles/2019/11/utcnow.html). I will test it to check how it behaves.

perone: Indeed it triggers an exception:
```
TypeError: can't compare offset-naive and offset-aware datetimes
```

Another alternative is to go with the option of having both tz-aware datetimes but without timestamp conversion:

```python
not_valid_after = self.__cached_signing_certificate.cert.not_valid_after
not_valid_after_tzutc = not_valid_after.replace(
    tzinfo=timezone.utc
)
if datetime.now(timezone.utc) > not_valid_after_tzutc:
    raise ExpiredCertificate
```

What do you think ?

woodruffw: That works for me -- frankly I'd also be okay with `utcnow` despite the footguns, since equality isn't defined between the two kinds of datetimes (the `TypeError` you noticed) and it's what `cryptography` uses internally (_and_ this object never gets exposed anywhere).

That being said, I'm fine with either, so whichever you prefer to do!

perone:
PTAL @woodruffw

sigstore/sign.py
woodruffw: (I take back what I said about `utcnow` being acceptable -- Python is marking it deprecated starting with 3.12, so we shouldn't use it at all. My bad!)

woodruffw (approved):
LGTM, thank you @perone!

woodruffw:
Thanks again @perone, especially for testing our RCs!

perone:
Thanks for the quick reviews @woodruffw !

Issue #654, opened 2023-05-19
Verification: Switch to new-style claim extensions


woodruffw:
Per https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md: the new claim extensions have been stabilized and are present in all currently issued certificates. We should prefer them when possible.

woodruffw:
This isn't a breaking change from the user's perspective, so I'm going to defer this until after 2.0.

woodruffw:
Opened #715 for part of this.

woodruffw:
Deduping: #425.

PR #715, opened 2023-07-24
policy: initial new extension OIDs

woodruffw:
Can't use these yet because they have nested DER encodings.

xref https://github.com/pyca/cryptography/issues/9283 for one possible route forward here.

See #654.


di:
Is this for https://github.com/sigstore/sigstore-python/issues/654?

woodruffw:
Obviated by #1004

Issue #718, opened 2023-07-27
Add a method to combine certificate and signature files into a bundle


sethmlarson:
**Description**

Currently on python.org/downloads we have a mix of releases that either have certificate/signatures separate or that with a bundle as well. We'd like to move to using bundles but in order to simplify the guidance we give users on verifying the release artifacts we'd like to backfill bundles to the releases that don't have bundles.

In order to accomplish this, I discussed on Slack with @woodruffw and we decided on a subcommand `sigstore bundle` which would take a signature and certificate, do a TLog lookup similar to verifying, and then output a bundle from those materials.

woodruffw:
Thanks @sethmlarson! I'm going to hijack the top-level comment here to copy over what I've written up (preserving your summary as well) 🙂

woodruffw:
CC @di for opinions here, as well as @haydentherapper as a `cosign` maintainer (maybe you've had a similar problem/requirements?)

haydentherapper:
The closest analogy in Cosign would be the `cosign attach` commands that upload detached metadata to a container in a registry. We're just starting to add bundle support in Cosign, so I haven't thought through this case, but I think that's a great suggestion.

woodruffw:
Teeing this up now that we have a `sigstore plumbing` command hierarchy.

PR #719, opened 2023-07-27
Add VerificationMaterials.to_bundle()

sethmlarson:
#### Summary

Part of https://github.com/sigstore/sigstore-python/issues/718, adds the API to convert `VerificationMaterials` that have a Rekor entry loaded into a `Bundle`.

#### Release Note

Will add a release note.

#### Documentation

I don't think this needs a docs update?

### DIFF
--- a/sigstore/verify/models.py
+++ b/sigstore/verify/models.py
@@ -462,6 +462,26 @@ class VerificationMaterials:
                 "Must have Rekor entry before converting to a Bundle"
             )
         rekor_entry = self._rekor_entry
+        assert rekor_entry is not None
+
+        inclusion_proof: InclusionProof | None = None
+        if rekor_entry.inclusion_proof is not None:
+            inclusion_proof = InclusionProof(
+                log_index=rekor_entry.inclusion_proof.log_index,
+                root_hash=bytes.fromhex(rekor_entry.inclusion_proof.root_hash),
+                tree_size=rekor_entry.inclusion_proof.tree_size,
+                hashes=[
+                    bytes.fromhex(hash_hex)
+                    for hash_hex in rekor_entry.inclusion_proof.hashes
+                ],
+                checkpoint=Checkpoint(envelope=rekor_entry.inclusion_proof.checkpoint),
+            )
+
+        inclusion_promise: InclusionPromise | None = None
+        if rekor_entry.inclusion_promise:
+            inclusion_promise = InclusionPromise(
+                signed_entry_timestamp=base64.b64decode(rekor_entry.inclusion_promise)
+            )
 
         bundle = Bundle(
             media_type="application/vnd.dev.sigstore.bundle+json;version=0.2",
@@ -480,25 +500,8 @@ class VerificationMaterials:
                         log_id=LogId(key_id=bytes.fromhex(rekor_entry.log_id)),
                         kind_version=KindVersion(kind="hashedrekord", version="0.0.1"),
                         integrated_time=rekor_entry.integrated_time,
-                        inclusion_promise=InclusionPromise(
-                            signed_entry_timestamp=base64.b64decode(
-                                rekor_entry.inclusion_promise
-                            )
-                        ),
-                        inclusion_proof=InclusionProof(
-                            log_index=rekor_entry.inclusion_proof.log_index,
-                            root_hash=bytes.fromhex(
-                                rekor_entry.inclusion_proof.root_hash
-                            ),
-                            tree_size=rekor_entry.inclusion_proof.tree_size,
-                            hashes=[
-                                bytes.fromhex(hash_hex)
-                                for hash_hex in rekor_entry.inclusion_proof.hashes
-                            ],
-                            checkpoint=Checkpoint(
-                                envelope=rekor_entry.inclusion_proof.checkpoint
-                            ),
-                        ),
+                        inclusion_promise=inclusion_promise,
+                        inclusion_proof=inclusion_proof,
                         canonicalized_body=base64.b64decode(rekor_entry.body),
                     )
                 ],
--- a/sigstore/verify/models.py
+++ b/sigstore/verify/models.py
@@ -461,8 +461,7 @@ class VerificationMaterials:
             raise InvalidMaterials(
                 "Must have Rekor entry before converting to a Bundle"
             )
-        rekor_entry = self._rekor_entry
-        assert rekor_entry is not None
+        rekor_entry: LogEntry = self._rekor_entry  # type: ignore[assignment]
 
         inclusion_proof: InclusionProof | None = None
         if rekor_entry.inclusion_proof is not None:
--- a/sigstore/verify/models.py
+++ b/sigstore/verify/models.py
@@ -32,10 +32,25 @@ from cryptography.x509 import (
     load_pem_x509_certificate,
 )
 from pydantic import BaseModel
-from sigstore_protobuf_specs.dev.sigstore.bundle.v1 import Bundle
+from sigstore_protobuf_specs.dev.sigstore.bundle.v1 import (
+    Bundle,
+    VerificationMaterial,
+)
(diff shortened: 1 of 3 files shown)


sigstore/verify/models.py
```diff
@@ -171,8 +186,6 @@ class VerificationMaterials:
     certificate: Certificate
     """
     The certificate that attests to and contains the public signing key.
```
sethmlarson, line 188: Can revert this TODO removal, had it in place when I was still considering certificate chains.

sigstore/verify/models.py
```diff
@@ -438,3 +451,62 @@ def rekor_entry(self, client: RekorClient) -> LogEntry:
...
 
         return entry
+
+    def to_bundle(self) -> Bundle:
+        """Converts VerificationMaterials into a Bundle. Requires that
+        the VerificationMaterials have a Rekor entry loaded. This is
+        the reverse operation of VerificationMaterials.from_bundle()
+        """
+        if not self.has_rekor_entry:
+            raise ValueError("Must have Rekor entry before converting to a Bundle")
```
woodruffw, line 461: Nitpick: exception type here:

```suggestion
            raise InvalidMaterials("Must have Rekor entry before converting to a Bundle")
```

sigstore/verify/models.py
```diff
@@ -438,3 +451,62 @@ def rekor_entry(self, client: RekorClient) -> LogEntry:
             raise InvalidRekorEntry
 
         return entry
+
+    def to_bundle(self) -> Bundle:
+        """Converts VerificationMaterials into a Bundle. Requires that
+        the VerificationMaterials have a Rekor entry loaded. This is
+        the reverse operation of VerificationMaterials.from_bundle()
+        """
+        if not self.has_rekor_entry:
```
woodruffw, line 460: This makes sense to me, but just to confirm we're on the same page: are you thinking that the CLI that uses this will pre-populate the Rekor entry? I think that's the right way to do this; just making sure 🙂

test/unit/verify/test_models.py
```diff
@@ -86,3 +87,13 @@ def test_verification_materials_offline_no_checkpoint(self, signing_bundle):
             InvalidMaterials, match="expected checkpoint in inclusion proof"
         ):
             signing_bundle("bundle_no_checkpoint.txt", offline=True)
+
+    def test_verification_materials_to_bundle_round_trip(self, asset, signing_bundle):
```
woodruffw, line 91: Thanks! Mind also adding a testcase for the `InvalidMaterials` case above? It should be "just" a matter of loading `VerificationMaterials` from a `.crt` and `.sig` instead and then not explicitly loading in the Rekor entry.

sigstore/verify/models.py
sethmlarson: That's how I was thinking about it yes!

woodruffw:
> #### Documentation
> I don't think this needs a docs update?

Nope, I don't think so -- it'll be auto-generated.

test/unit/verify/test_models.py
sethmlarson: Can do!

sethmlarson:
@woodruffw Implemented your comments. I force pushed because I needed to add a `--signoff`.

woodruffw:
> @woodruffw Implemented your comments. I force pushed because I needed to add a `--signoff`.

Thanks, LGTM! I think this just needs a CHANGELOG entry and then we're good to go.

test/unit/verify/test_models.py
```diff
@@ -86,3 +87,24 @@ def test_verification_materials_offline_no_checkpoint(self, signing_bundle):
...
+        bundle = signing_bundle("bundle.txt").to_bundle()
+
+        with asset("bundle.txt").open(mode="rb", buffering=0) as io:
+            round_tripped_bundle = VerificationMaterials.from_bundle(
+                input_=io, bundle=bundle, offline=False
```
woodruffw, line 96: I could be wrong, but since we're loading a bundle that contains an inclusion proof, I think this can be `offline=True`:

```suggestion
                input_=io, bundle=bundle, offline=True
```

woodruffw:
Lintage:

```
sigstore/verify/models.py:479: error: Item "None" of "Optional[LogEntry]" has no attribute "log_index"  [union-attr]
sigstore/verify/models.py:480: error: Item "None" of "Optional[LogEntry]" has no attribute "log_id"  [union-attr]
sigstore/verify/models.py:482: error: Item "None" of "Optional[LogEntry]" has no attribute "integrated_time"  [union-attr]
sigstore/verify/models.py:485: error: Item "None" of "Optional[LogEntry]" has no attribute "inclusion_promise"  [union-attr]
sigstore/verify/models.py:485: error: Argument 1 to "b64decode" has incompatible type "Union[B64Str, None, Any]"; expected "Union[str, Buffer]"  [arg-type]
sigstore/verify/models.py:489: error: Item "None" of "Optional[LogEntry]" has no attribute "inclusion_proof"  [union-attr]
sigstore/verify/models.py:489: error: Item "None" of "Union[LogInclusionProof, None, Any]" has no attribute "log_index"  [union-attr]
sigstore/verify/models.py:491: error: Item "None" of "Optional[LogEntry]" has no attribute "inclusion_proof"  [union-attr]
sigstore/verify/models.py:491: error: Item "None" of "Union[LogInclusionProof, None, Any]" has no attribute "root_hash"  [union-attr]
sigstore/verify/models.py:493: error: Item "None" of "Optional[LogEntry]" has no attribute "inclusion_proof"  [union-attr]
sigstore/verify/models.py:493: error: Item "None" of "Union[LogInclusionProof, None, Any]" has no attribute "tree_size"  [union-attr]
sigstore/verify/models.py:496: error: Item "None" of "Optional[LogEntry]" has no attribute "inclusion_proof"  [union-attr]
sigstore/verify/models.py:496: error: Item "None" of "Union[LogInclusionProof, None, Any]" has no attribute "hashes"  [union-attr]
sigstore/verify/models.py:499: error: Item "None" of "Optional[LogEntry]" has no attribute "inclusion_proof"  [union-attr]
sigstore/verify/models.py:499: error: Item "None" of "Union[LogInclusionProof, None, Any]" has no attribute "checkpoint"  [union-attr]
sigstore/verify/models.py:502: error: Item "None" of "Optional[LogEntry]" has no attribute "body"  [union-attr]
Found 16 errors in 1 file (checked [27](https://github.com/sigstore/sigstore-python/actions/runs/5683866017/job/15406553799?pr=719#step:5:28) source files)
```

sigstore/verify/models.py
```diff
@@ -438,3 +451,67 @@ def rekor_entry(self, client: RekorClient) -> LogEntry:
...
+            raise InvalidMaterials(
+                "Must have Rekor entry before converting to a Bundle"
+            )
+        rekor_entry = self._rekor_entry
+        assert rekor_entry is not None
```
woodruffw, line 465: `bandit` doesn't like `assert`, maybe we can make this a `type: ignore` with an explicit annotation to make `mypy` happy?

sethmlarson: Done!

woodruffw (dismissed):
LGTM, thanks @sethmlarson!

Were you planning on adding a `CHANGELOG` entry? If not, I can add one in a bit.

sethmlarson:
@woodruffw If you could add one that'd be appreciated, I couldn't immediately figure out what the right annotation would be :)

woodruffw:
> If you could add one that'd be appreciated, I couldn't immediately figure out what the right annotation would be :)

It'd probably be "added" for a new public API surface, just for future reference! But I'll push to your branch.

woodruffw (approved):
LGTM, but I'm the last pusher so @di or @tetsuo-cpp will need to merge here.

tetsuo-cpp (approved):

PR #723, opened 2023-07-31
models: require checkpoint in embedded inclusion proof

woodruffw:


.github/workflows/conformance.yml
```diff
@@ -22,6 +22,6 @@ jobs:
       - name: install sigstore-python
         run: python -m pip install .
 
-      - uses: sigstore/sigstore-conformance@954233f9a4ea6d0b355626b283b4f837f465e4ec # v0.0.5
+      - uses: sigstore/sigstore-conformance@887e2090d0f5f8a4487b778d0bf6448e55ed4005
```
woodruffw, line 25: NB: This is the current `HEAD` of the conformance suite; we can either merge as-is or block this PR on https://github.com/sigstore/sigstore-conformance/pull/92

sigstore/transparency.py
```diff
@@ -131,7 +131,13 @@ def __post_init__(self) -> None:
         """
         Invariant preservation.
         """
-        if self.inclusion_proof is None and self.inclusion_promise is None:
+
+        # An inclusion proof isn't considered present unless its checkpoint
+        # is also present.
+        has_incluson_proof = (
+            self.inclusion_proof is not None and self.inclusion_proof.checkpoint
```
tetsuo-cpp, line 138: Do we want to log something when there's an inclusion proof but no checkpoint? Seems like it might be a bit confusing that we silently ignore the inclusion proof.

woodruffw: Hmm, I need to think about where it makes sense to add this check + log: it might make more sense in `from_bundle`, since that's really the only place we'd expect user error to cause a mismatch here.

woodruffw: https://github.com/sigstore/sigstore-python/pull/723/commits/2ffcf089f34062db982358ee4fde32dda1064147

tetsuo-cpp (approved):
LGTM!

PR #745, opened 2023-08-29
Drop support for Python 3.7

di:
Python 3.7 is currently EOL: https://devguide.python.org/versions/

Current versions of `sigstore-python` (<=1.1.2) will continue to be usable from Python 3.7.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,7 +15,7 @@ jobs:
       # NOTE: We intentionally lint against our minimum supported Python.
       - uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1
         with:
-          python-version: "3.7"
+          python-version: "3.8"
           cache: "pip"
           cache-dependency-path: pyproject.toml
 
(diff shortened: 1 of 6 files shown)


.github/workflows/lint.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       # NOTE: We intentionally lint against our minimum supported Python.
       - uses: actions/setup-python@61a6322f88396a6271a6ee3565807d608ecaddd1
         with:
-          python-version: "3.7"
+          python-version: "3.8"
```
woodruffw, line 18: I think we can use `python-version-file` here now and elsewhere:

```suggestion
          python-version-file: "pyproject.toml"
```

That should reduce the number of locations we need to update these going forwards 🙂 

Ref: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-file-input

pyproject.toml
woodruffw, line 1: Not strictly necessary, but `ruff`'s `target-version` can also be bumped:

https://github.com/sigstore/sigstore-python/blob/a56b06839a6834bfb740c01ff06f2a337844eb65/pyproject.toml#L132

woodruffw (approved):

woodruffw (dismissed):

.github/workflows/lint.yml
woodruffw: Noting for future reference: this causes us to resolve the latest compatible version instead, which will generally be wrong in terms of what we want with linting (we want the earliest compatible version).

PR #778, opened 2023-09-26
Prep 2.0.0

woodruffw:
~~Draft until we're ready to merge this.~~

### DIFF
--- a/sigstore/transparency.py
+++ b/sigstore/transparency.py
@@ -24,9 +24,9 @@ from pydantic import (
     BaseModel,
     ConfigDict,
     Field,
-    FieldValidationInfo,
     StrictInt,
     StrictStr,
+    ValidationInfo,
     field_validator,
 )
 from pydantic.dataclasses import dataclass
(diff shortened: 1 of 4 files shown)


sigstore/transparency.py
```diff
@@ -24,9 +24,9 @@
     BaseModel,
     ConfigDict,
     Field,
-    FieldValidationInfo,
     StrictInt,
     StrictStr,
+    ValidationInfo,
```
woodruffw, line 29: I haven't root caused this, but this looks like a semver-breaking change in Pydantic.

woodruffw: Hmm, more borkage with Pydantic: https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/actions/runs/6315666569/job/17148672809

woodruffw: https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/pull/6

jku: > I haven't root caused this, but this looks like a semver-breaking change in Pydantic.

Either that or I used supposedly non-public API by accident... https://github.com/pydantic/pydantic/pull/7542 happened last friday

woodruffw: Ah yep, could be. It looks like `FieldValidationInfo` might have been deprecated at some point (prior to removal), but I can't find a good reference for that (or for why).

woodruffw:
This should be safe to merge whenever; we can create the release anytime after it's merged.

jku (approved):
LGTM.

For reference on that pydantic issue
* I believe FieldValidationInfo still works but the way they marked it deprecated (which seems to be what happened last friday) is what mypy flags as suspicious
* the fix here looks right to me

di (approved):

PR #791, opened 2023-10-13
Tweak some ignore-hints to make mypy 1.6 happy

jku:
These linter errors appeared first in https://github.com/sigstore/sigstore-python/pull/790  workflow run


jku:
Hmm, so current main branch with mypy 1.6 and py 3.11 complains 

    error: Unused "type: ignore" comment
    
but if I remove the ignore hint, then  mypy 1.6 and py 3.8 complains

    error: Returning Any from function declared to return "bytes"  [no-any-return]

I'll try some alternative hints...

woodruffw:
Thanks @jku!

Issue #628, opened 2023-04-24
Support DSSE-style enveloped signatures


woodruffw:
This is part of Sigstore bundle support: we currently only support "raw" signatures, while some users of Sigstore may chose to use enveloped DSSE-style signatures.

From protobuf-specs:

```proto
// An authenticated message of arbitrary type.
message Envelope {
  // Message to be signed. (In JSON, this is encoded as base64.)
  // REQUIRED.
  bytes payload = 1;

  // String unambiguously identifying how to interpret payload.
  // REQUIRED.
  string payloadType = 2;

  // Signature over:
  //     PAE(type, payload)
  // Where PAE is defined as:
  // PAE(type, payload) = "DSSEv1" + SP + LEN(type) + SP + type + SP + LEN(payload) + SP + payload
  // +               = concatenation
  // SP              = ASCII space [0x20]
  // "DSSEv1"        = ASCII [0x44, 0x53, 0x53, 0x45, 0x76, 0x31]
  // LEN(s)          = ASCII decimal encoding of the byte length of s, with no leading zeros
  // REQUIRED (length >= 1).
  repeated Signature signatures = 3;
}

message Signature {
  // Signature itself. (In JSON, this is encoded as base64.)
  // REQUIRED.
  bytes sig = 1;

  // *Unauthenticated* hint identifying which public key was used.
  // OPTIONAL.
  string keyid = 2;
}
```

...where `payload` is the to-be-signed attestation statement.

xref https://github.com/sigstore/fulcio/issues/1131

Work tracker:

- [x] Add `sigstore-rekor-types` (#788)
- [x] Use `sigstore-rekor-types` to provide DSSE creation and retrieval support in our Rekor bindings
- [x] Expose DSSE signing via the Python API (#804)
    - [x] https://github.com/sigstore/sigstore-python/pull/921
- [x] Expose DSSE verification via the Python API 
    - [x] https://github.com/sigstore/sigstore-python/pull/904
    - [x] #930 
    - [x] #962 
- [ ] (Maybe?) expose DSSE signing and verification via the `sigstore` CLI

woodruffw:
xref https://github.com/sigstore/fulcio/issues/1131 for original motivating context.

woodruffw:
Thinking about this some more, I think the "right" way to do this is similar to what `sigstore-js` has done:

1. We need a new `sigstore-rekor-types` (or similar) package that exposes Rekor's OpenAPI types as Python types
[...]

Issue #666, opened 2023-06-02
Provide API that take in a hash instead of io bytes


laurentsimon:
**Description**

The current sign and verify APIs take as input an `io[bytes]`:
- signing: https://github.com/sigstore/sigstore-python/blob/main/sigstore/sign.py#L119
- verifying: https://github.com/sigstore/sigstore-python/blob/main/sigstore/verify/models.py#L238

There are limitations to expecting an input buffer:

For signing, we may want to sign something like a tree structure, a folder, or something that does not have an obvious array representation and is application specific. In this case it would be useful to let the caller hash the content and pass it to the sign function. This would not work for edDSA signature - because they need the raw text message - but works for ECDSA.

For verification, same limitation. In addition, imagine you want to create a remote service (REST API) that verifies attestations / signatures. The signed message you may be too large to be sent over the network (e.g., a large SBOM), so you may want to let the caller hash the document and send over only the attestation and the hash of the message. There is one limitation to allowing this: it's is more prone to mistakes. There have been cases where a verifier would mistakenly take in the hash provided by an attacker instead of the message they intended to verify. So maybe we could dangerous APIs under a "danger" folder. FWIW, the Tink crypto library puts dangerous functions under a `subtle` folder.

woodruffw:
> For signing, we may want to sign something like a tree structure, a folder, or something that does not have an obvious array representation and is application specific.

I'm not sure that we want to encourage this on an ecosystem level -- if you want to sign over a tree structure or a directory or something like that, then you should canonicalize that into something that looks like a stream of bytes (e.g., a `tar` archive) and sign over that. Could you say more about your use case, and why that kind of "directory-to-file" representation wouldn't work?

> In this case it would be useful to let the caller hash the content and pass it to the sign function. This would not work for edDSA signature - because they need the raw text message - but works for ECDSA.

FWIW, this is essentially what we do internally (since we current use ECDSA, as you noted) -- we hash the message ahead of time and pass it into the underlying signing APIs as a `Prehashed` object. 

I wouldn't be opposed to exposing that as a public API as well, although I'm hesitant to further introduce public expectations around the fact that our current signature algorithm is defined over the digest rather than the input itself. As you noted, that won't work for EdDSA, and we _may_ want to switch to EdDSA in the near future (although ideally we'd switch to Ed25519-ph, which would avoid that problem).

[...]

PR #804, opened 2023-10-18
API-level DSSE signing support

woodruffw:
Adds support for DSSE signature generation (over an in-toto statement) at the API level. Not exposed via the `sigstore` CLI yet.

See #628.


woodruffw:
CCing @laurentsimon, just so you're aware: this PR will change the `sign(...)` API a bit, so #666 may be affected by it 🙂

sigstore/sign.py
```diff
@@ -174,10 +177,9 @@ def _signing_cert(
 
     def sign(
         self,
-        input_: IO[bytes],
+        input_: IO[bytes] | Statement,
```
woodruffw, line 180: Flagging for review: this puts a third-part type (in-toto-attestation's `Statement`) in our public API.

Normally I'd be neutral on this, but it has a few other "meh" factors:

1. `in-toto-attestation` is still pre-1.0, meaning that we probably shouldn't lash our stable semver'd public API to it;
2. `Statement` is a thin wrapper for a `protobuf` generated `protobuf.Message` class, which both leaks `protobuf` implementation details into the public API *and* may change soon-ish (see https://github.com/in-toto/attestation/issues/291).

Another option here would be to create our own `sigstore.intoto_or_whatever.Statement` type, which would be a thin constructable shim over `in_toto_attestation.Statement`.

sigstore/sign.py
```diff
@@ -308,31 +331,46 @@ def signer(
         yield Signer(identity_token, self, cache)
 
 
-class SigningResult(BaseModel):
+@dataclass
+class SigningResult:
```
woodruffw, line 335: Flagging for review: the changes here indicate that we should probably just drop this type entirely and switch to `Bundle` as our "signing result" type. It'd be strictly equivalent and would save us a layer of types.

woodruffw:
For posterity: this is a small demo script that you can use to exercise the new DSSE functionality:

```python
#!/usr/bin/env python

import logging

from in_toto_attestation.v1.resource_descriptor import ResourceDescriptor
from in_toto_attestation.v1.statement import Statement

logging.getLogger().setLevel(logging.DEBUG)

ctx = SigningContext.staging()
with ctx.signer(identity_token=Issuer.staging().identity_token()) as signer:
    stmt = Statement(
        subjects=[
            ResourceDescriptor(
                name="null",
                digest={
                    "sha256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
                },
            ).pb,
        ],
        predicate={"Data": "", "Timestamp": "2023-12-07T00:37:58Z"},
        predicate_type="https://cosign.sigstore.dev/attestation/v1",
    )
    res = signer.sign(stmt)
    print(res)
```

woodruffw:
R4R but marking as blocked, pending resolution of https://github.com/in-toto/attestation/pull/315

tetsuo-cpp:
Will get onto this at the beginning of next week.

sigstore/sign.py
tetsuo-cpp: I'm not against a shim, but I don't see this causing a problem in practice. Even if/when they move to betterproto, the way users will construct a `Statement` will be pretty much identical. If the `in-toto-attestation` API completely changes before 1.0 to the point where it somehow becomes painful for our users (maybe their API takes different types?), we'll probably need to make breaking changes to the shim anyway.

sigstore/sign.py
tetsuo-cpp: I think consolidating to using `Bundle` is a good idea. I'm happy to follow up with a change to do that to keep things simple in this PR.

tetsuo-cpp (approved):
LGTM!

sigstore/sign.py
jku: This sounds right

jku (commented):
I have nothing to add here but left some high level questions in the issue

sigstore/sign.py
woodruffw: Thanks both! @tetsuo-cpp, I'll assign you a follow-up for removing `SigningResult`.

woodruffw:
For tracking purposes: our upstream blocker is itself blocked on https://github.com/danielgtaylor/python-betterproto/pull/551

sigstore/sign.py
woodruffw: That's fair. I'll consider this a non-blocker for now, which means this PR is good to go.

woodruffw:
This should be good to go -- I'm setting to auto-merge, pending review 🙂

woodruffw:
> * adding this signing capability without defining the verification side is a bit sketchy (even if I understand that it might not make sense to implement inside sigstore-python): I expect after this PR even the sigstore-python CLI might happily "verify" a bundle with an intoto statement?  This is not entirely wrong but could maybe give the wrong impression to the user (implying that something in the statement has been verified, when it has not).

Yeah, I'm not sure what to do about this -- IMO sigstore-python itself should never attempt to verify these statements (that belongs at a higher-up level), so maybe this is an API documentation problem? Something like a big warning on `verify(...)` saying that the API only verifies the signature itself modulo `VerificationPolicy`, which only concerns the Sigstore bits and not any arbitrary metadata in the payload itself?

> I think merging as is (after conflict fixes) should be ok though if the above items are in plans.

Yeah, I'd like to do an E2E test here once we have verification in place. I could probably hack it together as an integration-style test with another client for verification, but I figure it's not worth the effort versus merging now + working on a verification test in the follow-up.

jku:
> Yeah, I'd like to do an E2E test here once we have verification in place. I could probably hack it together as an integration-style test with another client for verification

Maybe I'm missing something but couldn't you do a test that does signing like your demo and then verifies the bundle (leaving a TODO for verifying the intoto statement) -- just to make sure the round trip works fine?

>> I expect after this PR even the sigstore-python CLI might happily "verify" a bundle with an intoto statement?

I wonder if I'm correct here and if we want to do something about this?
* should the CLI **not** succeed when the payload is a intoto statement (until we know how to expose this)
* should there should be some output on CLI when the payload is a intoto statement

I'm really not sure, this is just hand waving

woodruffw:
> Maybe I'm missing something but couldn't you do a test that does signing like your demo and then verifies the bundle (leaving a note that the intoto statement is not verified) -- just to make sure the round trip works fine?

Not with the current APIs -- I've only implemented DSSE signing, not verification. The verification APIs still expect only a "bare" signature, so attempting to verify a bundle containing a DSSE message will fail (probably with an abstruse Pydantic error).

> I wonder if I'm correct here and if we want to do something about this?

Yes, this is why I don't want to expose these APIs via the CLI in the immediate term 🙂 -- my plan was to have the APIs themselves present in the 3.0 series, after which we can iterate on an appropriate CLI design.

jku (approved):
Okay, sounds good then :+1:

PR #811, opened 2023-10-27
treewide: switch to `ruff format`

woodruffw:
This speeds up our `lint` and `reformat` targets a bit, and drops a dev dependency (since we're already using ruff).


tetsuo-cpp (approved):
LGTM! This is awesome.

Issue #814, opened 2023-11-16
Support `{input}.sigstore.json` 


woodruffw:
Per the [Sigstore Client Spec](https://docs.google.com/document/d/1kbhK2qyPPk8SLavHzYSDM8-Ueul9_oxIMVFuWMWKz0E/edit)

Two parts:

1. `sigstore sign` should _probably_ produce `{input}.sigstore.json` by default, rather than `{input}.sigstore` 
2. `sigstore verify` should accept either `{input}.sigstore.json` *or* `{input}.sigstore`, whichever is present (and probably fail if both are present)

(2) is pretty straightforward, and can be done in a backwards-compatible manner. (1) is a breaking change, so we'll either need to feature flag it or defer it until 3.0.

woodruffw:
Opened #820 for (2) above.

woodruffw:
Marking with 3.0, since the default `sigstore sign` change will be incompatible.

PR #820, opened 2023-12-04
cli: search for `{input}.sigstore.json` by default

woodruffw:
This tweaks the default input collection in `sigstore verify`, allowing `{input}.sigstore.json` to take priority when attempting to discover "implicit" inputs. 

`{input}.sigstore` is also still discovered, preserving compatibility with existing outputs. However, discovering it now produces a logged warning telling the user that a future release will deprecate `.sigstore` in favor of `.sigstore.json`.

See https://github.com/sigstore/sigstore-python/issues/814.

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -765,7 +765,7 @@ def _collect_verification_state(
         if cert is None:
             cert = file.parent / f"{file.name}.crt"
         if bundle is None:
-            bundle = file.parent / f"{file.name}.sigstore"
+            bundle = file.parent / f"{file.name}.sigstore.json"
 
         missing = []
         if args.signature or args.certificate:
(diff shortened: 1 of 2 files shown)


sigstore/_cli.py
```diff
@@ -775,11 +775,37 @@ def _collect_verification_state(
...
+                else:
+                    # NOTE: Logically `{input}.sigstore` is what's missing here, but
+                    # we only fail this check if the the `--bundle` or `{input}.sigstore.json`
+                    # forms are *also* missing. So we use the former in the missing set,
+                    # since it'll result in a more useful error message.
+                    missing.append(str(bundle))
+            else:
+                # Don't allow the user to implicitly verify `{input}.sigstore.json` if
+                # `{input}.sigstore` is also present, since this implies user confusion.
+                if bundle != legacy_default_bundle and legacy_default_bundle.is_file():
```
woodruffw, line 803: NB: We need this inequality check because a user can pass in `--bundle` with the same value as `legacy_default_bundle`, which is what the conformance suite does.

jku (commented):
This combines two items: 
* handling implicit vs explicit bundle and 
* handling legacy implicit bundle vs sigstore.json implicit bundle.

There's at least one issue left from that: if an explicit bundle is used, but the file does not exist then we use legacy implicit bundle. This seems wrong.

woodruffw:
> There's at least one issue left from that: if an explicit bundle is used, but the file does not exist then we use legacy implicit bundle. This seems wrong.

Good catch, thanks -- that does seem wrong. I'll fix that.

PR #824, opened 2023-12-05
_cli: use rich's logging handler

woodruffw:
This adds `rich` as a dependency, using its `RichHandler` logging handler to make our logging much prettier by default (when invoked from the CLI).

I'm not 100% sure this is worth it; opening for discussion/consideration 🙂 

Pros:

* Makes our logging much easier to read/visually distinguish. This is particularly helpful for cases where we use `logger.warning(...)` to emit deprecation notices, which otherwise could be missed in the monochromatic/unformatted output.

Cons:

* Heavyweight?

Demo:

[asciicast](https://asciinema.org/a/SLESuadskrCIMJMnZvLFOaI0N)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ dependencies = [
   "pyjwt >= 2.1",
   "pyOpenSSL >= 23.0.0",
   "requests",
+  "rich ~= 13.0",
   "securesystemslib",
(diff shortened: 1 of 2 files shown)


di (approved):
I don't hate it!

jku:
In hindsight, as a sigstore library consumer it's a bit annoying: I get multiple new dependencies in my lock file because of this

PR #832, opened 2023-12-13
Prep 2.1.0

tetsuo-cpp:


woodruffw (changes requested):
Sorry, I misspoke earlier -- mind prepping this as 2.1.0? Technically it includes a behavior addition (the CLI change in the log) so semver says we should bump minor instead of patch.

woodruffw (dismissed):
Thanks @tetsuo-cpp!

woodruffw (approved):

PR #839, opened 2023-12-15
Upgrade actions/upload-artifact & actions/download-artifact

jku:
This contains dependabot upgrade commits for both upload-artifact and download-artifact action: The assumption being tested here is that the new download-artifact requires new upload-artifact...


jku:
The assumption is false: closing this as this is not more useful than the actual dependabot PRs

PR #840, opened 2023-12-15
dependabot: Manage the internal action as well

jku:
Make the internal action in .github/actions/upload-coverage managed by dependabot. 

This (and the resulting dependabot PR) should unblock #835


woodruffw (approved):
LGTM, thanks.

These upload/download action changes are also causing pain for Python packaging, cf. https://github.com/actions/upload-artifact/issues/472

PR #843, opened 2023-12-18
Refactor upload-coverage action

jku:
#### Summary

Refactor upload-coverage to enable upgrading the upload-artifact and download-artifact actions (as the new upload-artifact does not allow uploading with the same name several times).
* Use unique artifact names, then download all artifacts before combining them for coverage
* Upgrade the actions

#### Release Note

N/A

### DIFF
--- a/.github/actions/upload-coverage/action.yml
+++ b/.github/actions/upload-coverage/action.yml
@@ -22,7 +22,7 @@ runs:
       shell: bash
     - uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
       with:
-        name: coverage-data
+        name: coverage-data-${{ steps.coverage-uuid.outputs.COVERAGE_UUID }}
         path: |
           .coverage.*
           *.lcov
--- a/.github/actions/upload-coverage/action.yml
+++ b/.github/actions/upload-coverage/action.yml
@@ -20,7 +20,7 @@ runs:
         fi
       id: coverage-uuid
       shell: bash
-    - uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+    - uses: actions/upload-artifact@c7d193f32edcb7bfad88892161225aeda64e9392 # v4.0.0
       with:
         name: coverage-data-${{ steps.coverage-uuid.outputs.COVERAGE_UUID }}
         path: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,12 +93,13 @@ jobs:
       - name: download coverage data
         uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
         with:
-          name: coverage-data
+          path: all-artifacts/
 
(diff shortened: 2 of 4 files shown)


jku:
> jku force-pushed the refactor-upload-coverage branch 10 times, ...

That took a few tries :joy: Looks fine now I think.

.github/actions/upload-coverage/action.yml
```diff
@@ -20,9 +20,9 @@ runs:
         fi
       id: coverage-uuid
       shell: bash
-    - uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+    - uses: actions/upload-artifact@c7d193f32edcb7bfad88892161225aeda64e9392 # v4.0.0
       with:
-        name: coverage-data
```
jku, line 25: The artifact name could be more useful but there is a FIXME about that 10 lines above already

.github/actions/upload-coverage/action.yml
```diff
@@ -20,9 +20,9 @@ runs:
...
+        name: coverage-data-${{ steps.coverage-uuid.outputs.COVERAGE_UUID }}
```
jku, line 25: The artifact name could be more useful but there is a FIXME about that 10 lines above already

.github/workflows/ci.yml
```diff
@@ -91,14 +91,15 @@ jobs:
...
         with:
-          name: coverage-data
+          path: all-artifacts/
 
       - name: combine coverage data
         id: combinecoverage
         run: |
           set +e
+          cp all-artifacts/coverage-data-*/.coverage.* .
           python -m coverage combine
```
woodruffw, line 103: This is a tiny thing, but I _think_ you can do this: 

```suggestion
          set +e
          python -m coverage combine all-coverage/
```

.github/workflows/ci.yml
```diff
@@ -91,14 +91,15 @@ jobs:
       - run: pip install coverage[toml]
 
       - name: download coverage data
-        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+        uses: actions/download-artifact@7a1cd3216ca9260cd8022db641d960b1db4d1be4 # v4.0.0
         with:
-          name: coverage-data
+          path: all-artifacts/
```
woodruffw, line 96: Nitpick, descriptive:

```suggestion
          path: all-coverage/
```

.github/workflows/ci.yml
jku: thanks, that's better. The argument needs to be `all-artifacts/coverage-data-*` though (download-artifacts creates extra directories when used like this).

.github/workflows/ci.yml
jku: I'm going to leave this unless you insist as the contents really are "all artifacts": if someone in the future adds another artifact upload earlier in the workflow, the unrelated artifact will show up in this directory. So any code that processes "all-artifacts" should be aware of that.

woodruffw: Sure, works for me.

woodruffw (approved):

PR #844, opened 2023-12-19
Separate TUF and trusted root management code

jku:
#### Summary

The purpose of this is to later enable both "--trust-root <FILE>" and some sort of "--offline" functionality.
 * Trust root can now be initialized from tuf, offline tuf or from a file
 * `_internal.tuf` module is now used only from the new `_internal.trustroot` module
 * all "trust root parsing" was moved from `_internal.tuf` to `_internal.trustroot`
 * Tests are modified to use the trustroot module now but they still (also) test the internal TUF implementation details
 * The new functionality (offline & from_file) is tested but is not exposed to UI

#### Some details

The main functional change as I see it is in TrustedRoot/TrustUpdater initialization. There are three possible scenarios (once this is exposed through UI):
* A trust root is provided as a file: TrustedRoot never instantiates a TrustUpdater
* offline is requested: A TrustUpdater is instantiated but it does no TUF operations (only returns trust root from cache)
* default: TrustUpdater is instantiated and it refreshes from repository immediately (since we know it will be used)

This leads to some minor test changes but nothing too bad.

#### Release Note

No release notes, I think: This should not change API or cli at this point

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -40,7 +40,7 @@ from sigstore._internal.rekor.client import (
     RekorClient,
     RekorKeyring,
 )
-from sigstore._internal.tuf import TrustUpdater
+from sigstore._internal.trustroot import CustomTrustedRoot
 from sigstore._utils import PEMCert
 from sigstore.errors import Error
 from sigstore.oidc import (
@@ -650,16 +650,16 @@ def _sign(args: argparse.Namespace) -> None:
     elif args.fulcio_url == DEFAULT_FULCIO_URL and args.rekor_url == DEFAULT_REKOR_URL:
         signing_ctx = SigningContext.production()
     else:
-        # Assume "production" keys if none are given as arguments
-        updater = TrustUpdater.production()
+        # Assume "production" trust root if no keys are given as arguments
+        trusted_root = CustomTrustedRoot.production()
         if args.ctfe_pem is not None:
             ctfe_keys = [args.ctfe_pem.read()]
         else:
-            ctfe_keys = updater.get_ctfe_keys()
+            ctfe_keys = trusted_root.get_ctfe_keys()
         if args.rekor_root_pubkey is not None:
             rekor_keys = [args.rekor_root_pubkey.read()]
         else:
-            rekor_keys = updater.get_rekor_keys()
+            rekor_keys = trusted_root.get_rekor_keys()
 
         ct_keyring = CTKeyring(Keyring(ctfe_keys))
         rekor_keyring = RekorKeyring(Keyring(rekor_keys))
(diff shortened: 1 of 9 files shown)


sigstore/_internal/trustroot.py
```diff
@@ -0,0 +1,148 @@
...
+    # valid.
+    if now < period.start:
+        return False
+
+    # If we want Expired keys, the key is valid at this point. Otherwise, check
+    # that we are within range.
+    return allow_expired or (period.end is None or now <= period.end)
+
+
+class CustomTrustedRoot(TrustedRoot):
```
woodruffw, line 59: My 0.02c: I'm a fan of the "import as private, export same name" idiom:

```python
from whereever import TrustedRoot as _TrustedRoot

class TrustedRoot(_TrustedRoot):
    ...
```

This avoids a naming decision and is still unambiguous, since Python types convey their namespace 🙂

woodruffw:
Reviewing now!

@jku: mind doing an update/rebase? I'd do it myself, but that then causes problems with the number of required approvers.

sigstore/_cli.py
```diff
@@ -650,16 +650,16 @@ def _sign(args: argparse.Namespace) -> None:
...
         if args.ctfe_pem is not None:
             ctfe_keys = [args.ctfe_pem.read()]
         else:
-            ctfe_keys = updater.get_ctfe_keys()
+            ctfe_keys = trusted_root.get_ctfe_keys()
         if args.rekor_root_pubkey is not None:
             rekor_keys = [args.rekor_root_pubkey.read()]
         else:
-            rekor_keys = updater.get_rekor_keys()
+            rekor_keys = trusted_root.get_rekor_keys()
```
jku, line 662: there's a slight change of functionality here (in contrast to what I said in the PR message):
this now leads to TUF repo getting  refreshed even if all keys are provided via command line... I'll think about how to avoid that

woodruffw: Making sure I understand: this causes the TUF repo to get refreshed because `TrustedRoot.production()` is doing the refresh, rather than deferring it to one of the `get_*_keys()` calls, right?

jku: Yes. In the future (if the local keys can only be provided via --trust-root) this would not be an issue but in this transition state it could be

woodruffw: Makes sense, thanks. We can hold off if you feel strongly here, but I'm personally okay with this -- we intend to remove these flags anyways, so IMHO we can deal with a temporary unnecessary network request.

woodruffw (approved):
LGTM! 

@jku feel free to merge unless you'd like to address https://github.com/sigstore/sigstore-python/pull/844#discussion_r1440622945 in this PR; I'm fine either way 🙂

sigstore/_cli.py
jku: Let's leave it like this then and move on with next steps

PR #855, opened 2024-01-02
dependabot: group GHA updates

woodruffw:
This will cut our automated PR volume down slightly: GHA updates from Dependabot will now be batched into one big PR, rather than one PR per updated action.


jku (approved):
I believe this leads to two groups both named "actions" but I don't see that as being a problem -- "policy" for both groups should be the same

Issue #854, opened 2024-01-02
API: Remove `SigningResult`


woodruffw:
Per discussion in #804: `SigningResult` is now mostly an internal duplication of a `Bundle`, so we should just use `Bundle` directly.

This will be a breaking API change, so slating for 3.0.

tetsuo-cpp:
@woodruffw I can do this just before we cut 3.0. I assume we wouldn't want this in immediately since it would introduce breaking changes to any point releases that we need to make between now and then.

woodruffw:
Makes sense to me!

woodruffw:
#860 is making this a little more pressing, so I'll do this today.

PR #860, opened 2024-01-09
feat: Enable signing with pre-computed hash

laurentsimon:
closes https://github.com/sigstore/sigstore-python/issues/666

Let me know if you'd like me to add a dedicated unit test for this new option.

Context: I'm working on adding a new hash type to rekor to support hashing directories. I will need to follow up to add support for it in https://github.com/trailofbits/sigstore-rekor-types. 

Test example with pre-computed sha256:
```python
from sigstore import hashes as sigstore_hashes
from sigstore_protobuf_specs.dev.sigstore.common.v1 import HashAlgorithm

with file.open(mode="rb", buffering=0) as io:
    digest = hashlib.sha256(io.read())
    hashed = sigstore_hashes.Hashed(digest=digest, algorithm=HashAlgorithm.SHA2_256)
    result = signer.sign(hashed)
```
#### Summary

This PR enables passing a pre-hashed value instead of the content to be hashed.

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -686,21 +686,9 @@ def _sign(args: argparse.Namespace) -> None:
     with signing_ctx.signer(identity) as signer:
         for file, outputs in output_map.items():
             logger.debug(f"signing for {file.name}")
-            with file.open(mode="rb", buffering=0) as iof:
+            with file.open(mode="rb", buffering=0) as io:
                 try:
-                    import hashlib
-                    import io
-                    from sigstore import hashes as sigstore_hashes
-                    from sigstore_protobuf_specs.dev.sigstore.common.v1 import HashAlgorithm
-
-                    hash = hashlib.sha256()
-                    hash.update(iof.read())
-                    digest = hash.digest()
-                    print(digest.hex())
-                    hashed = sigstore_hashes.Hashed(digest=digest, algorithm=HashAlgorithm.SHA2_256)
-                    result = signer.sign(hashed)
-
-                    #result = signer.sign(input_=io)
+                    result = signer.sign(input_=io)
                 except ExpiredIdentity as exp_identity:
                     print("Signature failed: identity token has expired")
                     raise exp_identity
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -686,7 +686,7 @@ def _sign(args: argparse.Namespace) -> None:
     with signing_ctx.signer(identity) as signer:
         for file, outputs in output_map.items():
             logger.debug(f"signing for {file.name}")
-            with file.open(mode="rb", buffering=0) as io:
+            with file.open(mode="rb", buffering=0) as fio:
                 try:
                     result = signer.sign(input_=io)
                 except ExpiredIdentity as exp_identity:
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -686,9 +686,21 @@ def _sign(args: argparse.Namespace) -> None:
     with signing_ctx.signer(identity) as signer:
         for file, outputs in output_map.items():
             logger.debug(f"signing for {file.name}")
-            with file.open(mode="rb", buffering=0) as io:
+            with file.open(mode="rb", buffering=0) as iof:
                 try:
-                    result = signer.sign(input_=io)
+                    import hashlib
+                    import io
+                    from sigstore import hashes as sigstore_hashes
+                    from sigstore_protobuf_specs.dev.sigstore.common.v1 import HashAlgorithm
+
+                    hash = hashlib.sha256()
+                    hash.update(iof.read())
+                    digest = hash.digest()
+                    print(digest.hex())
+                    hashed = sigstore_hashes.Hashed(digest=digest, algorithm=HashAlgorithm.SHA2_256)
+                    result = signer.sign(hashed)
+
+                    #result = signer.sign(input_=io)
                 except ExpiredIdentity as exp_identity:
                     print("Signature failed: identity token has expired")
                     raise exp_identity
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -688,7 +688,7 @@ def _sign(args: argparse.Namespace) -> None:
             logger.debug(f"signing for {file.name}")
             with file.open(mode="rb", buffering=0) as io:
                 try:
-                    result = signer.sign(input_=iof)
+                    result = signer.sign(input_=io)
                 except ExpiredIdentity as exp_identity:
                     print("Signature failed: identity token has expired")
                     raise exp_identity
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -688,7 +688,21 @@ def _sign(args: argparse.Namespace) -> None:
             logger.debug(f"signing for {file.name}")
             with file.open(mode="rb", buffering=0) as io:
                 try:
-                    result = signer.sign(input_=io)
+                    # NOTE: This is where signin is performed.
+                    import hashlib, io
+                    from cryptography.hazmat.primitives import hashes
+                    from cryptography.hazmat.primitives.asymmetric.utils import Prehashed
+                    class DIRHASH_SHARD1G_SHA256(hashes.HashAlgorithm):
+                        name = "dirhash_shard1G_sha256"
+                        digest_size = 32
+                        block_size = -1
+
+                    hash = hashlib.sha256()
+                    hash.update(iof.read())
+                    content = hash.digest()
+                    contentio = io.BytesIO(content)
+                    result = signer.sign(contentio, Prehashed(DIRHASH_SHARD1G_SHA256()))
+                    #result = signer.sign(input_=iof)
                 except ExpiredIdentity as exp_identity:
                     print("Signature failed: identity token has expired")
                     raise exp_identity
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -686,7 +686,7 @@ def _sign(args: argparse.Namespace) -> None:
     with signing_ctx.signer(identity) as signer:
         for file, outputs in output_map.items():
             logger.debug(f"signing for {file.name}")
-            with file.open(mode="rb", buffering=0) as fio:
+            with file.open(mode="rb", buffering=0) as io:
                 try:
                     result = signer.sign(input_=io)
                 except ExpiredIdentity as exp_identity:
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -688,21 +688,7 @@ def _sign(args: argparse.Namespace) -> None:
             logger.debug(f"signing for {file.name}")
             with file.open(mode="rb", buffering=0) as io:
                 try:
-                    # NOTE: This is where signin is performed.
-                    import hashlib, io
-                    from cryptography.hazmat.primitives import hashes
-                    from cryptography.hazmat.primitives.asymmetric.utils import Prehashed
-                    class DIRHASH_SHARD1G_SHA256(hashes.HashAlgorithm):
-                        name = "dirhash_shard1G_sha256"
-                        digest_size = 32
-                        block_size = -1
-
-                    hash = hashlib.sha256()
-                    hash.update(iof.read())
-                    content = hash.digest()
-                    contentio = io.BytesIO(content)
-                    result = signer.sign(contentio, Prehashed(DIRHASH_SHARD1G_SHA256()))
-                    #result = signer.sign(input_=iof)
+                    result = signer.sign(input_=iof)
                 except ExpiredIdentity as exp_identity:
                     print("Signature failed: identity token has expired")
                     raise exp_identity
--- a/sigstore/_utils.py
+++ b/sigstore/_utils.py
@@ -23,8 +23,9 @@ import hashlib
 import sys
 from typing import IO, NewType, Union
 
-from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import ec, rsa
+<<<<<<< HEAD
 from cryptography.hazmat.primitives.asymmetric.utils import Prehashed
 from cryptography.x509 import (
     Certificate,
@@ -32,9 +33,13 @@ from cryptography.x509 import (
     Version,
     load_der_x509_certificate,
 )
+=======
+from cryptography.x509 import Certificate, ExtensionNotFound, Version
+>>>>>>> f67b019 (Change API)
 from cryptography.x509.oid import ExtendedKeyUsageOID, ExtensionOID
 from sigstore_protobuf_specs.dev.sigstore.common.v1 import HashAlgorithm
 
+from sigstore import hashes as sigstore_hashes
 from sigstore.errors import Error
 
 if sys.version_info < (3, 11):
@@ -159,26 +164,13 @@ def key_id(key: PublicKey) -> KeyID:
 
     return KeyID(hashlib.sha256(public_bytes).digest())
 
-def hazmat_digest_to_bundle(algo: str) -> HashAlgorithm:
-    lookup = {hashes.SHA256().name: HashAlgorithm.SHA2_256}
-    if algo in lookup:
-        return HashAlgorithm(lookup[algo])
-    return ValueError(f"unknown digest algorithm {algo}")
+def get_digest(input_: IO[bytes] | sigstore_hashes.Hashed) -> sigstore_hashes.Hashed:
+    if isinstance(input_, sigstore_hashes.Hashed):
+        return input_
 
-def get_digest(
-        input_: IO[bytes],
-        algorithm_: Prehashed = None,
-    ) -> (bytes, Prehashed):
-    if algorithm_ is None:
-        return sha256_streaming(input_), Prehashed(hashes.SHA256())
-
-    if isinstance(algorithm_, Prehashed):
-        # Check we have a 256-bit digest size for compatibility with secp256r1.
-        if algorithm_.digest_size != 32:
-            return ValueError(f"invalid digest size ({algorithm_.digest_size}), expected 32")
-        return input_.getvalue(), algorithm_
-
-    raise ValueError("invalid arguments")
+    # NOTE: Not able to check for the type `TypeError: Subscripted generics cannot be used with class and instance checks`
+    # when calling isinstance(_input, IO[bytes])
+    return sigstore_hashes.Hashed(digest=sha256_streaming(input_), algorithm=HashAlgorithm.SHA2_256)
 
 def sha256_streaming(io: IO[bytes]) -> bytes:
     """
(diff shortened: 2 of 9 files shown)


sigstore/sign.py
```diff
@@ -172,12 +172,26 @@ def _signing_cert(
 
             return certificate_response
 
+    # https://github.com/sigstore/rekor/issues/1299
```
laurentsimon, line 175: ignore those links

sigstore/sign.py
```diff
@@ -172,12 +172,26 @@ def _signing_cert(
...
+        """Public API for signing blobs and pre-hashed content"""
+        input_digest, algorithm = get_digest(input_, algorithm_)
+        return self._sign_digest(input_digest, algorithm)
+        
+    def _sign_digest(
+        self,
+        input_digest_: bytes,
+        algorithm_: str,
+    ) -> SigningResult:
+        print(f"algo: {algorithm_}")
```
laurentsimon, line 193: ignore the print

sigstore/sign.py
```diff
@@ -172,12 +172,26 @@ def _signing_cert(
...
-        """Public API for signing blobs"""
-        input_digest = sha256_streaming(input_)
+        """Public API for signing blobs and pre-hashed content"""
+        input_digest, algorithm = get_digest(input_, algorithm_)
+        return self._sign_digest(input_digest, algorithm)
+        
+    def _sign_digest(
+        self,
+        input_digest_: bytes,
+        algorithm_: str,
```
jku, line 191: I assume you mean Prehashed here?

sigstore/sign.py
```diff
@@ -387,7 +411,8 @@ def to_bundle(self) -> Bundle:
             verification_material=material,
             message_signature=MessageSignature(
                 message_digest=HashOutput(
-                    algorithm=HashAlgorithm.SHA2_256,
+                    #algorithm=HashAlgorithm.SHA2_256,
+                    algorithm=self.digest_algorithm,
```
jku, line 415: Can you talk me through this (possibly I just don't understand the protobuf details):
* we're creating protobuf message HashOutput where algorithm is a string (that comes from cryptography?)
* but the protobuf HashAlgorithm is actually an enum 
* HashAlgorithm has a single documented value (SHA2_256)

laurentsimon: HashAlgorithm comes from Sigstore `sigstore_protobuf_specs.dev.sigstore.common.v1` which comes from `https://github.com/sigstore/protobuf-specs/blob/d96fa4407cb4489233f133ab036c0ed98c2756a4/protos/sigstore_common.proto#L39`. The value is a string, ifaict.

sigstore/sign.py
```diff
@@ -175,9 +174,17 @@ def _signing_cert(
     def sign(
         self,
         input_: IO[bytes],
```
laurentsimon, line 176: Note: I considered using `input_ IO[bytes] | bytes` where `bytes` type would be used with `Prehashed` but I think it further complicates the API. So I've kept `IO[bytes]` instead. Please advise.

laurentsimon:
tests seem to be failing because the bundle uses a different name for the hash algo (`SHA2_256`) instead of `SHA256` for the standard crypto lib. I'll fix that while you review the PR.

sigstore/sign.py
jku: I don't think it works that way. The python bindings might not complain when that r.uns it but the protobuf serialization likely won't work.

But I suppose if you were asking for high level feedback, the real questions related to this are: 
* how is this going to work in bundle that is an interop format? 
* should the bundle even try to document the used algorithm (since the verifier I guess is anyway expected to know the magic precalculation method that is required to actually verify) -- does the protobuf need a "custom" algorithm value?

jku:
If you run `make reformat` you'll start to get useful lint results as well

sigstore/_utils.py
```diff
@@ -139,6 +141,26 @@ def key_id(key: PublicKey) -> KeyID:
 
     return KeyID(hashlib.sha256(public_bytes).digest())
 
+def hazmat_digest_to_bundle(algo: str) -> HashAlgorithm:
```
laurentsimon, line 144: naive attempt to fix the conformance tests without running them locally. Still complaining that the enum number is invalid

sigstore/sign.py
laurentsimon: re: yeah you're right about types I double checked. I replied to your question about bundle in the other comment. TL;DR: We're adding a hash in https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L39 which will rekor will support for hashedrekord. This PR _only_ makes it possible to pre-compute _a_ hash supported by rekor. Right it only supports sha256 because rekor only supports that hash.

Let me know if I'm misunderstanding something. Thanks

laurentsimon: Please also see https://github.com/trailofbits/sigstore-rekor-types/issues/22

laurentsimon:
> If you run `make reformat` you'll start to get useful lint results as well

I've run these commands and it works locally. However in CI it's failing. Any ideas?

sigstore/verify/models.py
```diff
@@ -179,9 +179,9 @@ class VerificationMaterials:
     Represents the materials needed to perform a Sigstore verification.
     """
 
-    input_digest: bytes
+    hashed_input: sigstore_hashes.Hashed
```
laurentsimon, line 182: This seems like a breaking change :(

laurentsimon:
I've made the changes, PTAL

sigstore/verify/models.py
woodruffw: Yep, it is. That's fine IMO since we're on a breakage path for 3.0 anyways, and #804 will also break compatibility here.

sigstore/_utils.py
```diff
@@ -139,6 +141,13 @@ def key_id(key: PublicKey) -> KeyID:
 
     return KeyID(hashlib.sha256(public_bytes).digest())
 
+def get_digest(input_: IO[bytes] | sigstore_hashes.Hashed) -> sigstore_hashes.Hashed:
+    if isinstance(input_, sigstore_hashes.Hashed):
+        return input_
+
+    # NOTE: Not able to check for the type `TypeError: Subscripted generics cannot be used with class and instance checks`
+    # when calling isinstance(_input, IO[bytes])
```
woodruffw, line 149: Extreme nitpick, but IMO we can drop this `NOTE`: the `sha256_streaming` call's parameter hint will keep this checked for us.

woodruffw:
> I've run these commands and it works locally. However in CI it's failing. Any ideas?

Hmm, this might be an outdated local toolchain. What does `ruff --version` say?

sigstore/hashes.py
woodruffw, line 1: Thinking out loud: this is a pretty small API surface for its own module. Rather than committing to a new API path like this (`sigstore.hashes`), maybe we could do one of the following:

* Turn this into `sigstore._hashes`, and re-export `Hashed` from `sigstore.sign`
* Remove this entirely, and inline it into `sigstore.sign`

Thoughts on either of these?

sigstore/sign.py
```diff
@@ -313,9 +314,14 @@ class SigningResult(BaseModel):
...
+    digest_algorithm: HashAlgorithm
+    """
+    The digest algorithm to use for the hash.
+    """
+
     input_digest: HexStr
     """
-    The hex-encoded SHA256 digest of the input that was signed for.
+    The hex-encoded digest of the input that was signed for.
     """
```
woodruffw, line 325: NB: Both of these fields will disappear with #804 and we intend to remove `SigningResult` entirely with #854, so we should probably avoid more fields here. Given that `HashAlgorithm` will currently always be invariant, maybe we can just hardcode it with a `NOTE` for now?

sigstore/hashes.py
laurentsimon: does it matter that that the `verify/*` will be needing it? If not, I'm fine with any of the above.

sigstore/sign.py
laurentsimon: as soon as we update rekor (probably next week), it will no longer be an invariant, ie there wil be 2 supported hashes (sha256 and some dirhash hash)

woodruffw: Gotcha -- in that case, we should probably bump up the priority on #854, since `SigningResult` has outlived its utility in the world of Sigstore bundles 🙂

woodruffw:
> I updated to ruff `0.1.13`. I managed to locally fix the files. CI still complaining.

Gotcha, sorry for the mess here -- I'll do some debugging on the CI today. It looks like `lint` is still failing, although this time because we have public APIs without docstrings.

sigstore/sign.py
laurentsimon: SGTM. So you'd land https://github.com/sigstore/sigstore-python/issues/854 then this PR I suppose? Or you don't really care?

woodruffw: Tangent, but is dirhash documented anywhere? I see there's a dirhash standard on GitHub but I wasn't sure if it was the same thing.

(Part of what I'm wondering is what the mapping between a dirhash and a standard algorithm is -- how do we form a `Prehashed` for them?)

sigstore/hashes.py
woodruffw: Ah yeah, good point. Given that, I think this is fine as its own module.

laurentsimon: Alright. Keeping as-is then. Thanks

laurentsimon:
@woodruffw out of curiosity, what is the timeline for v3 release https://github.com/sigstore/sigstore-python/milestone/5?

woodruffw:
> @woodruffw out of curiosity, what is the timeline for v3 release https://github.com/sigstore/sigstore-python/milestone/5?

I don't think we have a fixed timeline yet, it's mostly just a matter of when DSSE support and this get merged in, along with a few CLI deprecations. 

(To speed things along, we could definitely do one or more v3 prereleases, if you're looking to consume the APIs downstream.)

sigstore/sign.py
woodruffw: That matches my understanding, thanks! We should probably make sure that subclassing `HashAlgorithm` like that for external hashes is okay in terms of PyCA's APIs, but otherwise that seems reasonable to me.

laurentsimon:
> > @woodruffw out of curiosity, what is the timeline for v3 release https://github.com/sigstore/sigstore-python/milestone/5?
> 
> I don't think we have a fixed timeline yet, it's mostly just a matter of when DSSE support and this get merged in, along with a few CLI deprecations.
> 
> (To speed things along, we could definitely do one or more v3 prereleases, if you're looking to consume the APIs downstream.)

That'd be nice yes! I'd be interested in a prelrelease with this PR merged in, so we can start testing e2e on our side. What would be a timeline for it?

woodruffw:
> That'd be nice yes! I'd be interested in a prelrelease with this PR merged in, so we can start testing e2e on our side. What would be a timeline for it?

We could make that pre-release as soon as this gets merged in 🙂 

(We won't make any semver guarantees around pre-releases, but I think that's already given.)

laurentsimon:
> > That'd be nice yes! I'd be interested in a prelrelease with this PR merged in, so we can start testing e2e on our side. What would be a timeline for it?
> 
> We could make that pre-release as soon as this gets merged in 🙂
> 
> (We won't make any semver guarantees around pre-releases, but I think that's already given.)

I'd probably just ask that the rekor changes are in too. But I'll let you know once this is done.

laurentsimon:
I'll wait for https://github.com/sigstore/sigstore-python/pull/862 to land and update this PR. Anything else I should address in the PR?

woodruffw:
> I'll wait for #862 to land and update this PR. Anything else I should address in the PR?

Nope, I think you're good. Thanks a ton for your work here!

jku:
Looks good (didn't try it yet though). 

I just have the one comment, same as before: using cryptography to lookup algorithm names for some other system smells a bit.

This is about hazmat_algorithm():
* the method is only used outside of the class to get the algorithm name (`as_prehashed()` could easily avoid using it)
* the algo name is supposed to be compatible with _other sigstore components, not cryptography_: we can't expect the hazmat algorithms to use the same strings that sigstore uses. 

I'm not 100% sure if this is complete but... does this do what you need: 
```
def as_prehashed(self) -> Prehashed:
    if self.algorithm == HashAlgorithm.SHA2_256:
        return Prehashed(hashes.SHA256())
    raise ...

def as_hashedrekord_algorithm() -> hashedrekord.Algorithm
    if self.algorithm == HashAlgorithm.SHA2_256:
            return hashedrekord.Algorithm.SHA256
    raise ...
```

woodruffw:
#862 is landed, so this needs a deconflict. I'm also planning to merge #804 soon, so you _may_ want to wait until that's landed before deconflicting (since it touches many of the same lines).

laurentsimon:
> #862 is landed, so this needs a deconflict. I'm also planning to merge #804 soon, so you _may_ want to wait until that's landed before deconflicting (since it touches many of the same lines).

OK I'll wait for #804 to land then.

jku:
#804 is in main now

laurentsimon:
I've rebased. However I was not able to tests locally, keeps getting this error:

`spec=rekor_types.hashedrekord.HashedrekordV001Schema(
         ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'rekor_types' has no attribute 'hashedrekord'. Did you mean: 'Hashedrekord'?`

jku:
> How do you typically test?

the makefile tries to manage a virtual env for testing:
* this gets used by `make test` or can be generated by `make dev`
* once the env exists exists you can enter it for manual testing with `source env/bin/activate`

FYI, the dependency version pinning is a bit strange:
* pyproject.toml documents the actual **required** runtime versions 
* install/requirements.txt documents pinned versions -- but this is **only** updated at release time

For sigstore-rekor-types: you want 0.0.12 as documented by pyproject.toml. The virtual environment should already have this version.

laurentsimon:
> > How do you typically test?
> 
> the makefile tries to manage a virtual env for testing:
> 
> * this gets used by `make test` or can be manually generated by `make dev`
> * once the env exists exists you can enter it for manual testing with `source env/bin/activate`
> 
> FYI, the dependency version pinning is a bit strange:
> 
> * pyproject.toml documents the actual **required** runtime versions
> * install/requirements.txt documents pinned versions -- but this is **only** updated at release time (I think this is a bit confusing and I'm not sure what the purpose is)
> 
> For sigstore-rekor-types: you want 0.0.12 as documented by pyproject.toml. The virtual environment should already have this version.

Thank you, that was super helpful. I was able to make the changes and the conformance tests are now passing. PTAL.

woodruffw (commented):
This looks great to me @laurentsimon! One last question, and then I think we're good to merge here.

sigstore/hashes.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def as_hashedrekord_algorithm(self) -> rekor_types.hashedrekord.Algorithm:
+        if self.algorithm == HashAlgorithm.SHA2_256:
+            return rekor_types.hashedrekord.Algorithm.SHA256
+        raise ValueError(f"unknown hash algorithm: {self.algorithm}")
+
+    def as_prehashed(self) -> Prehashed:
+        if self.algorithm == HashAlgorithm.SHA2_256:
+            return Prehashed(hashes.SHA256())
+        raise ValueError(f"unknown hash algorithm: {self.algorithm}")
```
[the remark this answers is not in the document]
laurentsimon: +1, they do not. Let's move them to private. I meant to ask and forgot. Alternatively I make the entire file private `_hash.py`. What do you prefer?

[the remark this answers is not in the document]
woodruffw: Given that `Hashed` will be in the publish signing API, I think we can leave `hashes.py` public!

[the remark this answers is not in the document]
laurentsimon: > Given that Hashed will be in the public signing API, I think we can leave hashes.py public!

what was I thinking... changes made. Thanks

sigstore/_cli.py
```diff
@@ -686,9 +686,21 @@ def _sign(args: argparse.Namespace) -> None:
...
         for file, outputs in output_map.items():
             logger.debug(f"signing for {file.name}")
-            with file.open(mode="rb", buffering=0) as io:
+            with file.open(mode="rb", buffering=0) as iof:
                 try:
-                    result = signer.sign(input_=io)
+                    import hashlib
+                    import io
+                    from sigstore import hashes as sigstore_hashes
+                    from sigstore_protobuf_specs.dev.sigstore.common.v1 import HashAlgorithm
```
woodruffw, line 694: Can we move these imports to the top level? Inline imports like this tend to be a lot harder to test/easy to introduce circularities with 🙂

sigstore/_cli.py
```diff
@@ -686,9 +686,21 @@ def _sign(args: argparse.Namespace) -> None:
...
+
+                    hash = hashlib.sha256()
+                    hash.update(iof.read())
+                    digest = hash.digest()
+                    print(digest.hex())
+                    hashed = sigstore_hashes.Hashed(digest=digest, algorithm=HashAlgorithm.SHA2_256)
+                    result = signer.sign(hashed)
+
+                    #result = signer.sign(input_=io)
```
woodruffw, line 703: Actually, looking at this more, why are we doing the pre-hash here? I think `sign(...)` should probably continue to take `IO[bytes]` in this case, right?

sigstore/_cli.py
laurentsimon: accidental commit that was only used for testing. Removed.

woodruffw:
LGTM, although lint is still failing. @laurentsimon are you able to diagnose that? Otherwise I can try and do it, if you give me push access to your branch.

laurentsimon:
> LGTM, although lint is still failing. @laurentsimon are you able to diagnose that? Otherwise I can try and do it, if you give me push access to your branch.

linter passes locally, so I think you're in a better position to debug that part :)
You have write access to the PR branch, I think

woodruffw:
Fixed up the CI (sorry for the pain here, I need to figure out what's going on).

This LGTM, but @jku or another maintainer will need to approve now that I'm the last pusher. Given that, I'll spend some time writing some tests here.

test/unit/test_sign.py
```diff
@@ -116,3 +122,27 @@ def test_identity_proof_claim_lookup(id_config, monkeypatch):
...
+        bundle = signer.sign(hashed)
+
+    assert bundle.message_signature.message_digest.algorithm == hashed.algorithm
+    assert bundle.message_signature.message_digest.digest == hashed.digest
+
+    input_.seek(0)
+    materials = VerificationMaterials.from_bundle(
+        input_=input_, bundle=bundle, offline=False
+    )
+    verifier.verify(materials=materials, policy=UnsafeNoOp())
```
woodruffw, line 148: I've added this to confirm that the round-trip works as expected: we sign over a `Hashed(...)` and then verify over the pre-image.

woodruffw (approved):

jku (approved):
Looks good!

woodruffw:
Thanks a ton for your hard work here @laurentsimon!

PR #862, opened 2024-01-12
API: remove SigningResult

### LINKED ISSUE #854: API: Remove `SigningResult`

woodruffw:
This removes the `SigningResult` model, and replaces it with `Bundle` in APIs where a `SigningResult` was previously returned.

Closes #854.

Unblocks #860.

cc @laurentsimon for visibility 🙂

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -41,7 +41,7 @@ from sigstore._internal.rekor.client import (
     RekorKeyring,
 )
 from sigstore._internal.trustroot import TrustedRoot
-from sigstore._utils import PEMCert
+from sigstore._utils import PEMCert, cert_der_to_pem
 from sigstore.errors import Error
 from sigstore.oidc import (
     DEFAULT_OAUTH_ISSUER_URL,
@@ -698,10 +698,12 @@ def _sign(args: argparse.Namespace) -> None:
                     raise exp_certificate
 
             print("Using ephemeral certificate:")
-            print(result.cert_pem)
+            cert = result.verification_material.x509_certificate_chain.certificates[0]
+            cert_pem = cert_der_to_pem(cert.raw_bytes)
+            print(cert_pem)
 
             print(
-                f"Transparency log entry created at index: {result.log_entry.log_index}"
+                f"Transparency log entry created at index: {result.verification_material.tlog_entries[0].log_index}"
             )
 
             sig_output: TextIO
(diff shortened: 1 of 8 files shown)


sigstore/_cli.py
```diff
@@ -698,10 +698,12 @@
                     raise exp_certificate
 
             print("Using ephemeral certificate:")
-            print(result.cert_pem)
+            cert = result.verification_material.x509_certificate_chain.certificates[0]
+            cert_pem = cert_der_to_pem(cert.raw_bytes)
+            print(cert_pem)
```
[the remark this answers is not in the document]
woodruffw: [wet raspberry noise]

woodruffw:
This should be good to go -- the `Bundle` attributes are slightly different than the old `SigningResult` ones, but just about everything relevant is still present (the log entry UUIDs are gone, but these were only part of the Rekor REST API anyways).

woodruffw:
(More generally, this makes me wonder if we can remove `LogEntry` too and just use the protobuf-specs `TransparencyLogEntry`. But that might be harder of a lift.)

sigstore/_utils.py
```diff
@@ -126,6 +131,19 @@ def base64_encode_pem_cert(cert: Certificate) -> B64Str:
...
 
 
+def cert_der_to_pem(der: bytes) -> str:
+    """
+    Converts a DER-encoded X.509 certificate into its PEM encoding.
+
+    Returns a string containing a PEM-encoded X.509 certificate.
+    """
+
+    # NOTE: Technically we don't have to round-trip like this, since
```
tetsuo-cpp, line 141: I don't understand what this comment is saying.

tetsuo-cpp (approved):
LGTM!

sigstore/_utils.py
woodruffw: I can probably just remove it: what I was trying to say was that PEM format is just `pem_headers(base64(der))`, so we _technically_ don't need to do a round-trip through the X.509 parser like this implementation is doing. However, doing the round-trip is simpler, since there's no dedicated DER-to-PEM transform API.

tetsuo-cpp: Ah, gotcha. That's fine.

PR #876, opened 2024-01-24
_cli: remove non-bundle inputs

woodruffw:
Very WIP. This is the first step towards being able to remove `VerificationMaterials`, which in turn will allow us to plumb `Bundle` through as our top-level type instead.

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -41,7 +41,7 @@ from sigstore._internal.rekor.client import (
     RekorKeyring,
 )
 from sigstore._internal.trustroot import TrustedRoot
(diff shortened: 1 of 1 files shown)


woodruffw:
https://github.com/sigstore/sigstore-conformance/issues/122 will ultimately determine how we proceed here.

woodruffw:
Putting this on hold for now: the right approach for the time being is to continue to support these inputs, but to internally transform them into the appropriate `Bundle` type.

woodruffw:
This is stale and will conflict with a related refactor in #937, so closing.

PR #877, opened 2024-01-25
verify: Remove a workaround for pre-3.8 Python

jku:
After this we have no uses of typing-extensions. It was not actually listed as a (direct) dependency at all.


woodruffw (approved):
Thanks @jku!

PR #884, opened 2024-02-02
Fix cryptography deprecations

### LINKED ISSUE #878: Use of deprecated cryptography properties
```
CryptographyDeprecationWarning: Properties that return a naïve datetime 
object have been deprecated. Please switch to not_valid_before_utc.
  sign_date = materials.certificate.not_valid_before
```
This and similar warnings for `not_valid_after` appear from both `sigstore/sign.py` and `sigstore/verify/verifier.py`.

I believe the suggested fixes should work fine.

jku:
`Certificate.not_valid_before` and `Certificate.not_valid_after` have been deprecated in cryptography 42.

* Require cryptography >= 42 (because the replacement fields are only available there)
* Use the new fields in `verify` and `sign`

This seems straight forward to me but review from someone certificate savvy wouldn't hurt.  The only interesting bit to me seems to be that we now give X509Store  a datetime that is explicitly UTC -- but I don't see that changing anything.

Fixes #878

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 ]
 dependencies = [
   "appdirs ~= 1.4",
-  "cryptography >= 39",
+  "cryptography >= 42",
   "id >= 1.1.0",
   "importlib_resources ~= 5.7; python_version < '3.11'",
   "in-toto-attestation == 0.9.3",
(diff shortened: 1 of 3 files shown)


woodruffw (approved):
Thanks @jku!

PR #893, opened 2024-02-04
Merge CLs from 2.1.x series

woodruffw:
This cherry-picks the release prep work from `series/2.1.x` back onto `main`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - series/*
   pull_request:
   schedule:
     - cron: '0 12 * * *'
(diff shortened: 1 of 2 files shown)


jku:
Am I correct that the release pipeline just allowed you to make releases to pypi from unprotected branches without review  but branch protection now requires someone review the changes that have been already released to pypi? :thinking: 

That seems incorrect.

woodruffw:
> Am I correct that the release pipeline just allowed you to make releases to pypi from unprotected branches without review but branch protection now requires someone review the changes that have been already released to pypi? 🤔

Yes. This seems completely backwards to me, and I'm not sure why it let me do the release unilaterally (there should at minimum have been a workflow signoff for release, but it didn't prompt me for that either).

I'm going to look at strengthening the branch rules here, and there may be some release protections we can add as well.

woodruffw:
Hmm, I don't have the ability to add additional restrictions here (I'm not a Sigstore org admin or a `sigstore-python` repo admin). I think this will probably need some Pulumi updates. I can look into that on Monday.

jku (approved):

PR #901, opened 2024-02-12
sigstore: v3 bundles

woodruffw:
Bumps `sigstore-protobuf-specs`, giving us access to the latest Bundle version.

### DIFF
--- a/sigstore/verify/models.py
+++ b/sigstore/verify/models.py
@@ -57,7 +57,7 @@ from sigstore import hashes as sigstore_hashes
 from sigstore._internal.rekor import RekorClient
 from sigstore._utils import (
     B64Str,
-    KnownBundleType,
+    BundleType,
     PEMCert,
     base64_encode_pem_cert,
     cert_is_leaf,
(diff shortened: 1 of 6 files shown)


woodruffw:
I think this is good to go. It adds support for verifying v3 bundles, but still emits v2 ones. 

In terms of switching signing to v3: I think the 3.x release series would be reasonable.

sigstore/verify/models.py
```diff
@@ -317,22 +318,22 @@ def from_bundle(
...
-        if bundle.media_type == _BUNDLE_0_1:
+        if media_type == KnownBundleType.BUNDLE_0_1:
             if not inclusion_promise:
                 raise InvalidMaterials("bundle must contain an inclusion promise")
             if inclusion_proof and not inclusion_proof.checkpoint.envelope:
                 logger.debug(
                     "0.1 bundle contains inclusion proof without checkpoint; ignoring"
                 )
-        elif bundle.media_type == _BUNDLE_0_2:
+        elif media_type == KnownBundleType.BUNDLE_0_2:
```
jku, line 336: what about `BUNDLE_0_3` here?

woodruffw: Whoops, thanks.

jku (approved):
Seems quite reasonable as is. Some comments : 
* tests would be nice but I can see how they become more viable once the signing works too
* `KnownBundleType` is a bit of a mouthful: it could probably be just `BundleType`

woodruffw:
> * `KnownBundleType` is a bit of a mouthful: it could probably be just `BundleType`

Yeah, I like that 🙂 -- I'll make a quick change (and also update the CHANGELOG).

jku (approved):

PR #904, opened 2024-02-14
sigstore: prep verify APIs for DSSE

woodruffw:
~~WIP.~~

This is the first step towards DSSE verification support, breaking apart some of our internal design mistakes around the state carried in `VerificationMaterials`. Merging this will allow us to be more flexible about the inputs passed to `verify()`, including DSSE statements and prehashed inputs as `Hashed` objects.


sigstore/verify/verifier.py
```diff
@@ -143,6 +143,7 @@ def staging(cls) -> Verifier:
 
     def verify(
         self,
+        input_: IO[bytes],
         materials: VerificationMaterials,
         policy: VerificationPolicy,
     ) -> VerificationResult:
```
woodruffw, line 149: @jku @laurentsimon do either of you have strong opinions about the input type here? I made it `IO[bytes]` to mirror our current `sign(...)` API, but (IMO) using IO types has been a weakness for us (it pushes effects into the API here, makes the API stateful, etc.).

As an alternative, I'm thinking of making this just take a `bytes` instead -- I think that'll be perfectly fine for ordinary input sizes, and users with huge inputs can use the `Hashed` interface (once I add it here).

woodruffw: I've changed this to take `bytes` instead, per thoughts above. Once this is merged, I'll do a similar refactor PR for `sign(...)`.

test/unit/verify/test_models.py
```diff
@@ -29,29 +32,43 @@ class TestVerificationMaterials:
     def test_rekor_entry_inconsistent_cve_2022_36056(
         self, signing_materials, signing_bundle
     ):
-        a_materials = signing_materials("a.txt")
-        offline_rekor_materials = signing_bundle("bundle.txt")
+        (_, a_materials) = signing_materials("a.txt")
+        (file, offline_rekor_materials) = signing_bundle("bundle.txt")
+
+        with file.open(mode="rb", buffering=0) as input_:
```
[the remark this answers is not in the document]
woodruffw: That works, although to go one step further: I think `signing_materials`/`signing_bundle` should probably just return `bytes`/`Hashed` directly, since the plan is to have both `sign` and `verify` take those types rather than do their own I/O 🙂

PR #910, opened 2024-02-26
Adding sct verification for certificates on verify flow

### LINKED ISSUE #886: `sigstore.verify` should check signing cert inclusion in the CT log
`sigstore-python` doesn't currently check the signing cert's presence in the CT log in the `verify` flow, but we are required to per the Sigstore Client specs:

> Verifier MUST extract the `SignedCertificateTimestamp` embedded in the leaf certificate, and verify it as in [RFC 9162 §8.1.3](https://datatracker.ietf.org/doc/html/rfc9162#name-validating-scts), using the verification key from the Certificate Transparency Log.

or:

> [...] Instead, Verifiers MAY perform online verification by fetching and validating inclusion proofs ([RFC 9162 §8.1.4](https://datatracker.ietf.org/doc/html/rfc9162#name-fetching-inclusion-proofs)) against a signed tree head. The Verifier SHOULD fetch the signed tree head in a manner that prevents equivocation by the Certificate Transparency log (e.g., by requiring signatures from independent “witnesses”).

javanlacerda:
#### Summary
It intends to close #886

We don't have sct verification for the certificate during the verification, only for sign. These PR implements sct verification for verifying flow.

#### Documentation

cc @jku

### DIFF
--- a/sigstore/_internal/sct.py
+++ b/sigstore/_internal/sct.py
@@ -162,21 +162,24 @@ class UnexpectedSctCountException(Exception):
     pass
 
 
-def get_sct_from_certificate(certificate: Certificate) -> SignedCertificateTimestamp:
+def _get_precertificate_signed_certificate_timestamps(
+    certificate: Certificate,
+) -> PrecertificateSignedCertificateTimestamps:
     # Try to retrieve the embedded SCTs within the cert.
     try:
         precert_scts_extension = certificate.extensions.get_extension_for_class(
             PrecertificateSignedCertificateTimestamps
         ).value
-    except ExtensionNotFound as ex:
-        raise ValueError("No PrecertificateSignedCertificateTimestamps found for the certificate")
+    except ExtensionNotFound:
+        raise ValueError(
+            "No PrecertificateSignedCertificateTimestamps found for the certificate"
+        )
 
     if len(precert_scts_extension) != 1:
         raise UnexpectedSctCountException(
             f"Unexpected embedded SCT count in response: {len(precert_scts_extension)} != 1"
         )
-    sct: SignedCertificateTimestamp = precert_scts_extension[0]
-    return sct
+    return precert_scts_extension
 
 
 def _cert_is_ca(cert: Certificate) -> bool:
--- a/sigstore/_internal/sct.py
+++ b/sigstore/_internal/sct.py
@@ -24,7 +24,12 @@ from typing import List, Optional
 
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import ec, rsa
-from cryptography.x509 import Certificate, ExtendedKeyUsage, ExtensionNotFound
+from cryptography.x509 import (
+    Certificate,
+    ExtendedKeyUsage,
+    ExtensionNotFound,
+    PrecertificateSignedCertificateTimestamps,
+)
 from cryptography.x509.certificate_transparency import (
     LogEntryType,
     SignedCertificateTimestamp,
@@ -147,6 +152,31 @@ def _get_issuer_cert(chain: List[Certificate]) -> Certificate:
     if _is_preissuer(issuer):
         issuer = chain[1]
     return issuer
+
+
+class UnexpectedSctCountException(Exception):
+    """
+    Number of percerts scts is wrong
+    """
+
+    pass
+
+
+def get_sct_from_certificate(certificate: Certificate) -> SignedCertificateTimestamp:
+    # Try to retrieve the embedded SCTs within the cert.
+    try:
+        precert_scts_extension = certificate.extensions.get_extension_for_class(
+            PrecertificateSignedCertificateTimestamps
+        ).value
+    except ExtensionNotFound as ex:
+        raise ValueError("No PrecertificateSignedCertificateTimestamps found for the certificate")
+
+    if len(precert_scts_extension) != 1:
+        raise UnexpectedSctCountException(
+            f"Unexpected embedded SCT count in response: {len(precert_scts_extension)} != 1"
+        )
+    sct: SignedCertificateTimestamp = precert_scts_extension[0]
+    return sct
 
 
 def _cert_is_ca(cert: Certificate) -> bool:
--- a/sigstore/_internal/trustroot.py
+++ b/sigstore/_internal/trustroot.py
@@ -98,13 +98,7 @@ class TrustedRoot(_TrustedRoot):
     @staticmethod
     def _get_tlog_keys(tlogs: list[TransparencyLogInstance]) -> Iterable[bytes]:
         """Return public key contents given transparency log instances."""
-
-        for key in tlogs:
-            if not _is_timerange_valid(key.public_key.valid_for, allow_expired=False):
-                continue
-            key_bytes = key.public_key.raw_bytes
-            if key_bytes:
-                yield key_bytes
+        return [key.public_key.raw_bytes for key in tlogs]
 
     @staticmethod
     def _get_ca_keys(
@@ -119,10 +113,10 @@ class TrustedRoot(_TrustedRoot):
                 yield cert.raw_bytes
 
     def get_ctfe_keys(self) -> list[bytes]:
-        """Return the active CTFE public keys contents."""
+        """Return the CTFE public keys contents."""
         ctfes: list[bytes] = list(self._get_tlog_keys(self.ctlogs))
         if not ctfes:
-            raise MetadataError("Active CTFE keys not found in trusted root")
+            raise MetadataError("CTFE keys not found in trusted root")
         return ctfes
 
     def get_rekor_keys(self) -> list[bytes]:
@@ -130,7 +124,7 @@ class TrustedRoot(_TrustedRoot):
         keys: list[bytes] = list(self._get_tlog_keys(self.tlogs))
 
         if len(keys) != 1:
-            raise MetadataError("Did not find one active Rekor key in trusted root")
+            raise MetadataError("Did not find one Rekor key in trusted root")
         return keys
 
     def get_fulcio_certs(self) -> list[Certificate]:
--- a/sigstore/verify/verifier.py
+++ b/sigstore/verify/verifier.py
@@ -121,7 +121,7 @@ class Verifier:
         establishing the trust chain for the signing certificate and signature.
         """
         self._rekor = rekor
-        self._fulcio_X509_certificate_chain: List[X509] = [
+        self._fulcio_certificate_chain: List[X509] = [
             X509.from_cryptography(parent_cert)
             for parent_cert in fulcio_certificate_chain
         ]
@@ -167,7 +167,7 @@ class Verifier:
         # method been called on it. To get around this, we construct a new one for every `verify`
         # call.
         store = X509Store()
-        for parent_cert_ossl in self._fulcio_X509_certificate_chain:
+        for parent_cert_ossl in self._fulcio_certificate_chain:
             store.add_cert(parent_cert_ossl)
 
         # In order to verify an artifact, we need to achieve the following:
--- a/sigstore/verify/verifier.py
+++ b/sigstore/verify/verifier.py
@@ -25,7 +25,7 @@ from typing import List, cast
 
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.primitives.asymmetric import ec
-from cryptography.x509 import Certificate, ExtendedKeyUsage, KeyUsage
+from cryptography.x509 import Certificate, ExtendedKeyUsage, KeyUsage, load_pem_x509_certificate
 from cryptography.x509.oid import ExtendedKeyUsageOID
 from OpenSSL.crypto import (
     X509,
@@ -34,6 +34,7 @@ from OpenSSL.crypto import (
     X509StoreContextError,
 )
 from pydantic import ConfigDict
+from sigstore._internal.fulcio.client import get_sct_from_certificate
 
 from sigstore._internal.merkle import (
     InvalidInclusionProofError,
@@ -43,7 +44,9 @@ from sigstore._internal.rekor.checkpoint import (
     CheckpointError,
     verify_checkpoint,
 )
+
 from sigstore._internal.rekor.client import RekorClient
+from sigstore._internal.sct import verify_sct
 from sigstore._internal.set import InvalidSETError, verify_set
 from sigstore._internal.trustroot import TrustedRoot
 from sigstore._utils import B64Str, HexStr
@@ -113,6 +116,7 @@ class Verifier:
         establishing the trust chain for the signing certificate and signature.
         """
         self._rekor = rekor
+        self.bbb = fulcio_certificate_chain
 
         self._fulcio_certificate_chain: List[X509] = []
         for parent_cert in fulcio_certificate_chain:
@@ -193,6 +197,15 @@ class Verifier:
             return CertificateVerificationFailure(
                 exception=store_ctx_error,
             )
+        
+        # step 2 todo update order
+        sct = get_sct_from_certificate(materials.certificate)
+        verify_sct(
+            sct,
+            materials.certificate,
+            self.bbb,
+            self._rekor._ct_keyring
+        )
 
         # 2) Check that the signing certificate contains the proof claim as the subject
         # Check usage is "digital signature"
--- a/sigstore/verify/verifier.py
+++ b/sigstore/verify/verifier.py
@@ -48,7 +48,10 @@ from sigstore._internal.rekor.checkpoint import (
     verify_checkpoint,
 )
 from sigstore._internal.rekor.client import RekorClient
-from sigstore._internal.sct import get_sct_from_certificate, verify_sct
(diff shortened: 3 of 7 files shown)


woodruffw:
Thanks @javanlacerda! Just give me a ping when this is ready for review 🙂

javanlacerda:
> Thanks @javanlacerda! Just give me a ping when this is ready for review 🙂

Sure!

Actually I opened the draft PR because I am trying to understand why the tests on `test_verifier.py` are failing. I am investigating it but I thought I could be good open the PR and ask for thoughts about it.

The test `test_verifier_offline` specifically is failing to verify the `bundle.txt`, but running by CLI with the command `sigstore --staging verify identity --bundle=test/unit/assets/bundle.txt.sigstore --cert-identity a@tny.town --cert-oidc-issuer https://github.com/login/oauth test/unit/assets/bundle.txt` it works.

jku:
I've had a look at the test case `test_verifier_offline`:
* I think I can see how it fails: the cert list used in `verify_sct()` seems to change when running in pytest (possibly just sorted differently?)
* I can't see why that happens at all -- I wonder if there's an issue in `mock_staging_tuf`? 


Below snippet should duplicate the test outside pytest but it does not fail like the test does:
```
>>> bundle = Bundle().from_json(open("test/unit/assets/bundle.txt.sigstore").read())
>>> materials = VerificationMaterials.from_bundle(input_=open("test/unit/assets/bundle.txt", "rb"), bundle=bundle, offline=True)
>>> verifier = Verifier.staging()
>>> verifier.verify(materials, NullPolicy())
VerificationSuccess(success=True)
```

jku:
> I wonder if there's an issue in mock_staging_tuf?

This is definitely related: if you remove `mock_staging_tuf` the test starts succeeding

sigstore/_internal/_utils.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    pass
+
+
+def get_sct_from_certificate(certificate: Certificate) -> SignedCertificateTimestamp:
+    # Try to retrieve the embedded SCTs within the cert.
+    precert_scts_extension = certificate.extensions.get_extension_for_class(
+        PrecertificateSignedCertificateTimestamps
+    ).value
```
jku, line 31: Now that we're running this code on certs coming from even more unknown sources, I think we need to add validation: It's likely just ExtensionNotFound that needs to be handled -- maybe turn it into ValueError here so the calling sites can just handle ValueError if they want?

woodruffw:
> @woodruffw do you remember the story behind this? Do we just update the embedded copy ?

Nope, I have no clue why that is 🙂

jku:
rebasing on #915 makes `test_verifier_offline` pass.

After that there seems to still be some oneline failures:
```
FAILED test/unit/verify/test_verifier.py::test_verifier_one_verification - sigstore._internal.sct.InvalidSCTKeyError
FAILED test/unit/verify/test_verifier.py::test_verifier_multiple_verifications - sigstore._internal.sct.InvalidSCTKeyError
FAILED test/unit/verify/test_verifier.py::test_verifier_email_identity - sigstore._internal.sct.InvalidSCTKeyError
FAILED test/unit/verify/test_verifier.py::test_verifier_policy_check - sigstore._internal.sct.InvalidSCTKeyError
FAILED test/unit/verify/test_verifier.py::test_verifier_invalid_signature - sigstore._internal.sct.InvalidSCTKeyError
FAILED test/unit/verify/test_verifier.py::test_verifier_invalid_online_missing_inclusion_proof - sigstore._internal.sct.InvalidSCTKeyError
```

all of these seem to be failing because the CTKeyring does not contain the key that matches `sct.logid`.

I am guessing this happens because our CTKeyring only contains active keys only: we have not needed to care about active vs non-active so far but in this case I believe we want all keys, not just active ones.

Assuming this is true, some design ideas to fix that:
* Keyring
  * Keyring could support looking up both active keys and all keys... this does not look very useful in practice though 
  * Or we could just have multiple different Keyrings: at least in ctfe case I think we always want active keys when signing, but we want all keys when verifying? I'm not 100% sure how that applies to e.g. RekorClient that is used in both cases -- is it fine to use all keys there?
* TrustedRoot has to be modified so it can return inactive ctfe keys too `(_get_tlog_keys` does the filtering currently): simple solution might be to rename `get_ctfe_keys` to `get_active_ctfe_keys()` and then add `get_all_ctfe_keys`: it's up to the call sites to choose what they need
* as an aside, TrustedRoot should likely return Keyrings directly (it's just made hard by _cli.py needing to support options like `--ctfe` so maybe we don't try it yet)

javanlacerda:
> rebasing on #915 makes `test_verifier_offline` pass.
Great!

javanlacerda:
> all of these seem to be failing because the CTKeyring does not contain the key that matches `sct.logid`.
> 
> I am guessing this happens because our CTKeyring only contains active keys only: we have not needed to care about active vs non-active so far but in this case I believe we want all keys, not just active ones.
> 
> Assuming this is true, some design ideas to fix that:
> 
> * Keyring
>   
>   * Keyring could support looking up both active keys and all keys... this does not look very useful in practice though
>   * Or we could just have multiple different Keyrings: at least in ctfe case I think we always want active keys when signing, but we want all keys when verifying? I'm not 100% sure how that applies to e.g. RekorClient that is used in both cases -- is it fine to use all keys there?
> * TrustedRoot has to be modified so it can return inactive ctfe keys too `(_get_tlog_keys` does the filtering currently): simple solution might be to rename `get_ctfe_keys` to `get_active_ctfe_keys()` and then add `get_all_ctfe_keys`: it's up to the call sites to choose what they need
> * as an aside, TrustedRoot should likely return Keyrings directly (it's just made hard by _cli.py needing to support options like `--ctfe` so maybe we don't try it yet)

Thank you @jku!! I will try all of these approaches!

javanlacerda:
Hey @jku, I performed modifications to return all keys instead of the active ones and it worked. I trying to figure out why the conformance is failing

javanlacerda:
I am wondering if I should have 2 separate functions, 1 that returns de active and other that returns both

jku:
> Hey @jku, I performed modifications to return all keys instead of the active ones and it worked. I trying to figure out why the conformance is failing

I did some more tests last night and I can reproduce the failure with signature bundles we produce:
```
sigstore sign README.md
sigstore verify identity --cert-identity whatever --cert-oidc-issuer whatever README.md
```
* SCT verify works during `sign`
* SCT verify fails during `verify identity`
* Looking at some debug prints it seems that the payload for the SCT signature verification is different when it runs during sign vs when it runs during verify

So there's a bug in how the payload is constructed during verify

javanlacerda:
> issuer_key_id

Hmmm, will the right cert always be at the index 1?

jku:
> Hmmm, will the right cert always be at the index 1?

oh no. It's just an example showing where the issue is: verify_sct() apparently picks the incorrect issuer certificate

sigstore/_internal/sct.py
```diff
@@ -132,21 +135,41 @@ def _pack_digitally_signed(
...
-    # If we do not have any EKU, we certainly do not have CT Ext
-    except ExtensionNotFound:
-        return False
+def _get_issuer_cert(chain: List[Certificate]) -> Certificate:
+    for cert in chain:
+        if "CN=sigstore-intermediate" in [
+            name.rfc4514_string() for name in cert.subject.rdns
+        ]:
+            return cert
+    return chain[0]
```
woodruffw, line 144: I think this is going to be unreliable, especially in the BYO PKI case -- the fact that the sigstore intermediate happens the have the `sigstore-intermediate` CN is probably an implementation detail 🙂 

I _think_ what we want to do here is make `_get_issuer_cert` have a signature like this:

```
def _get_issuer_cert(child: Certificate, candidates: list[Certificate]) -> Certificate
```

...and then have find the right cert in `candidates` by returning the first that has a `SubjectKeyIdentifier` that matches `child`'s `AuthorityKeyIdentifier`. Sigstore's X.509 profile should guarantee that both of those extensions are present.

woodruffw: Actually, thinking a little more: an even simpler implementation of `_get_issuer_cert` is to just call [verify_directly_issued_by](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Certificate.verify_directly_issued_by) with each candidate and return the first that matches.

javanlacerda: It makes a lot of sense for me!! I'll just do that

javanlacerda: Hey William, actually I was discussing with @haydentherapper and the root cause that the `_get_issuer_cert` was not working properly was that the `chain` is not exactly a chain but a `pool` of certificates for the verifying flow. So I performed a modification on the `verify_sct` call to pass the chain and not the pool. We still have a "bug" because these pool is called as a chain the the Verifier class, what can be confused.

sigstore/verify/verifier.py
```diff
@@ -113,11 +118,10 @@ def __init__(
...
         self._rekor = rekor
-
-        self._fulcio_certificate_chain: List[X509] = []
-        for parent_cert in fulcio_certificate_chain:
-            parent_cert_ossl = X509.from_cryptography(parent_cert)
-            self._fulcio_certificate_chain.append(parent_cert_ossl)
+        self._fulcio_X509_certificate_chain: List[X509] = [
+            X509.from_cryptography(parent_cert)
+            for parent_cert in fulcio_certificate_chain
+        ]
```
woodruffw, line 124: Nitpick, IMO no need for a rename: 

```suggestion
        self._fulcio_certificate_chain: List[X509] = [
            X509.from_cryptography(parent_cert)
            for parent_cert in fulcio_certificate_chain
        ]
```

sigstore/_internal/sct.py
woodruffw: Aha! Thank you for catching that -- that's always been a mild mystery to me (as to why we needed to search for the issuer at all, since it should always be the adjacent intermediate CA).

IMO we can rename the misleading parameter to `pool` or `trusted_certs` as part of this change -- both better convey that there's no "one true chain" here, and that the certs are part of a trust bundle rather than a part of a specific linear path.

sigstore/_internal/sct.py
```diff
@@ -149,6 +154,34 @@ def _get_issuer_cert(chain: List[Certificate]) -> Certificate:
...
+    """
+    Number of percerts scts is wrong
+    """
+
+    pass
+
+
+def _get_sct_from_certificate(
+    certificate: Certificate,
+) -> Any:
```
woodruffw, line 167: Hmm, can we refine the type here? This should return a `cryptography.x509.SignedCertificateTimestamp`, not `Any`: <https://cryptography.io/en/latest/x509/certificate-transparency/#cryptography.x509.certificate_transparency.SignedCertificateTimestamp>

sigstore/_internal/sct.py
haydentherapper: Out of scope for this PR, but another question is if the interface for [`get_fulcio_certs`](https://github.com/sigstore/sigstore-python/blob/8ea2a7c2c3664f932122a53daa67a09411beac49/sigstore/_internal/trustroot.py#L136) should be changed to return a list of chains rather than a concatenated list.

This is related to how `X509Store` works. It's not immediately clear to me how it does. It seems like you push all certificates that are trusted, intermediates or roots, into the store (which I assume is why the interface returns a single list rather than a list of lists). When we were testing, we saw the verified chain contain both the intermediate and root Fulcio certs, so I assume the store tries to verify down to a self-signed certificate? Again, unrelated to the PR, but I wanted to check that the intermediate, even if marked as trusted and not just a chain builder, is getting verified against a root.

haydentherapper: https://github.com/openssl/openssl/blob/master/crypto/x509/x509_vfy.c#L3204 - this is the "fun" function to read that answers this question. I _think_ openssl defines a trusted root certificate as a self-signed cert, so even if we add intermediates to the trust store, the chain must terminate with a self-signed cert. I think.

woodruffw: >  I think openssl defines a trusted root certificate as a self-signed cert, so even if we add intermediates to the trust store, the chain must terminate with a self-signed cert. I think...

It does, but that's actually technically incorrect of them 😅: RFC 3280/5280 only require chain construction to the first trust anchor, regardless of whether it's a self-signed root or not. OpenSSL's default behavior is to continue searching until it hits a self-signed root in the trusted set, which caused a decent amount of pain in ~2020 (IIRC) when one of the Root CAs that cross-signed for `Let's Encrypt Authority R3` expired. They added `X509_V_FLAG_PARTIAL_CHAIN` as a flag to enable the correct behavior, but it isn't the default.

TL;DR: We might need to pass `X509_V_FLAG_PARTIAL_CHAIN` if we really do need the "long chain." But I'm not sure that we do -- isn't it sufficient to just verify against the intermediate and return the shorter chain?

haydentherapper: Given the root is present in the trusted root file, I would prefer we still validate it. The intermediate being shipped in the trust root is more of an optimization, letting us avoid distributing a CRL if we need to revoke the intermediate cert. The PKI root of trust is supposed to be the Fulcio root cert.

Long-term, we should revisit whether we want a Fulcio intermediate. Might make sense to just issue off the root. Originally, the thought was we'd have one Fulcio root that could be used to issue intermediates for other use cases, eg code signing and timestamping. But given we haven't needed that yet, maybe we should just issue off the root.

woodruffw: > Given the root is present in the trusted root file, I would prefer we still validate it. The intermediate being shipped in the trust root is more of an optimization, letting us avoid distributing a CRL if we need to revoke the intermediate cert. The PKI root of trust is supposed to be the Fulcio root cert.

Makes sense! In that case, we can simulate the "long chain" building by passing only the Fulcio root into the trust store and including the intermediate via the `intermediates` parameter to `X509StoreContext`. That will ensure we never terminate before the self-signed TA, regardless of OpenSSL flags like `X509_V_FLAG_PARTIAL_CHAIN`.

sigstore/_internal/sct.py
javanlacerda: Actually I did that before, but I had problem with the linter, because [here](https://github.com/sigstore/sigstore-python/blob/8ea2a7c2c3664f932122a53daa67a09411beac49/sigstore/_internal/fulcio/client.py#L278C13-L278C16) the variable sct can be a `SignedCertificateTimestamp` but [here](https://github.com/sigstore/sigstore-python/blob/8ea2a7c2c3664f932122a53daa67a09411beac49/sigstore/_internal/fulcio/client.py#L301) it can be a `DetachedFulcioSCT`, and the linter doesn't understand the inheritance.

javanlacerda: I did a workaround for this by returning the `PrecertificateSignedCertificateTimestamps` and getting the `SignedCertificateTimestamp` on the caller flow instead.

woodruffw:
Thanks @javanlacerda! One last nitpick in https://github.com/sigstore/sigstore-python/pull/910#discussion_r1511786637, and I think this is good to go.

(Don't worry about force-pushing, BTW -- I do squash merges for PRs.)

woodruffw (approved):
LGTM! @javanlacerda if you merge/rebase against `main`, I should be able to merge this.

javanlacerda:
> LGTM! @javanlacerda if you merge/rebase against `main`, I should be able to merge this.

Done!

woodruffw:
Thanks @javanlacerda!

sigstore/_internal/trustroot.py
```diff
@@ -98,13 +98,7 @@ def staging(cls, offline: bool = False) -> "TrustedRoot":
...
     def _get_tlog_keys(tlogs: list[TransparencyLogInstance]) -> Iterable[bytes]:
         """Return public key contents given transparency log instances."""
-
-        for key in tlogs:
-            if not _is_timerange_valid(key.public_key.valid_for, allow_expired=False):
-                continue
-            key_bytes = key.public_key.raw_bytes
-            if key_bytes:
-                yield key_bytes
+        return [key.public_key.raw_bytes for key in tlogs]
```
jku, line 102: are we sure this change is correct for rekor keys?

jku:
This implementation found a bug in how the issuers certificates where handled, a case our existing test suite didn't cover.

Are we ok with just trusting the conformance suite for that?

jku:
The exception comments were never addressed but I suppose none of that is fatal: I do think we should be handling exceptions (but should not be adding custom exceptions for no reason) since we're now running  `_get_precertificate_signed_certificate_timestamps()` on content that could be coming from anywhere.

sigstore/_internal/trustroot.py
haydentherapper: What was the reason for removing `_is_timerange_valid`?

javanlacerda: `_is_timerange_valid` was the way to get only the active keys, but we should use inactive keys as well. Could we keep the generator allowing expired?

woodruffw: Whoops, I missed this comment during review. ++ to @haydentherapper's question about the timerange check.

CC @javanlacerda

PR #926, opened 2024-03-07
fix: rollback code for return generator allowing expired keys

### LINKED ISSUE #925: get_tlog_keys returning list instead of generator
**Description**
The PR #910 modified the function get_tlog_keys to return all keys instead of the active ones, the problem is that it is returning a list of keys instead of a generator, that was the expected pattern.

**Version**

javanlacerda:
#### Summary
Closes #925
The PR #910 modified the function for retuning all keys instead of the active keys. It should return a generator but the PR made return a list.

This PR makes the function return a generator with all keys by allowing expired keys on `_is_timerange_valid` function.


woodruffw (approved):

woodruffw:
Thanks @javanlacerda!

PR #930, opened 2024-03-11
sigstore: use our own Statement type

woodruffw:
This solves our "which serialization is right" problem with DSSE by replacing our use of `in_toto_attestation` with our own `Statement` type, which is surfaced to the user as an opaque `bytes` container but ensures (internally) that the JSON object structure is as expected.

Long term, we'll probably want to reintroduce `in_toto_attestation` as a dep in some form, but keep our `Statement` as the public API type for interface stability reasons.


sigstore/dsse.py
```diff
@@ -0,0 +1,184 @@
...
+
+        signature = key.sign(pae, ec.ECDSA(hashes.SHA256()))
+        return Envelope(
+            payload=self._contents,
+            payload_type=Statement._ENVELOPE_TYPE,
+            signatures=[Signature(sig=signature, keyid=None)],
+        )
+
+
+class _StatementBuilder:
```
woodruffw, line 131: NB: We will probably want to make this API public at some point, but not yet.

sigstore/dsse.py
```diff
@@ -0,0 +1,184 @@
...
+            stmt = _Statement(
+                type_="https://in-toto.io/Statement/v1",
+                subjects=self._subjects,
+                predicate_type=self._predicate_type,
+                predicate=self._predicate,
+            )
+        except ValidationError as e:
+            raise ValueError(f"invalid statement: {e}")
+
+        return Statement(stmt.model_dump_json(by_alias=True).encode())
```
woodruffw, line 184: A consideration here: technically we make no canonical/serialization guarantees other than "it's valid JSON", but we *could* do RFC 8785 for good measure here.

sigstore/dsse.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+class Envelope:
```
woodruffw, line 170: Added this as a thin wrapper for `sigstore_protobuf_specs.io.intoto.Envelope`, so that we don't leak a protobuf model into the public API here.

jku (approved):
Code looks good to me.  I'm not too familiar with in-toto but since the signing is essentially still experimental/private  API (because _StatementBuilder and _Statement are private and CLI does not expose this) seems fine to merge.

Are there conformance tests for statement bundles already?

Should we have a full round trip test with both sign & verify for the statement bundle?

test/unit/test_sign.py
```diff
@@ -148,3 +149,27 @@ def test_sign_prehashed(staging):
...
+            [_Subject(name="null", digest={"sha256": hashlib.sha256(b"").hexdigest()})]
+        )
+        .predicate_type("https://cosign.sigstore.dev/attestation/v1")
+        .predicate(
+            {
+                "Data": "",
+                "Timestamp": "2023-12-07T00:37:58Z",
+            }
+        )
+    ).build()
```
jku, line 172: what's the deal with this `(...).build()`  structure? Looks weird

woodruffw: Yeah, that's how Python wants it -- I think there are a few other valid formats, but `ruff` and `black` both favor that one

woodruffw:
> Are there conformance tests for statement bundles already?

I believe so, but we currently xfail them because we only support signing, not verifying. Once this is merged I can work on verification support.

> Should we have a full round trip test with both sign & verify for the statement bundle?

Yep, I'll do this with the aforementioned followup 🙂

PR #935, opened 2024-03-13
sign: fix envelope type

woodruffw:
We need the inner proto type here.


woodruffw:
I've updated `test_sign_dsse` to test this as well.

jku (approved):

PR #936, opened 2024-03-14
Key management improvement

### LINKED ISSUE #927: Improve our internal key management
sigstore-python's internal key management is currently a bit of a mess. The following parties are at play:

1. TUF/the trust root/bundle: this is the ultimate source of all of our "a priori" key material.
2. Fulcio: the trust root distributes one or more Fulcio CAs as chains of X.509 certificates. Internally, we mash these into a single pool and return certificates with expired timeranges at the trustroot level (which is probably incorrect).
3. Certificate Transparency ("CTFE"): the trust root distributes one or more CTFE keys, which we load so long as they are valid or expired (but not from the future).
4. Rekor: the trust root distributes one or more Rekor keys, which we load similarly to CTFE keys.

Now, the status quo:

1. Our internal `RekorClient` has `_ct_keyring` and `_rekor_keyring` members, representing the CTFE and Rekor keyrings respectively as loaded from the trust root. The latter makes a little sense; the former makes no sense (since CTFE has nothing to do with Rekor). 
2. Our internal `FulcioClient` has no keyrings or certificate pools internally.
3. We currently make no distinction in loaded keys/certs between signing and verifying, which is almost certainly incorrect.

Now, a solution:

1. We should keep key management entirely within the `TrustedRoot` and its APIs. In practice, that means that `TrustedRoot` should have APIs roughly like the following:
    ```python
    TrustedRoot.rekor_keyring(disposition="sign") # a suitable set of Rekor keys for use in the signing process
    TrustedRoot.ctfe_keyring(disposition="verify") # a suitable set of CTFE keys for use during verification
    ```

    ...and so forth. The `disposition` kwarg is pretty ugly; we just need _some_ way to differentiate between the set of keys that are valid in each context.
1. `SigningContext` and `Verifier` should probably take a `TrustedRoot` directly. 
1. `RekorClient` should have no key material within it at all.

CC @jku for thoughts.

javanlacerda:
Closes #927

It makes a refactor on code for a better key management. It makes the trusted root fully responsible by the key management, concentrating the responsibilities there. 

It also removes the unused flags from the CLI.  These properties are already provided by the trusted root.
Removed flags:
 * `--certificate-chain`
 * `--rekor-root-pubkey`
 * `-ctfe`
#### Summary

It is a refactor to make the code cleaner and cohesive.

#### Documentation

cc @woodruffw @jku


jku (commented):
I left some specific comments in code. The main feedback is that these items from william are the core ones:

> * SigningContext and Verifier should probably take a TrustedRoot directly.
> * RekorClient should have no key material within it at all.

Just to add to those:
* SigningContext and Verifier could still have staging() and production() helpers to make things easier for API users but  the normal constructor call should look something like   `Verifier(rekor_client, trusted_root)`
* CLI should create the TrustedRoot, and pass it to Verifier and SigningContext
* the CLI legacy args could be supported by "modifying" the TrustedRoot based on those args
* Verifier and SigningContext know that their purposes are  verifying and signing: this does not need to be an argument in any of their public methods
* RekorClient should not need a trust root or a purpose 

It's a long list of things but I think when we manage that , the code is hopefully significantly easier to understand.

jku:
I did not address the actual key choices in different situations... It's friday afternoon and I'm pretty sure I'll get it wrong. If you can document in a comment which keys are used in which situations in your PR that would probably be useful.

The basic idea with KeyringPurpose seems good.

woodruffw:
Thanks @javanlacerda! I agree with @jku's feedback, and I'll add some of my own in a bit as well.

For deconflicting: this will probably interact a bit with #937. In particular, that changeset _may_ end up removing `RekorClient` from `Verifier`, since the `Verifier` will become totally offline and will only need the Rekor keyring, not Rekor REST APIs.

sigstore/_internal/trustroot.py
```diff
@@ -58,6 +62,19 @@ def _is_timerange_valid(period: TimeRange | None, *, allow_expired: bool) -> boo
     return allow_expired or (period.end is None or now <= period.end)
 
 
+class KeyringPurpose(str, Enum):
+    """
+    Keyring purpose typing
+    """
+
+    SIGN = "Signing"
+    VERIFY = "Verifying"
```
woodruffw, line 71: Nitpick: either all-lower or all-upper enum values are idiomatic:

```suggestion
    SIGN = "sign"
    VERIFY = "verify"
```

sigstore/_internal/trustroot.py
```diff
@@ -117,16 +138,29 @@ def _get_ca_keys(
...
+    def rekor_keyring(self, purpose: KeyringPurpose) -> RekorKeyring:
+        """Return public key contents given certificate authorities."""
+
+        if purpose is KeyringPurpose.VERIFY:
+            allow_expired = True
+        elif purpose is KeyringPurpose.SIGN:
+            allow_expired = False
+        else:
+            # TODO(javan): Create a specific exception
+            raise Exception()
```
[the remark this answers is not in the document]
woodruffw: Agreed -- we could also remove the indirection here entirely, and have `_get_tlog_keys(tlogs, purpose)`. But no super strong opinion there.

sigstore/_internal/trustroot.py
```diff
@@ -32,9 +38,12 @@
     TrustedRoot as _TrustedRoot,
 )
 
+from sigstore._internal.keyring import Keyring
 from sigstore._internal.tuf import DEFAULT_TUF_URL, STAGING_TUF_URL, TrustUpdater
 from sigstore.errors import MetadataError
 
+RekorKeyring = NewType("RekorKeyring", Keyring)
```
woodruffw, line 45: This feels a little out of place 🙂 -- IMO we should either put this back into `rekor/client.py` *or* remove `keyring.py` entirely and inline it into this module; I have a slight preference for the latter (since the Keyring abstraction no longer makes sense outside of a trust root).

woodruffw:
@javanlacerda JFYI: it's easier to review the diffs here if you don't force-push each time -- it's okay to have a messy commit history while the PR is in progress, since we can squash or amend it before merge 🙂

javanlacerda:
> @javanlacerda JFYI: it's easier to review the diffs here if you don't force-push each time -- it's okay to have a messy commit history while the PR is in progress, since we can squash or amend it before merge 🙂

I apologize!! I had to do that as I forgot to sign one of the commits in the middle 😞

woodruffw:
> I apologize!! I had to do that as I forgot to sign one of the commits in the middle 😞

No worries! I personally do `alias gc=git commit --signoff`, but also don't worry too much about missing ones: I can also fix that during squash.

woodruffw:
This is looking pretty good to me, modulo the design question about the CLI flags. That boils down to whether we want to remove the CLI flags outright (fine by me, personally), or whether we want to keep them around for a bit and internally convert them into a `TrustRoot`.

I personally lean towards the former: we've been wanting to get rid of them for a while, and doing so will make future improvements much more straightforward. If @jku agrees, I think we can just go ahead and delete all of that 🙂

sigstore/_internal/trustroot.py
```diff
@@ -58,48 +76,201 @@ def _is_timerange_valid(period: TimeRange | None, *, allow_expired: bool) -> boo
...
 
         If `offline`, will use trust root in local TUF cache. Otherwise will
         update the trust root from remote TUF repository.
         """
         path = TrustUpdater(url, offline).get_trusted_root_path()
-        return cls.from_file(path)
+        trusted_root = cls.from_file(path)
+        trusted_root.args = args
+        trusted_root.purpose = purpose
+        return trusted_root
```
jku, line 228:
```suggestion
        return cls.from_file(path, args, purpose)
```

same suggestion applies to all other heper constructors

sigstore/_internal/trustroot.py
```diff
@@ -117,20 +288,36 @@ def _get_ca_keys(
             for cert in ca.cert_chain.certificates:
                 yield cert.raw_bytes
 
-    def get_ctfe_keys(self) -> list[bytes]:
+    def rekor_keyring(self) -> RekorKeyring:
+        """Return public key contents given certificate authorities."""
+
+        return RekorKeyring(self._get_rekor_keys())
```
jku, line 294: I believe you can just remove this wrapper method, and change 
 `_get_rekor_keys(self) -> Keyring`
 into
 `rekor_keyring(self) -> RekorKeyring`
right?

sigstore/_internal/rekor/client.py
```diff
@@ -255,13 +250,10 @@ def production(cls, trust_root: TrustedRoot) -> RekorClient:
...
-        rekor_keys = trust_root.get_rekor_keys()
-        ctfe_keys = trust_root.get_ctfe_keys()
-
         return cls(
             DEFAULT_REKOR_URL,
-            RekorKeyring(Keyring(rekor_keys)),
-            CTKeyring(Keyring(ctfe_keys)),
+            rekor_keyring=trust_root.rekor_keyring(),
+            ct_keyring=trust_root.ct_keyring(),
         )
```
jku, line 257: I was kind of expecting RekorClient to not need at least one of these (that the users of these keyrings actually have access to a proper keyring and should be using these)... but maybe that's fine to leave for future PR?

jku (commented):
Looks really nice, this seems like the right direction. As for CLI arguments: I think I'm fine with removing them -- we could ping the slack channel with the specific list of arguments we intend to remove to see if anyone complains...

Are we going to lose testing coverage if the arguments are removed? probably not?

I might not have time this week to do proper testing/review but I left some minor comments, generally this looks good. I would really love more "description of changes" (as a simple example, if you move a bunch of code from one file to another, maybe mention if it's a direct copy or somehow modified).

javanlacerda:
> Looks really nice, this seems like the right direction. As for CLI arguments: I think I'm fine with removing them -- we could ping the slack channel with the specific list of arguments we intend to remove to see if anyone complains...
> 
> Are we going to lose testing coverage if the arguments are removed? probably not?
> 
> I might not have time this week to do proper testing/review but I left some minor comments, generally this looks good. I would really love more "description of changes" (as a simple example, if you move a bunch of code from one file to another, maybe mention if it's a direct copy or somehow modified).

Awesome! I will start making more comments about decisions. 

Do you guys think we should create a different issue/PR for removing the flags? cc @jku @woodruffw

woodruffw:
> Are we going to lose testing coverage if the arguments are removed? probably not?

I _believe_ not -- we basically don't hit the `_cli.py` routines at all in tests at the moment.

> Do you guys think we should create a different issue/PR for removing the flags?

I'm okay with either way. If we do it in a separate PR, we should merge that one first to ensure we don't have to adapt these changes too much. But if you find it easier to just update this one, I don't mind a larger diff here 🙂

javanlacerda:
I did these forced push because I just removed a remaining `print` 😆 @woodruffw

sigstore/_internal/rekor/client.py
javanlacerda: I just removed any trustedroot and keyring dependency from rekor client :)

sigstore/_internal/trustroot.py
javanlacerda: it makes sense

sigstore/_cli.py
woodruffw, line 1: I think this file also needs deletions for the CLI flags themselves: the `_parser()` function needs to be updated to remove each of the arg definitions that have had their usages removed below.

sigstore/_internal/trustroot.py
```diff
@@ -117,20 +279,27 @@ def _get_ca_keys(
...
 
+        keys: list[bytes]
+        if self.args and self.args.rekor_root_pubkey:
+            keys = self.args.rekor_root_pubkey.read()
+        else:
+            keys = list(self._get_tlog_keys(self.tlogs, self.purpose))
         if len(keys) != 1:
             raise MetadataError("Did not find one Rekor key in trusted root")
-        return keys
+        return RekorKeyring(Keyring(keys))
```
woodruffw, line 292: This should work:

```suggestion
        return RekorKeyring(keys)
```

sigstore/_internal/trustroot.py
```diff
@@ -117,20 +279,27 @@ def _get_ca_keys(
...
+
+    def ct_keyring(self) -> CTKeyring:
+        """Return keyring with key for CTFE."""
+        if self.args and self.args.ctfe_pem:
+            ctfes = [self.args.ctfe_pem.read()]
+        else:
+            ctfes = list(self._get_tlog_keys(self.ctlogs, self.purpose))
+        if not ctfes:
+            raise MetadataError("CTFE keys not found in trusted root")
+        return CTKeyring(Keyring(ctfes))
```
woodruffw, line 302:
```suggestion
        return CTKeyring(ctfes)
```

sigstore/_internal/trustroot.py
```diff
@@ -58,48 +76,192 @@ def _is_timerange_valid(period: TimeRange | None, *, allow_expired: bool) -> boo
...
 
+    args: Optional[Namespace] = None
+    purpose: KeyringPurpose
+
     @classmethod
-    def from_file(cls, path: str) -> "TrustedRoot":
+    def from_file(
+        cls,
+        path: str,
+        args: Optional[Namespace] = None,
```
woodruffw, line 202: This looks off -- I don't think we want to push the CLI state into the `TrustedRoot` like this. I _think_ we want to remove this entirely, right?

javanlacerda: Yes, you're right! I did that as a workaround for having args, but it seems weird and all args should be removed.

CHANGELOG.md
```diff
@@ -363,6 +363,11 @@ This is a corrective release for [2.1.1].
   replacing the material that was previously baked into `sigstore._store`
   ([#351](https://github.com/sigstore/sigstore-python/pull/351))
 
+### Changed
+* `sigstore verify and sign` doesn't support `--certificate-chain` and `--rekor-root-pubkey`
```
woodruffw, line 367: This should be under `### Removed`, I think 🙂

javanlacerda: Right! Changing now!

sigstore/_internal/trustroot.py
javanlacerda: I agree with you that it should work as RekorKeyring is a subtype of Keyring, but it doesn't :/

README.md
```diff
@@ -137,7 +132,7 @@ usage: sigstore sign [-h] [-v] [--identity-token TOKEN] [--oidc-client-id ID]
                      [--oauth-force-oob] [--no-default-files]
                      [--signature FILE] [--certificate FILE] [--bundle FILE]
                      [--output-directory DIR] [--overwrite] [--staging]
-                     [--rekor-url URL] [--rekor-root-pubkey FILE]
+                     [--rekor-url URL]
                      [--fulcio-url URL] [--ctfe FILE]
```
woodruffw, line 136: I believe we also want to get rid of `--ctfe`, since it's also loaded from the TrustedRoot.

CHANGELOG.md
```diff
@@ -364,7 +364,7 @@ This is a corrective release for [2.1.1].
   ([#351](https://github.com/sigstore/sigstore-python/pull/351))
 
 ### Removed
-* `sigstore verify and sign` doesn't support `--certificate-chain` and `--rekor-root-pubkey`
+* CLI: The `--certificate-chain` and `--rekor-root-pubkey` flags have been entirely removed ([#936](https://github.com/sigstore/sigstore-python/pull/936))
   during verification anymore.
```
woodruffw:
```suggestion
```

javanlacerda:
> Thanks @javanlacerda, this is looking pretty good! I'm going to give these changes a spin locally and do a final review afterwards, with the goal of merging today.
> 
> (Mind updating the branch again?)

Done! :)

jku:
please leave comments about changes (especially when force pushing): even a _"rebase, no code changes"_ is useful.

/gcbrun

woodruffw (approved):

woodruffw (approved):
LGTM and didn't observe any breakage in local testing. Thanks @javanlacerda!

PR #945, opened 2024-03-19
sigstore: use rfc8785 for SET canonicalization

woodruffw:
SET canonicalization is defined over RFC 8785, but we were using securesystemslib's `encode_canonical`, which uses an entirely different (but coincidentally identical, in this case) canonicalization scheme.

CC @haydentherapper for confirmation that SETs do indeed use RFC 8785 -- that understanding is based on https://github.com/sigstore/rekor/blob/4fcdcaa58fd5263560a82978d781eb64f5c5f93c/openapi.yaml#L467-L471 🙂

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ dependencies = [
   "pyOpenSSL >= 23.0.0",
   "requests",
   "rich ~= 13.0",
-  "securesystemslib",
+  "rfc8785",
   "sigstore-protobuf-specs ~= 0.3",
   # NOTE(ww): Under active development, so strictly pinned.
   "sigstore-rekor-types == 0.0.12",
--- a/sigstore/transparency.py
+++ b/sigstore/transparency.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 
 from typing import Any, List, Optional
 
+import rfc8785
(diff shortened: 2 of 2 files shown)


woodruffw:
Blocked on me making `rfc8785` work on Python 3.8 and 3.9.

haydentherapper:
That is correct. We use a Golang JSON canonicalization library - https://github.com/sigstore/rekor/blob/6cdd40d56e924cf13e09b40d53a04faed5e3e8c6/pkg/api/entries.go#L65C24-L65C41 - that implements RFC 8785. (Though the [comment](https://github.com/cyberphone/json-canonicalization/blob/ba74d44ecf5f39a0508561875584eb08f888e808/go/src/webpki.org/jsoncanonicalizer/jsoncanonicalizer.go#L18) in the library should be updated to the RFC and not the draft!)

woodruffw:
Thanks for confirming! This eases my concern that Sigstore effectively requires two different JSON canonicalization schemes 😅

haydentherapper (approved):

jku (approved):
looks good. 

(to be clear I've not reviewed the rfc8785 implementation just took a quick look)

PR #952, opened 2024-04-02
Bump protobuf-specs, handle v3 media types

woodruffw:
This bumps sigstore-protobuf-specs to 0.3.1 and ensures that we handle the new `application/vnd.dev.sigstore.bundle.v0.3+json` media type, with a (warned) fallback for the old format.

Also adds tests for both.


sigstore/verify/models.py
```diff
@@ -180,7 +180,13 @@ def _verify_bundle(self) -> None:
             raise InvalidBundle(f"unsupported bundle format: {self._inner.media_type}")
 
         # Extract the signing certificate.
-        if media_type == BundleType.BUNDLE_0_3:
+        if media_type in (BundleType.BUNDLE_0_3, BundleType.BUNDLE_0_3_ALT):
+            if media_type == BundleType.BUNDLE_0_3_ALT:
+                _logger.warning(
+                    f"media type {BundleType.BUNDLE_0_3_ALT} is discouraged; "
+                    f"consider using {BundleType.BUNDLE_0_3} for v3 bundles"
+                )
```
jku, line 188: I understand the intent but this seems unproductive as verifying "historical" bundles is an expected use case. Bundles with BUNDLE_0_3_ALT were correct at the time they were made (as far as this code knows) so should be considered correct now: we should just verify it without complaints and should not imply that a mistake has been made.

jku (changes requested):
Looks good otherwise but I disagree about the warning: let me know what you think

sigstore/verify/models.py
woodruffw: Fair argument. I'm okay with not warning here; will remove.

jku (approved):

PR #953, opened 2024-04-02
sigstore, test: honor PublicKeyDetails when loading Keyrings

woodruffw:
See https://github.com/sigstore/sigstore-conformance/issues/140 for some context -- we were previously ignoring `PublicKeyDetails` entirely.

This adds a `Key` abstraction that gets constructed from the `PublicKey` message, handling `PublicKeyDetails` in the process. It also cleans up some of our error types and old tests (like `test_ctfe.py`, which is entirely obviated by the trustroot APIs and their tests now).

Separately, this will also "fix" our handling of key IDs per https://github.com/sigstore/rekor/issues/2062, in the sense that we'll go from failing closed on key ID mismatch to trying all keys in the keyring.


sigstore/_internal/trustroot.py
```diff
@@ -102,72 +106,116 @@ class KeyringSignatureError(KeyringError):
...
+        if public_key.key_details in self._RSA_SHA_256_DETAILS:
+            hash_algorithm = hashes.SHA256()
+            key = load_der_public_key(public_key.raw_bytes, types=(rsa.RSAPublicKey,))
+        elif public_key.key_details in self._EC_DETAILS_TO_HASH:
+            hash_algorithm = self._EC_DETAILS_TO_HASH[public_key.key_details]
+            key = load_der_public_key(
+                public_key.raw_bytes, types=(ec.EllipticCurvePublicKey,)
+            )
+        else:
+            raise InvalidKeyError(f"unsupported key type: {public_key.key_details}")
```
woodruffw, line 147: NB: This brings us to parity with the previous behavior, i.e. allowing ECDSA and RSA. But it gives us the flexility to allow additional keytypes in the future 🙂

test/unit/internal/test_ctfe.py
woodruffw, line 1: NB: This file entirely tested the old keyring loading APIs, including APIs that were never used anyways (like `Keyring.add`)

sigstore/_internal/trustroot.py
haydentherapper: +1 to a nice error message

sigstore/_internal/trustroot.py
```diff
@@ -102,72 +106,119 @@ class KeyringSignatureError(KeyringError):
...
+        _PublicKeyDetails.PKIX_RSA_PKCS1V15_3072_SHA256,
+        _PublicKeyDetails.PKIX_RSA_PKCS1V15_4096_SHA256,
+    }
+
+    _EC_DETAILS_TO_HASH: ClassVar[dict[_PublicKeyDetails, hashes.HashAlgorithm]] = {
+        _PublicKeyDetails.PKIX_ECDSA_P256_SHA_256: hashes.SHA256(),
+        _PublicKeyDetails.PKIX_ECDSA_P384_SHA_384: hashes.SHA384(),
+        _PublicKeyDetails.PKIX_ECDSA_P521_SHA_512: hashes.SHA512(),
+    }
+
```
haydentherapper, line 131: no need to do this here, but is it relatively straightforward to add ed25519 down the line?

haydentherapper (commented):
> Separately, this will also "fix" our handling of key IDs per https://github.com/sigstore/rekor/issues/2062, in the sense that we'll go from failing closed on key ID mismatch to trying all keys in the keyring.

Only concern with this is it's inefficient to verify with all keys. Is the motivation to avoid dealing with the key ID calculation and continuing to support RSA?

sigstore/_internal/trustroot.py
woodruffw: Yep! It should just be another branch (below) in the future.

woodruffw:
> Only concern with this is it's inefficient to verify with all keys. Is the motivation to avoid dealing with the key ID calculation and continuing to support RSA?

Pretty much -- I intend to fix the key ID calculation as well, but I didn't want to blow this diff up too much 🙂

jku (approved):
looks good, thanks. 

test_trust_root.py is really begging to be simplified a bit... but this is fine.

PR #959, opened 2024-04-04
sigstore, test: drastically simplify error types

woodruffw:
WIP.

Another chore/refactor PR: this drastically reduces the number of error types we have floating around the codebase, in favor of a single unified `VerificationError` (which we can specialize in the future, as necessary).

Additionally, this changes the `verify(...)` API from returning a `VerificationResult` to using exceptional error handling. There are three justifications for this:

1. It allows us to remove the `VerificationResult` model and its subclasses, which are mostly just duplication over existing error types
2. It reduces the likelyhood of `VerificationError`s leaking through the API accidentally
3. It aligns the API with that the DSSE/in-toto API will do, which is use a return value on the success case and exceptions for all errors.

In the process, this also brings our coverage up a bit.

### DIFF
--- a/sigstore/errors.py
+++ b/sigstore/errors.py
@@ -117,3 +117,9 @@ class RootError(Error):
         Unable to establish root of trust.
 
         This error may occur when the resources embedded in this distribution of sigstore-python are out of date."""
+
+
+class VerificationError(Error):
+    """
+    Raised whenever any phase or subcomponent of Sigstore verification fails.
+    """
--- a/sigstore/_internal/merkle.py
+++ b/sigstore/_internal/merkle.py
@@ -30,17 +30,10 @@ import typing
 from typing import List, Tuple
 
 from sigstore._utils import HexStr
+from sigstore.errors import VerificationError
 
 if typing.TYPE_CHECKING:
     from sigstore.transparency import LogEntry
-
-
-class InvalidInclusionProofError(Exception):
-    """
-    Raised if the Merkle inclusion proof fails.
-    """
-
-    pass
 
 
 _LEAF_HASH_PREFIX = 0
@@ -112,8 +105,8 @@ def verify_merkle_inclusion(entry: LogEntry) -> None:
 
     # Check against the number of hashes.
     if len(inclusion_proof.hashes) != (inner + border):
-        raise InvalidInclusionProofError(
-            f"Inclusion proof has wrong size: expected {inner + border}, got "
+        raise VerificationError(
+            f"inclusion proof has wrong size: expected {inner + border}, got "
             f"{len(inclusion_proof.hashes)}"
         )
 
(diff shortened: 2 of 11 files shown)


sigstore/_cli.py
```diff
@@ -227,7 +224,7 @@ def _parser() -> argparse.ArgumentParser:
         "-v",
         "--verbose",
         action="count",
-        default=argparse.SUPPRESS,
+        default=0,
```
jku, line 227: IIRC correctly this will break "-v" at some location of the command as now every parser will set verbose value (even if a previous parser set a non-zero value).... maybe it was `sigstore -v cmd` that breaks this way (see f5dbd71d)

or does your overwriting of `args.verbose` deal with this issue?

sigstore/errors.py
```diff
@@ -26,9 +27,9 @@ class Error(Exception):
     def diagnostics(self) -> str:
         """Returns human-friendly error information."""
 
-        return """An issue occurred."""
+        return str(self)
 
-    def print_and_exit(self, raise_error: bool = False) -> None:
+    def log_and_exit(self, logger: Logger, raise_error: bool = False) -> None:
```
jku, line 32: this is maybe unrelated: is `raise_error` useful for anything?

jku (approved):
Nice,
* Totally agree about the exception simplification
* `Verifier.verify()` API change is a little annoying but it does seem to make a lot of code more straight forward

left a question on the argument parsing, otherwise LGTM.

sigstore/errors.py
woodruffw: So-so: we use it to spit out the original traceback when `args.verbose >= 1`, but I could be convinced to just inline that logic directly into `_cli.py`'s error handling.

sigstore/_cli.py
woodruffw: Ah yeah, I had forgotten why we used `SUPPRESS` 😅. I think this does indeed break that change, so I'll need to re-think this.

woodruffw: Hmm, I did some quick tests, and this still works as expected, albeit the `sigstore -v subcommand -v` pattern is still only `verbose=1` (same as it was with https://github.com/sigstore/sigstore-python/commit/f5dbd71df7afb507ea901623173f2f6e3935fb5a). So I _think_ this is fine to merge as-is.

Things I tested:

```
sigstore -v sign   # works, sets DEBUG on sigstore logger
sigstore -vv sign  # works, sets DEBUG on root logger
sigstore sign -v   # works, sets DEBUG on sigstore logger
sigstore sign -vv  # works, sets DEBUG on root logger
```

sigstore/errors.py
woodruffw: (I'll put this in the refactoring issue.)

PR #961, opened 2024-04-05
Prepare 2.1.4 release

### LINKED ISSUE #960: pin securesystemslib in 2.1.x series
Since the next release is a pretty big API change, it might be safest if we pin securesystemslib in 2.1.x branch already. This was already fixed in main by constraining python-tuf to 4.x (as python-tuf is now more strict about securesystemslib).

See comment in #958 for more detail.

I'll do a PR, we can still discuss if this is needed or not

jku:
This is for **series/2.1.x**

Only change is pinning securesystemslib so future major or minor
releases will not be used (current release is 0.31.0)

We can still discuss if this makes sense or is necessary.

Fixes #960

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+## [2.1.4]
+
+## Fixed
+
+* Pinned `securesystemslib` dependency strictly to prevent future breakage
+
 ## [2.1.3]
 
 ## Fixed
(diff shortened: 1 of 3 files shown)


woodruffw:
No objection from me! Future-proofing makes sense.

jku:
I won't 6pm on friday but I can do this on monday. 

Just to make sure I've understood the process:
* merging to the banch does not trigger anything yet
* pushing a tag that looks like a release will trigger a GitHub release
* GitHub release triggers the actual release workflow

woodruffw:
> I won't 6pm on friday but I can do this on monday. 
> 
> 
> 
> Just to make sure I've understood the process:
> 
> * merging to the banch does not trigger anything yet
> 
> * pushing a tag that looks like a release will trigger a GitHub release
> 
> * GitHub release triggers the actual release workflow

Correct except the last part: you have to manually create the GH release from tag, and the release CI takes over from there 🙂

(I typically run `gh release create` locally once the tag is minted.)

PR #956, opened 2024-04-03
sigstore, test: break apart DSSE/artifact sign APIs

woodruffw:
Per https://github.com/sigstore/sigstore-python/issues/628: this turns `sign(...)` into `sign_artifact(...)` and `sign_intoto(...)` respectively. In doing so, we're also able to simplify and deduplicate the internals of the `Signer` class a bit.

PR #962, opened 2024-04-05
Initial DSSE verify APIs

woodruffw:
~~Very WIP.~~

This follows #956, splitting `verify` into `verify_artifact` and `verify_dsse`.

TODO:

- [x] Complete the log entry consistency check for DSSE verification
- [x] Unit tests
- [x] Write some examples


sigstore/verify/verifier.py
```diff
@@ -133,75 +147,137 @@ def verify(
...
+                "cannot perform DSSE verification on a bundle without a DSSE envelope"
+            )
+
+        dsse._verify(bundle.signing_certificate.public_key(), envelope)
+
+        # (8): verify the consistency of the log entry's body against
+        #      the other bundle materials.
+        _ = bundle.log_entry
+
+        # TODO
```
woodruffw, line 252: NB: Consistency for DSSE bundles isn't implemented, pending a better understanding around how the inserted entry's `envelopeHash` is computed.

woodruffw:
I need to do a second pass over this, but DSSE verification is now functional on this branch!

You can use this snippet to test-drive it (using production Sigstore, although replacing the `production()` calls with `staging()` will also work):

```python
import hashlib
import logging

from sigstore.dsse import _StatementBuilder, _Subject
from sigstore.verify import Verifier, policy

logger = logging.getLogger("sigstore")
logger.setLevel("DEBUG")

ctx = SigningContext.production()
stmt = (
    _StatementBuilder()
    .subjects(
        [_Subject(name="null", digest={"sha256": hashlib.sha256(b"").hexdigest()})]
    )
    .predicate_type("https://cosign.sigstore.dev/attestation/v1")
    .predicate(
        {
            "Data": "",
            "Timestamp": "2023-12-07T00:37:58Z",
        }
    )
).build()

with ctx.signer(Issuer.production().identity_token()) as signer:
    bundle = signer.sign(stmt)

verifier = Verifier.production()
verifier.verify_dsse(bundle, policy.UnsafeNoOp())

print(f"OK: integrated at index: {bundle.log_entry.log_index}")
```

sigstore/verify/verifier.py
```diff
@@ -133,75 +144,165 @@ def verify(
...
+        if not (
+            bundle.signing_certificate.not_valid_before_utc
+            <= integrated_time
+            <= bundle.signing_certificate.not_valid_after_utc
+        ):
+            raise VerificationError(
+                "invalid signing cert: expired at time of Rekor entry"
+            )
+
+    def verify_dsse(
```
woodruffw, line 209: NB: This is slightly out of whack with `sign_intoto` in #956, since this operates at the DSSE level while #956 operates at the in-toto level. It might make more sense to turn this into `_verify_dsse` and then expose `verify_intoto` that performs the payload type check as a wrapper on top of `_verify_dsse`.

laurentsimon:
Thanks for putting this out. I've tested in my branch https://github.com/sigstore/model-transparency/pull/112 and it is working. The code currently expects an intoto statement instead of `bytes`, but that's good enough for our demo!

sigstore/verify/verifier.py
```diff
@@ -99,26 +98,38 @@ def staging(cls) -> Verifier:
...
     ) -> None:
         """
-        Public API for verifying.
-
-        `input_` is the input to verify, either as a buffer of contents or as
-        a prehashed `Hashed` object.
-
-        `bundle` is the Sigstore `Bundle` to verify against.
+        Performs the signing certificate verification steps that are shared between
+        `verify_intoto` and `verify_artifact`.
```
jku, line 106:
```suggestion
        `verify_dsse` and `verify_artifact`.
```
?

sigstore/verify/verifier.py
jku: maybe. On the other hand payload handling an application problem with intoto anyway so maybe this is fine?

jku (approved):
I don't pretend to fully understand the DSSE verification details but the changes look correct to me. 

The API also looks roughly right: `verify_artifact()` is not like `verify_intoto/dsse()` so should be separate -- I don't have a strong opinion on which of verify_intoto() and verify_ddse() makes most sense

sigstore/_internal/rekor/__init__.py
```diff
@@ -30,6 +31,26 @@
 __all__ = ["RekorClient", "SignedCheckpoint"]
 
 
+def _dsse_from_parts(cert: Certificate, evp: dsse.Envelope) -> rekor_types.Dsse:
+    signature = rekor_types.dsse.Signature(
+        signature=evp._inner.signatures[0].sig,
```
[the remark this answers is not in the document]
woodruffw: It's an assumption for now, although my understanding is that the same assumption is made/asserted as an invariant by Rekor (i.e., a submitted DSSE entry can only have one signature).

CC @haydentherapper for confirmation 🙂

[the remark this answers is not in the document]
haydentherapper: Rekor allows DSSEs with multiple signatures - https://github.com/sigstore/rekor/blob/main/pkg/types/dsse/v0.0.1/entry.go#L371-L373. intoto v0.0.1 had this [limitation](https://github.com/sigstore/rekor/blob/main/pkg/types/intoto/v0.0.1/entry.go#L230-L231) for this was fixed in intoto v0.0.2 and the dsse kind.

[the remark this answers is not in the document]
woodruffw: Gotcha, maybe I should make this more flexible, then --  is my read correct that the logic here is "all signatures must verify for the entry to be accepted"?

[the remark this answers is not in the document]
haydentherapper: Correct!

sigstore/verify/verifier.py
woodruffw: Yeah, `verify_dsse` seems more appropriate to me ultimately (especially since `verify_intoto` would imply that we're actually verifying the in-toto statement, which is not actually happening).

Conversely, any thoughts on turning `sign_intoto` into `sign_dsse`? The main problem I see with that is that it'll rely on the user passing in an opaque `bytes` to sign, which means they'll also need to pass in an appropriate `payload_type` for the PAE encoding. I can see users doing all kinds of potentially incorrect things there, so maybe keeping signing at the JSON payload + in-toto layer is preferable.

sigstore/_internal/rekor/__init__.py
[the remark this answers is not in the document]
woodruffw: Alright, I've done some tweaking of this in https://github.com/sigstore/sigstore-python/pull/962/commits/e1753ffea3c3376068c13ded0375602f38c993e5 -- we now handle multiple signatures gracefully when they appear in the `Envelope`, although I suspect this will never happen in practice in the Sigstore setting (since the bundle only supports a single signing cert?)

sigstore/verify/verifier.py
jku: I think `sign_dsse()` would work as well if the docs are clear about how to use it for the specific intoto use case... but no strong opinions.

sigstore/_internal/rekor/__init__.py
```diff
@@ -27,7 +27,11 @@
 from .checkpoint import SignedCheckpoint
 from .client import RekorClient
 
-__all__ = ["RekorClient", "SignedCheckpoint"]
+__all__ = [
+    "RekorClient",
+    "SignedCheckpoint",
+    "_hashedrekord_from_parts",
```
jku, line 33: I suppose at this point it should be just `hashedrekord_from_parts()` without underscore. Since `rekor` is private, it does not seem important tough...

jku (approved):
latest changes seems good

sigstore/verify/verifier.py
woodruffw: Sounds good. I'm tempted to make it `sign_dsse` then, but I'll do that with a follow-up PR.

sigstore/_internal/rekor/__init__.py
woodruffw: Yeah, it probably should since it's entirely private APIs. I'll iterate on this with the follow-up PRs for refactoring.

woodruffw:
@laurentsimon @mihaimaruseac FYI we merged here, so you may need to update your refs to `main` if you need these changes for your demos!

PR #963, opened 2024-04-08
Backport slsa upgrade

jku:
**This is for series/2.1.x**

* 2.1.4 release failed
* Fix issue by backporting slsa action update
* Attempt again with 2.1.5

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -105,7 +105,7 @@ jobs:
       contents: write # To add assets to a release.
     # Currently this action needs to be referred by tag. More details at:
     # https://github.com/slsa-framework/slsa-github-generator#verification-of-provenance
-    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@c747fe7769adf3656dc7d588b161cb614d7abfee # v1.10.0
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.10.0
     with:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -105,7 +105,7 @@ jobs:
       contents: write # To add assets to a release.
     # Currently this action needs to be referred by tag. More details at:
     # https://github.com/slsa-framework/slsa-github-generator#verification-of-provenance
-    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.9.0
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@c747fe7769adf3656dc7d588b161cb614d7abfee # v1.10.0
```
woodruffw, line 108: LGTM, but re: comment above it: are we sure referring to it by SHA ref here is okay?

From the workflow's README:

> At present, the GitHub Actions provided in this repository as builders and generators MUST be referenced by tag in order for the slsa-verifier to be able to verify the ref of the trusted builder/generator's reusable workflow. It also needs to be referred as @vX.Y.Z, because the build will fail if you reference it via a shorter tag like @vX.Y or @vX.

woodruffw, line 108: Leaving as a suggestion in case the above is true:

```suggestion
    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.10.0
```

jku: you are right: we need to fix that in main as well

woodruffw (approved):

PR #992, opened 2024-04-29
sigstore: extract LogEntry conversions to their own functions

facutuesca:
This moves the logic that converts between `rekor_v1.TransparencyLogEntry` and `sigstore.models.LogEntry` to their own separate functions:

```python
    @classmethod
    def _from_dict_rekor(cls, dict_: dict[str, Any]) -> LogEntry:
    ...

    def _to_dict_rekor(self, is_message_signature: bool) -> dict[str, Any]:
    ...
```

The motivation for having these available is [PEP 740](https://peps.python.org/pep-0740/), which defines an attestation format for PyPI artifacts. This PEP doesn't define a specific transparency log entry format, rather it leaves the choice to the implementation. In practice, it makes sense to use the exact same structure as a `rekor_v1.TransparencyLogEntry` for these entries. Using a different structure is not necessary and would introduce another format conversion to keep track of.

This means that at some point, in order to verify artifacts using the PEP 740 attestation format, we will have to convert it (back) to a Sigstore bundle. The only API to construct a `Bundle` other than `Bundle.from_json` is `Bundle.from_parts`:

```python
    def from_parts(cls, cert: Certificate, sig: bytes, log_entry: LogEntry) -> Bundle:
```

But since it requires an already constructed `LogEntry`, it can't be used unless you already have one.

The new methods would allow a user of `sigstore-python` to create a `LogEntry` from an existing `rekor_v1.TransparencyLogEntry` without exposing the actual type, just a `dict[str, Any]`. This would look something like:

```python
def pypi_to_sigstore(pypi_attestation: Attestation) -> Bundle:
    ....
    tlog_entry = pypi_attestation.verification_material.transparency_entries[0]

    return Bundle.from_parts(..., log_entry = LogEntry._from_dict_rekor(pypi_attestation....))
```

cc @woodruffw

### DIFF
--- a/sigstore/models.py
+++ b/sigstore/models.py
@@ -253,7 +253,7 @@ class LogEntry:
             tlog_entry.canonicalized_body
         )
         if not isinstance(body_entry, (Hashedrekord, Dsse)):
-            raise ValueError("LogEntry is not of expected type")
+            raise InvalidBundle("log entry is not of expected type")
 
         tlog_entry.kind_version = rekor_v1.KindVersion(
             kind=body_entry.kind, version=body_entry.api_version
--- a/sigstore/models.py
+++ b/sigstore/models.py
@@ -252,9 +252,7 @@ class LogEntry:
         body_entry = TypeAdapter(ProposedEntry).validate_json(
             tlog_entry.canonicalized_body
         )
-        if not isinstance(body_entry, Hashedrekord) and not isinstance(
-            body_entry, Dsse
-        ):
+        if not isinstance(body_entry, (Hashedrekord, Dsse)):
             raise ValueError("LogEntry is not of expected type")
 
         tlog_entry.kind_version = rekor_v1.KindVersion(
--- a/sigstore/models.py
+++ b/sigstore/models.py
@@ -36,10 +36,12 @@ from pydantic import (
     Field,
     StrictInt,
     StrictStr,
+    TypeAdapter,
     ValidationInfo,
     field_validator,
 )
 from pydantic.dataclasses import dataclass
+from rekor_types import Dsse, Hashedrekord, ProposedEntry
 from sigstore_protobuf_specs.dev.sigstore.bundle import v1 as bundle_v1
 from sigstore_protobuf_specs.dev.sigstore.bundle.v1 import (
     Bundle as _Bundle,
@@ -222,7 +224,7 @@ class LogEntry:
             ),
         )
 
-    def _to_dict_rekor(self, is_message_signature: bool) -> dict[str, Any]:
+    def _to_dict_rekor(self) -> dict[str, Any]:
         inclusion_promise: rekor_v1.InclusionPromise | None = None
         if self.inclusion_promise:
             inclusion_promise = rekor_v1.InclusionPromise(
@@ -247,12 +249,17 @@ class LogEntry:
         )
 
         # Fill in the appropriate kind
-        if is_message_signature:
-            tlog_entry.kind_version = rekor_v1.KindVersion(
-                kind="hashedrekord", version="0.0.1"
-            )
-        else:
-            tlog_entry.kind_version = rekor_v1.KindVersion(kind="dsse", version="0.0.1")
+        body_entry = TypeAdapter(ProposedEntry).validate_json(
+            tlog_entry.canonicalized_body
+        )
+        if not isinstance(body_entry, Hashedrekord) and not isinstance(
+            body_entry, Dsse
+        ):
+            raise ValueError("LogEntry is not of expected type")
+
+        tlog_entry.kind_version = rekor_v1.KindVersion(
+            kind=body_entry.kind, version=body_entry.api_version
+        )
 
         tlog_entry_dict: dict[str, Any] = tlog_entry.to_dict()
         return tlog_entry_dict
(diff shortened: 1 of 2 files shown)


woodruffw:
Thanks @facutuesca! Could you add a round-trip test for these? Something simple where we ensure that `entry == from_dict(to_dict(entry))` works would be perfect 🙂

woodruffw:
(Test failures are OK; expected as part of ongoing flakiness from staging.)

facutuesca:
> Thanks @facutuesca! Could you add a round-trip test for these? Something simple where we ensure that `entry == from_dict(to_dict(entry))` works would be perfect 🙂

done!

sigstore/models.py
```diff
@@ -187,6 +186,77 @@ def _from_response(cls, dict_: dict[str, Any]) -> LogEntry:
...
+            log_index=tlog_entry.log_index,
+            inclusion_proof=parsed_inclusion_proof,
+            inclusion_promise=B64Str(
+                base64.b64encode(
+                    tlog_entry.inclusion_promise.signed_entry_timestamp
+                ).decode()
+            ),
+        )
+
+    def _to_dict_rekor(self, is_message_signature: bool) -> dict[str, Any]:
```
woodruffw, line 225: Rather than passing in `is_message_signature`, I think we should infer this from `self.body` :slightly_smiling_face:

facutuesca: Got it. I see that body has the following structure:
```json
{
    "apiVersion": "0.0.1",
    "kind": "hashedrekord",
    "spec": {"..."}
}
```

which corresponds to the [`ProposedEntry`](https://github.com/trailofbits/sigstore-rekor-types/blob/769065354dcd3f8fac175fda6c8dcd0df562dd7b/rekor_types/__init__.py#L127-L130) union type in `sigstore-rekor-types`.

These can be one of several [types](https://github.com/sigstore/rekor/tree/main/pkg/types). Our previous logic said that :
```python
        if isinstance(content, common_v1.MessageSignature):
            inner.message_signature = content
            tlog_entry.kind_version = rekor_v1.KindVersion(
                kind="hashedrekord", version="0.0.1"
            )
        else:
            inner.dsse_envelope = content._inner
            tlog_entry.kind_version = rekor_v1.KindVersion(kind="dsse", version="0.0.1")
```

How would that correspond to the different types that a `ProposedEntry` can be? What I'm asking is which `kind` in the body (`hashedrekord`, `cose`, `alpine`, `dsse`, etc) corresponds to the two kinds we were handling above (`hashedrekord` and `dsse`)

woodruffw: The two are mirrored, but sigstore-python (intentionally) only supports `hashedrekord` and `dsse` for now. So I think what we want here is to fail if we see any type other than those two (which should be impossible, since we should also reject them earlier in the stack).

facutuesca: done!

sigstore/models.py
```diff
@@ -247,12 +249,17 @@ def _to_dict_rekor(self, is_message_signature: bool) -> dict[str, Any]:
...
-                kind="hashedrekord", version="0.0.1"
-            )
-        else:
-            tlog_entry.kind_version = rekor_v1.KindVersion(kind="dsse", version="0.0.1")
+        body_entry = TypeAdapter(ProposedEntry).validate_json(
+            tlog_entry.canonicalized_body
+        )
+        if not isinstance(body_entry, Hashedrekord) and not isinstance(
+            body_entry, Dsse
+        ):
```
woodruffw, line 257: `isinstance` does a logical OR when given multiple types in a tuple 🙂 

```suggestion
        if not isinstance(body_entry, (Hashedrekord, Dsse)):
```

sigstore/models.py
```diff
@@ -187,6 +188,80 @@ def _from_response(cls, dict_: dict[str, Any]) -> LogEntry:
...
+            log_index=self.log_index,
+            log_id=common_v1.LogId(key_id=bytes.fromhex(self.log_id)),
+            integrated_time=self.integrated_time,
+            inclusion_promise=inclusion_promise,
+            inclusion_proof=inclusion_proof,
+            canonicalized_body=base64.b64decode(self.body),
+        )
+
+        # Fill in the appropriate kind
+        body_entry = TypeAdapter(ProposedEntry).validate_json(
```
woodruffw, line 252: Just making sure I understand: we need this `TypeAdapter` because `ProposedEntry` itself is a `typing.Union`, right?

sigstore/models.py
```diff
@@ -187,6 +188,80 @@ def _from_response(cls, dict_: dict[str, Any]) -> LogEntry:
...
+            tlog_entry.canonicalized_body
+        )
+        if not isinstance(body_entry, (Hashedrekord, Dsse)):
+            raise ValueError("LogEntry is not of expected type")
```
woodruffw, line 256:
```suggestion
            raise InvalidBundle("log entry is not of expected type")
```

woodruffw: (I'm okay with raising `InvalidBundle` here, since this will be an invalid bundle for sigstore-python's purposes.)

sigstore/models.py
facutuesca: yeah, exactly (like in this [example](https://docs.pydantic.dev/latest/concepts/unions/#union-validation-errors:~:text=to%20look%20at-,adapter%20%3D%20TypeAdapter(Union%5BDoubledList%2C%20StringsMap%5D),-try%3A%0A%20%20%20%20adapter))

woodruffw (approved):
LGTM, thanks @facutuesca!

PR #993, opened 2024-04-30
Disable staging in tests

### LINKED ISSUE #988: Add a switch to test suite to skip staging tests
I think the decision to mostly test with the staging environment is fine but I suggest we add an easy switch we can flip to skip staging  tests for a while: staging has no SLO so the sort of outages that we're currently seeing are just a part of doing business.

* Add a --skip-staging to conftest.py. I believe we have only a few production tests so I think the easiest way might be to add a "production" marker and then skip tests that have "online" marker but do not have "production" production marker?
* review the production tests: I think we have very few, should make sure at least a sign-verify cycle is done

jku:
* Add ability to disable staging tests in unit tests
* disable staging temporarily because it's currently not reliable

Possibly open issues:
* I have not taken a hard look at whether our coverage is still good enough without the staging tests or not
* `make test` default is not affected: developers need to `TEST_ARGS=--skip-staging  make test`
* If this is merged, need to file a reminder to remove `--skip-staging` at some point

Fixes #988

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
           unshare --map-root-user --net make test TEST_ARGS="--skip-online -vv --showlocals"
 
       - name: test
-        run: make test TEST_ARGS="-vv --showlocals"
+        run: make test TEST_ARGS="-vv --showlocals --skip-staging"
 
       - name: test (interactive)
         if: (github.event_name != 'pull_request') || !github.event.pull_request.head.repo.fork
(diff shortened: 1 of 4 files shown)


jku:
I had to do some refactoring in the fixtures: I think it's correct but I admit the whole pytest fixture design is a little too magical for me...

woodruffw (approved):

woodruffw:
Thanks @jku, LGTM. I agree that the current fixture design is way too magical 😅

woodruffw:
Merging for now; I'll file an issue to revisit this.

PR #997, opened 2024-04-30
test_sign: disable more staging tests

woodruffw:
~~WIP, trying other routes as well.~~

This ensures that tests using `env` to load the staging instance are also skipped by `--skip-staging`.

### DIFF
--- a/test/unit/conftest.py
+++ b/test/unit/conftest.py
@@ -262,9 +262,12 @@ def tuf_dirs(monkeypatch, tmp_path):
 
 @pytest.fixture
 def sign_ctx_and_ident_for_env(
+    pytestconfig,
     env: str,
 ) -> tuple[type[SigningContext], type[IdentityToken]]:
     if env == "staging":
+        if pytestconfig.getoption("skip-staging"):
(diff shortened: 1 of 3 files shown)


test/unit/conftest.py
```diff
@@ -262,9 +262,12 @@ def tuf_dirs(monkeypatch, tmp_path):
 
 @pytest.fixture
 def sign_ctx_and_ident_for_env(
+    pytestconfig,
     env: str,
 ) -> tuple[type[SigningContext], type[IdentityToken]]:
     if env == "staging":
+        if pytestconfig.getoption("skip-staging"):
+            pytest.skip("skipping staging test variant due to --skip-staging")
```
jku, line 270: this should not be needed: the tests that use this will have a "staging" or "production" mark automatically so the skip in pytest_runtest_setup() should happen already -- let me know if you're seeing something else

woodruffw: Whoops, you're completely right. I got tripped up by the "interactive" test stuff.

bobcallaway (approved):

Issue #425, opened 2023-01-11
Fulcio: Switch to new-style claim extensions


woodruffw:
Just filing this for tracking purposes: https://github.com/sigstore/fulcio/pull/945 will change Fulcio's certificate extensions to make them more generic, avoiding unnecessary references to implementation details for e.g. GitHub.

These changes will follow Fulcio's deprecation policy, so no action is immediately required on our part.

woodruffw:
This changes have been fully deployed, so we should begin supporting them.

jku:
I did some quick tests in https://github.com/jku/sigstore-python/commit/50418201b5ade7ddaa7dc2fdd9e86a7bccf04b42:
* backwards compat decision needs to be made: likely we should support verifying already existing certificates that contain github OIDs only, like the test data currently in tree? (my WIP patch does not)
* GitHubWorkflowRepository no longer exists: using Source Repository URI instead is easy and the CLI can paper over the difference
* `GitHubWorkflowName` no longer exists: it probably should never have been used anyway since it has no real claim value -- SAN already contains the workflow path. We may want to expose `Build Signer Digest` as policy instead (?) and the CLI could use "sha" argument as default value for that
* I'm not sure what is the value of `Build Signer URI` -- should the cli use the cert identity argument there too?

woodruffw:
FWIW, I also have an initial stab at this up at #715: it keeps the old extension handling in place while adding new APIs for the new extensions. My thinking was that backwards compatibility could be maintained by using the `AnyOf` building block to compose the old and new extensions.

(More generally, using these new extensions is blocked by the fact that they contain DER encodings for their values, which pyca/cryptography doesn't support for arbitrary third-party extensions yet. That's being tracked here: https://github.com/pyca/cryptography/issues/9283.)

jku:
> FWIW, I also have an initial stab at this up at #715: it keeps the old extension handling in place while adding new APIs for the new extensions. My thinking was that backwards compatibility could be maintained by using the `AnyOf` building block to compose the old and new extensions.

[...]

PR #1004, opened 2024-05-07
sigstore: add new verification policies for missing extensions

facutuesca:
The missing extensions (and their OIDCs) are documented [here](https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#136141572641--fulcio). 

This PR adds an extra base class for verification classes, `_SingleX509ExtPolicyV2` which is inherits from `_SingleX509ExtPolicy`, but expects the extension being verified to have a DER-encoded value (as opposed to just the raw strings). See [here](https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#extension-values) for more details.

This new base class is used for the extensions `1.3.6.1.4.1.57264.1.8` through `1.3.6.1.4.1.57264.1.22`. Also, a new test is added that checks that a GHA-signed artifact can be verified using all (old and new) extensions.

cc @woodruffw 

Closes https://github.com/sigstore/sigstore-python/issues/425.


sigstore/verify/policy.py
```diff
@@ -93,6 +112,41 @@ def verify(self, cert: Certificate) -> None:
             )
 
 
+class _SingleX509ExtPolicyDer(_SingleX509ExtPolicy):
```
woodruffw, line 115: Let's call this V2 or something, since they're both technically DER and Fulcio refers to these as "V2":

```suggestion
class _SingleX509ExtPolicyV2(_SingleX509ExtPolicy):
```

facutuesca: done!

pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
   "cryptography >= 42",
   "id >= 1.1.0",
   "importlib_resources ~= 5.7; python_version < '3.11'",
+  "pyasn1",
```
woodruffw, line 31: Nitpick: we probably don't need a strong constraint here, but this might dodge weirdness where someone tries to drop `sigstore` into a pre-existing env:

```suggestion
  "pyasn1 ~= 0.6",
```

Issue #985, opened 2024-04-25
Set user-agent in all HTTP requests


jku:
**Description**

It would be useful if sigstore-python had a user-agent  in all requests:

- [x] fulcio client (#1008)
- [x] rekor client (#1008)
- [ ] tuf client (I've just added this into python-tuf so we can use it when next release comes)
- [x] oidc (#1008)

Current user-agent is `python-requests/2.31.0`. I think `sigstore-python/X.Y.Z python-requests/2.31.0` (or just `sigstore-python/X.Y.Z`) would make sense.

This can be implemented one client at a time.

woodruffw:
Great idea!

woodruffw:
Completed as of https://github.com/sigstore/sigstore-python/pull/1017, I believe 🙂

PR #1008, opened 2024-05-09
sigstore: uniform user-agent with sigstore version

woodruffw:
This sets the UA in the Fulcio, Rekor, and OIDC clients. I haven't touched the TUF client since I wasn't sure how to access the underlying `requests.Session` there 🙂 

See #985.


jku (approved):
Nice, this will make load balancer logs a little easier to figure out and gives us a tiny bit of telemetry (which versions of sigstore-python are being used in the wild)

I'll handle tuf when 4.1 is released.

PR #1013, opened 2024-05-10
bump sigstore-protobuf-specs

woodruffw:
Bumps us to `~= 0.3.2`. This patch release enables type hinting, allowing us to refine the types a bit here.

### DIFF
--- a/sigstore/dsse.py
+++ b/sigstore/dsse.py
@@ -187,8 +187,7 @@ class Envelope:
         """
         Return a JSON string with this DSSE envelope's contents.
         """
-        # TODO: Unclear why mypy thinks this is returning `Any`.
-        return self._inner.to_json()  # type: ignore[no-any-return]
+        return self._inner.to_json()
 
 
 def _pae(type_: str, body: bytes) -> bytes:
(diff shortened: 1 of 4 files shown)


sigstore/dsse.py
```diff
@@ -217,7 +216,7 @@ def _sign(key: ec.EllipticCurvePrivateKey, stmt: Statement) -> Envelope:
         _Envelope(
             payload=stmt._contents,
             payload_type=Envelope._TYPE,
-            signatures=[Signature(sig=signature, keyid=None)],
+            signatures=[Signature(sig=signature)],
```
jku, line 219: How does this work if keyid is not marked optional in the protobuf -- am I missing something?

jku (approved):
If this works, great. If you can explain how Signature works without keyid, that's appreciated.

woodruffw:
> If you can explain how Signature works without keyid, that's appreciated.

From looking at the proto definition, the keyid field is entirely optional. But the code generator isn't smart enough to understand that optional means that the kwarg is also nullable, so I just removed it outright instead of explicitly setting it to `None` 😅

TL;DR: The generated message is completely unchanged, we're just correctly handling an overly strict type hint now.

Issue #780, opened 2023-09-28
`--cert-identity` should be optional on `sigstore verify github`


woodruffw:
Noticed while dogfooding with @ret2libc: `sigstore verify github` currently **requires** `--cert-identity`, even when the set of flags passed by the user might be sufficient for verification purposes.

This makes `sigstore verify github` invocations unnecessarily wordy, e.g.:

```bash
sigstore verify github action.py --cert-identity https://github.com/sigstore/gh-action-sigstore-python/.github/workflows/release.yml@refs/tags/v2.0.1
```

...when, for many users, this more concise invocation is probably sufficient:

```bash
sigstore verify github action.py --repository sigstore/gh-action-sigstore-python
```

woodruffw:
(This would be a backwards-compatible change, since `--cert-identity` would go from required to optional.)

haydentherapper:
Are these equivalent identifiers? The former is a specific "builder" identity. This aligns to SLSA's model for identity, that trust is rooted in the builder. In terms of trust boundary however, you could argue that a maintainer can modify any workflow in a repository, so it's unnecessary to pin to a specific builder.

Preferring the more narrowly scoped identity, could we specify an additional flag for "releaser" or "builder" like `release.yml@refs/tags/v2.0.1`, and then combine `repository` with this flag to construct the identity?

woodruffw:
Sorry, I completely missed this comment before 😅 

> Are these equivalent identifiers? The former is a specific "builder" identity. This aligns to SLSA's model for identity, that trust is rooted in the builder. In terms of trust boundary however, you could argue that a maintainer can modify any workflow in a repository, so it's unnecessary to pin to a specific builder.

Yeah, I think the argument here would be that, especially in the absence of strongly pinned reusable workflows, there's very little auditable difference between a ref + workflow filename and the repository URI itself -- in both cases you need to resort to checking the exact build commit anyways, since the repository can move both the filename and symbolic `ref` around as it pleases.

(I suspect this is `gh attestation`'s rationale as well, since they allow scoping as high as just the GitHub org without referencing the repo or workflow, etc.)

PR #1015, opened 2024-05-13
cli: allow DSSE verification

### LINKED ISSUE #999: Expose DSSE verification in the CLI
Per #628, we now have DSSE signing and verifying support in the APIs. We haven't, however, exposed it via the CLI, for a few reasons:

1. DSSE verification isn't conceptually tied to an independent input in the way that our current `sigstore verify ...` commands expect: with DSSE the input is embedded in the bundle, cryptographically verified, and then returned for subsequent policy verification. If we want to expose DSSE verification in the CLI, we'll need to decide whether to accomodate that pattern or whether to restrict it (e.g. by requiring a file to verify against, which must either fully match the payload, or must match the in-toto statement's subject digest, or something else).
2. It isn't 100% clear to me if/how we should return the verified DSSE payload: it might be binary, an in-toto statement, or something else. Maybe the CLI could only support in-toto JSON payloads.
3. We need to figure out how to communicate the DSSE verification model to users: `sigstore verify` is conceptually simple ("the file has a valid signature from some identity"), while DSSE introduces the idea of "a bundle is valid if the valid payload within it matches some policy." This might confuse users who conflate "the bundle has a valid signature" with "I don't need to validate the data within the bundle."

Minimally, we could _probably_ get away with leaving `sigstore verify` the way it is, but allowing a limited subset of DSSE bundles within it: ones that have an in-toto payload, which we could then verify the statement within against the digest of the independent input. This would probably cover >90% of use cases, e.g. verifying bundles from GitHub's attestation beta.

woodruffw:
This allows for a _very_ limited form of DSSE verification via the `sigstore` CLI. In particular, it allows a DSSE-containing bundle to be verified, but only if:

* The DSSE payload is JSON;
* The JSON is an in-toto statement;
* The in-toto statement has at least one subject matching the input file's digest.

This limited form is compatible with what GitHub Artifact Attestations is doing and, in particular, is compatible with how `gh attestation verify` behaves.

~~WIP while I clean things up a little.~~

Closes #999.

Closes https://github.com/sigstore/sigstore-python/issues/780.

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -834,7 +834,9 @@ def _verify_github(args: argparse.Namespace) -> None:
         inner_policies.append(
             policy.Identity(
                 identity=args.cert_identity,
-                issuer="https://token.actions.githubusercontent.com",
+                # We always explicitly check the issuer below, so configuring
+                # it here is unnecessary.
+                issuer=None,
             )
         )
     if args.workflow_trigger:
@@ -850,6 +852,13 @@ def _verify_github(args: argparse.Namespace) -> None:
 
     if not inner_policies:
         _die(args, "No verification options supplied")
+
+    # No matter what the user configures above, we require the OIDC issuer to
+    # be GitHub Actions. We add this below the check above, since it doesn't
+    # constitute a sufficient policy check on its own.
+    inner_policies.append(
+        policy.OIDCIssuer("https://token.actions.githubusercontent.com")
+    )
 
     policy_ = policy.AllOf(inner_policies)
 
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -813,31 +813,7 @@ def _verify_identity(args: argparse.Namespace) -> None:
         )
 
         try:
-            # If the bundle specifies a DSSE envelope, perform DSSE verification
-            # and assert that the inner payload is an in-toto statement bound
-            # to a subject matching the input's digest.
-            if bundle._dsse_envelope:
-                with file.open(mode="rb", buffering=0) as io:
-                    digest = sha256_digest(io)
-
-                type_, payload = verifier.verify_dsse(bundle=bundle, policy=policy_)
-                if type_ != dsse.Envelope._TYPE:
-                    raise VerificationError(
-                        f"expected JSON payload for DSSE, got {type_}"
-                    )
-
-                stmt = dsse.Statement(payload)
-                if not stmt._matches_digest(digest):
-                    raise VerificationError(
-                        f"in-toto statement has no subject for digest {digest.digest.hex()}"
-                    )
-            else:
-                verifier.verify_artifact(
-                    input_=hashed,
-                    bundle=bundle,
-                    policy=policy_,
-                )
-
+            _verify_common(verifier, hashed, bundle, policy_)
             print(f"OK: {file}")
         except Error as exc:
             _logger.error(f"FAIL: {file}")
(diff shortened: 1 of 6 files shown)


sigstore/dsse.py
```diff
@@ -97,10 +99,29 @@ def __init__(self, contents: bytes) -> None:
...
 
+    def _matches_digest(self, digest: Hashed) -> bool:
+        """
+        Returns a boolean indicating whether this in-toto Statement contains a subject
+        matching the given digest. The subject's name is **not** checked.
+
+        No digests other than SHA256 are currently supported.
+        """
+        if digest.algorithm != HashAlgorithm.SHA2_256:
+            raise VerificationError(f"unexpected digest algorithm: {digest.algorithm}")
```
woodruffw, line 114: Flagging: unnecessarily brittle, but we don't support anything other than SHA256 at the moment anyways.

sigstore/_cli.py
```diff
@@ -851,13 +847,48 @@ def _verify_github(args: argparse.Namespace) -> None:
...
+            raise VerificationError(
+                f"in-toto statement has no subject for digest {hashed.digest.hex()}"
+            )
+    else:
+        verifier.verify_artifact(
+            input_=hashed,
+            bundle=bundle,
+            policy=policy_,
+        )
+    pass
```
di, line 889:
```suggestion
```

di (dismissed):

sigstore/dsse.py
```diff
@@ -97,9 +99,28 @@ def __init__(self, contents: bytes) -> None:
...
+
+        for sub in self._inner.subjects:
+            sub_digest = sub.digest.root.get("sha256")
+            if sub_digest is None:
```
segiddins, line 118: why is this here?

segiddins: oh, it's because we only care about a single matching sha256

jku (commented):
DSSE parts look fine. The `verify github` change seems a bit fishy in that we no longer make sure the issuer is github -- maybe we should change `policy.Identity` to accept `identity: str | None, issuer: str` as args and then use that in _verify_github, or alternatively have a separate `policy.Issuer` that only checks the issuer?

woodruffw:
> DSSE parts look fine. The `verify github` change seems a bit fishy in that we no longer make sure the issuer is github -- maybe we should change `policy.Identity` to accept `identity: str | None, issuer: str` as args and then use that in _verify_github, or alternatively have a separate `policy.Issuer` that only checks the issuer?

Ah yeah, good catch -- I think we can re-use the already existing `policy.OIDCIssuer` for this 🙂

sigstore/_cli.py
```diff
@@ -846,18 +850,62 @@ def _verify_github(args: argparse.Namespace) -> None:
...
 
+    if not inner_policies:
+        _die(args, "No verification options supplied")
+
+    # No matter what the user configures above, we require the OIDC issuer to
+    # be GitHub Actions. We add this below the check above, since it doesn't
+    # constitute a sufficient policy check on its own.
+    inner_policies.append(
+        policy.OIDCIssuer("https://token.actions.githubusercontent.com")
+    )
```
woodruffw, line 861: Flagging: right now we allow any of the CLI options to be a "sufficient" option for verification with `sigstore verify github`, which is potentially misleading or insufficient. 

For example, a user could do `sigstore verify github ... --name foo`, which would only check the workflow's name and allow *any* GitHub user/org and repo, which probably isn't what they intended.

Given that, should be restrict this to a few sufficient combinations? Some ideas:

* `--cert-identity`, as before, to bind the full `user/repo` and workflow path
* `--repository`, to bind the `user/repo`
* Others, from the "V2" extensions?

woodruffw: Tweaked this with https://github.com/sigstore/sigstore-python/pull/1015/commits/41d3fc96717d659d9516c9c2d4e16fcb3366aa91 -- we now require `--cert-identity` OR `--repository`, which should be misuse-resistant without being burdensome.

woodruffw: (This is more or less consistent with what `gh attestation verify` does, although they also allow just `--owner`)

jku: 41d3fc9 seems quite reasonable.

jku (approved):
Looks great, thanks for all the work on this. Reasoning for required options seems good to me. 

Suggestions for docs (could file additional issues for these): 
* The README examples of github verification are now not completely up to date: we should add a `--repository` example (maybe as the first one)
* there is no "body text" in `sigstore verify github --help`: A description that mentions the required flags would make sense

PR #1017, opened 2024-05-14
Upgrade tuf dependency

jku:
The ngclient API that we use is fully backwards-compatible but I want to do two small changes so am bumping the required version:
* Don't poke into fetcher internals: the timeout is now the 30 seconds we wanted
* set user-agent through UpdaterConfig: this will end up as "sigstore-python/3.0.0rc2 tuf/5.0.0 python-requests/2.31.0" (a bit verbose perhaps but all of that is useful)


woodruffw (approved):

woodruffw:
Thanks @jku!

PR #1025, opened 2024-05-17
release: remove pip cache usage

woodruffw:
This is purely a defense in depth maneuver: an attacker could only poison our cache here by running arbitrary code on `main`. But we probably shouldn't be caching anyways, since it makes the release process/state more opaque 🙂 

Context: https://adnanthekhan.com/2024/05/06/the-monsters-in-your-build-cache-github-actions-cache-poisoning/

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,9 +21,9 @@ jobs:
 
       - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
         with:
+          # NOTE: We intentionally don't use a cache in the release step,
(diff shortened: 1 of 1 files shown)


jku (approved):
That makes sense.

There's several other things we could do to make the release more auditable and reproducible:
* smoke test should happen in a separate job -- venv is not a security barrier so all our runtime deps are now in a way release process dependencies as well
* preferably build and the actual build plugin would be pinned... this is a bit annoying to do though (because the build plugin is installed inside build, see [example](https://github.com/theupdateframework/python-tuf/blob/develop/.github/workflows/cd.yml#L30-L35))

PR #1039, opened 2024-06-05
dsse: add Envelope._from_json

woodruffw:
Adds `_from_json` and `__eq__` to Envelope. 

`_from_json` is currently a private API, since there have been no public third-party requests for it yet. But it could be made public in a minor release without any problems.


jku (approved):
:+1:

PR #1054, opened 2024-06-28
Add functionality to sign DSSE envelopes with arbitrary payloads

### LINKED ISSUE #982: Support for DSSE
The current API supports only intoto statement, and we'd like support for arbitrary DSSE blobs.

We discussed this in the past but the original DSSE issue has been closed, so creating this issue for tracking.

susperius:
#### Summary
Added functionality to sign DSSE envelopes with arbitrary payloads. This is to ensure model signing can in the future use sigstore with its own manifest format inside of a DSSE envelope.

Closes #982 

#### Release Note

* sign.py: `sign_dsse_envelope` takes an envelope as input value and adds a signature for its contents to the envelope
* dsse.py: `Envelope` provides a public classmethod to create an envelope based on an arbitrary payload type and payload

### DIFF
--- a/sigstore/dsse.py
+++ b/sigstore/dsse.py
@@ -210,6 +210,23 @@ class Envelope:
         inner = _Envelope().from_json(contents)
         return cls(inner)
 
+    @classmethod
+    def from_payload(cls, payload_type: str, payload: bytes) -> Envelope:
+        """Return an unsigned DSSE envelope.
+
+        Args:
+            payload_type (str): The envelope's payload type
+            payload (bytes): The envelope's payload
+
+        Returns:
+            Envelope: An unsigned DSSE envelope
+        """
+        inner = _Envelope(
+            payload=payload,
+            payload_type=payload_type,
+        )
+        return cls(inner)
+
     def to_json(self) -> str:
         """
         Return a JSON string with this DSSE envelope's contents.
@@ -254,6 +271,18 @@ def _sign(key: ec.EllipticCurvePrivateKey, stmt: Statement) -> Envelope:
             signatures=[Signature(sig=signature)],
         )
     )
+
+
+def _sign_envelope(
+        key: ec.EllipticCurvePrivateKey, envelope: Envelope) -> Envelope:
+    """
+    Sign the given envelope's payload and set the signature field
+    with the generated signature.
+    """
+    pae = _pae(envelope._inner.payload_type, envelope._inner.payload)
+    signature = key.sign(pae, ec.ECDS(hashes.SHA256()))
+    envelope._inner.signaures = [Signature(sig=signature)]
+    return envelope
 
 
 def _verify(key: ec.EllipticCurvePublicKey, evp: Envelope) -> bytes:
--- a/sigstore/dsse.py
+++ b/sigstore/dsse.py
@@ -20,6 +20,7 @@ Functionality for building and manipulating in-toto Statements and DSSE envelope
 from __future__ import annotations
 
 import logging
+from dataclasses import dataclass
 from typing import Any, Dict, List, Literal, Optional, Union
 
 from cryptography.exceptions import InvalidSignature
@@ -187,6 +188,17 @@ class _StatementBuilder:
         return Statement(stmt.model_dump_json(by_alias=True).encode())
 
 
+@dataclass
+class RawPayload:
+    """
+    Represents a raw payload.
+
+    This type can be signed and wrapped into a `Envelope`.
+    """
+    type: str
+    data: bytes
+
+
 class Envelope:
     """
     Represents a DSSE envelope.
@@ -212,7 +224,9 @@ class Envelope:
         return cls(inner)
 
     @classmethod
-    def from_payload(cls, payload_type: str, payload: bytes) -> Envelope:
+    def _from_payload(
+        cls, payload: RawPayload, sigs: list[Signature]
+    ) -> Envelope:
         """Return an unsigned DSSE envelope.
 
         Args:
@@ -223,8 +237,9 @@ class Envelope:
             Envelope: An unsigned DSSE envelope
         """
         inner = _Envelope(
-            payload=payload,
-            payload_type=payload_type,
+            payload=payload.data,
+            payload_type=payload.type,
+            signatures=sigs,
         )
         return cls(inner)
 
@@ -274,16 +289,17 @@ def _sign(key: ec.EllipticCurvePrivateKey, stmt: Statement) -> Envelope:
     )
 
 
-def _sign_envelope(
-        key: ec.EllipticCurvePrivateKey, envelope: Envelope) -> Envelope:
+def _sign_payload(
+        key: ec.EllipticCurvePrivateKey, payload: RawPayload) -> Envelope:
     """
     Sign the given envelope's payload and set the signature field
     with the generated signature.
     """
-    pae = _pae(envelope._inner.payload_type, envelope._inner.payload)
-    signature = key.sign(pae, ec.ECDS(hashes.SHA256()))
-    envelope._inner.signaures = [Signature(sig=signature)]
-    return envelope
+    pae = _pae(payload.payload_type, payload.payload)
+    signature = key.sign(pae, ec.ECDSA(hashes.SHA256()))
+    return Envelope._from_payload(
+        payload_=payload,
+        sigs=[Signature(sig=signature)])
 
 
 def _verify(key: ec.EllipticCurvePublicKey, evp: Envelope) -> bytes:
--- a/sigstore/dsse.py
+++ b/sigstore/dsse.py
@@ -1,4 +1,5 @@
 # Copyright 2022 The Sigstore Authors
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


sigstore/sign.py
```diff
@@ -229,6 +230,45 @@ def sign_dsse(
 
         return self._finalize_sign(cert, content, proposed_entry)
 
+    def sign_dsse_envelope(
```
woodruffw, line 233: I'm a little wary of a new top-level sign API, especially one that's so close to `sign_dsse` -- instead of making this a new function, what do you think about making `sign_dsse` take `Statement | Envelope`?

From there, we'd need to improve the docs. But I think that might be a little cleaner/easier to maintain long-term 🙂

sigstore/sign.py
```diff
@@ -229,6 +230,45 @@ def sign_dsse(
...
+                # behaviors/bugs (one pydantic, one datamodel-codegen):
+                # See: <https://github.com/pydantic/pydantic/discussions/7418#discussioncomment-9024927>
+                # See: <https://github.com/koxudaxi/datamodel-code-generator/issues/1903>
+                proposed_content=rekor_types.dsse.ProposedContent(  # type: ignore[call-arg]
+                    envelope=envelope.to_json(),
+                    verifiers=[b64_cert.decode()],
+                ),
+            ),
+        )
+        return self._finalize_sign(cert, envelope, proposed_entry)
```
woodruffw, line 270: Making sure I'm not missing anything: does this actually work yet? My understanding was that Rekor doesn't yet support arbitrary DSSE envelopes, since it expects a valid in-toto statement within the envelope.

CC @haydentherapper

sigstore/dsse.py
```diff
@@ -256,6 +274,18 @@ def _sign(key: ec.EllipticCurvePrivateKey, stmt: Statement) -> Envelope:
...
 
 
+def _sign_envelope(
+        key: ec.EllipticCurvePrivateKey, envelope: Envelope) -> Envelope:
+    """
+    Sign the given envelope's payload and set the signature field
+    with the generated signature.
+    """
+    pae = _pae(envelope._inner.payload_type, envelope._inner.payload)
+    signature = key.sign(pae, ec.ECDS(hashes.SHA256()))
```
woodruffw, line 284: Typo?

```suggestion
    signature = key.sign(pae, ec.ECDSA(hashes.SHA256()))
```

sigstore/dsse.py
```diff
@@ -210,6 +211,23 @@ def _from_json(cls, contents: bytes | str) -> Envelope:
         inner = _Envelope().from_json(contents)
         return cls(inner)
 
+    @classmethod
+    def from_payload(cls, payload_type: str, payload: bytes) -> Envelope:
```
woodruffw, line 215: Like elsewhere: we should start with this being a private API, before we make it public.

```suggestion
    def _from_payload(cls, payload_type: str, payload: bytes) -> Envelope:
```

woodruffw (commented):
Thanks @susperius!

Did an initial pass -- structure LGTM overall, but I want to confirm that this is actually supported by the Sigstore PGI before moving forwards 🙂

sigstore/sign.py
haydentherapper: That is correct, https://github.com/sigstore/rekor/blob/main/pkg/types/dsse/v0.0.1/entry.go#L112-L146, also pointed out in https://github.com/sigstore/sigstore-python/issues/982#issuecomment-2108913284.

The reason is so that Rekor can extract index keys from the statement. We can add other payload types.

woodruffw: Thanks @haydentherapper! I was feeling some deja vu at this 😅 

Given the above, I'm not sure we can merge this/it makes sense to merge until Rekor supports other payload types (since users who attempt to use it will hit a hard error and think that sigstore-python is broken).

(Another option here would be to use a DSSE envelope with an arbitrary payload, but to emit a `hashedrekord` entry instead. I have no strong opinion on whether this is a good idea or not, but it'll likely cause a lot of incompatibilities with the other clients in the ecosystem.)

sigstore/dsse.py
susperius: Sorry! I thought the test ran but apparently it didn't.

sigstore/dsse.py
susperius: It is the only "valid" way to create an envelope for users (except of just calling the private method).
Surely, I can change this but just like to get your thoughts on how otherwise a user can sign an envelope.

sigstore/sign.py
susperius: Oh ok, I missed that. Sorry!
We could log a warning when an arbitrary DSSE envelope is used and skip the Rekor  post in finalize sign.
Would that make sense or would it mess up the verification itself?

susperius:
> Thanks @susperius!
> 
> Did an initial pass -- structure LGTM overall, but I want to confirm that this is actually supported by the Sigstore PGI before moving forwards 🙂

Thank you for looking into this!

sigstore/sign.py
woodruffw: > Would that make sense or would it mess up the verification itself?

That would mess up the verification itself, unfortunately -- that would result in a bundle with no transparency log entries, which would both be invalid and also incompatible with other clients.

sigstore/dsse.py
woodruffw: Hmm, yeah. The problem here is that `from_payload` is a little misleading as an API: it produces an `Envelope`, but that `Envelope` is actually in an invalid state (since it's missing signatures).

Instead of passing an incomplete `Envelope` in for signing, maybe we could adapt the top-level `sign_dsse` signature:

```python
    def sign_dsse(
        self,
        input_: dsse.Statement | tuple[str, bytes],
    ) -> Bundle:
```

...where `tuple[str, bytes]` is the envelope type and arbitrary contents. That can then be passed into `dsse._sign` with some similar small tweaks.

(I still don't _love_ that signature, but I think it's a better starting point and avoids the need for a partial constructor here 🙂)

sigstore/sign.py
haydentherapper: Even if we made index keys optional for unparseable DSSEs, this would be a breaking change for log monitors that expect an intoto payload in order to extract subject values.

sigstore/dsse.py
susperius: How about using a dataclass?

```python
@dataclass
class DSSEPayload:
    payload_type: str
    payload: bytes
```

I think at some point the `sign_dsse` might need to be renamed :) but for now a dataclass would make it clearer for users.

WDYT?

woodruffw: That works for me, although maybe `dsse.RawPayload` to make it clear that this is the "raw" variant of possible payload types (`dsse.Statement` being another valid variant).

sigstore/sign.py
susperius: How could we proceed with this then?

* Would it be possible to use the SHA256 of the payload as index key for arbitrary payloads?
* Should we define our DSSE envelope payload type before and then add it to rekor?  <- this wouldn't be a solution for arbitrary payloads but helps the model signing effort.

sigstore/dsse.py
susperius: PTAL

sigstore/sign.py
```diff
@@ -211,22 +212,26 @@ def sign_dsse(
...
+                        verifiers=[b64_cert.decode()],
+                    ),
                 ),
-            ),
-        )
-
+            )
+        elif type(input_) is dsse.RawPayload:
+            content = dsse._sign_payload(self._private_key, input_)
+            # TODO: figure out an entry that works.
```
woodruffw, line 234: NB: this is still a hard blocker: the client specification requires bundles to have a transparency log entry, so we can't just skip it when the signed-over input isn't a DSSE.

sigstore/dsse.py
```diff
@@ -256,6 +289,19 @@ def _sign(key: ec.EllipticCurvePrivateKey, stmt: Statement) -> Envelope:
     )
 
 
+def _sign_payload(
```
woodruffw, line 292: Rather than a new `_sign_payload` function, can we make `_sign` take `Statement | RawPayload`? That'll require some internal checks but will be a bit shorter 🙂

sigstore/dsse.py
```diff
@@ -210,6 +223,26 @@ def _from_json(cls, contents: bytes | str) -> Envelope:
...
+
+        Returns:
+            Envelope: An unsigned DSSE envelope
+        """
+        inner = _Envelope(
+            payload=payload.data,
+            payload_type=payload.type,
+            signatures=sigs,
+        )
+        return cls(inner)
```
woodruffw, line 244: Merging `_sign_payload` with `_sign` should eliminate the need for this new helper 🙂

sigstore/sign.py
woodruffw: I think there are two possible resolutions here, both of which require some kind of upstream change:

* Option 1: Rekor learns about non-JSON DSSE payloads. I don't know what the timeline for this would be 🙂 
* Option 2: The client spec clarifies/expresses a judgement about using `dsse` bundles with non-`dsse` Rekor entry types (such as `hashedrekord`, which would work fine here *if* the DSSE envelope is canonicalized).

sigstore/sign.py
susperius: @haydentherapper I meant to bring your focus back to this by re-triggering the review ;)

sigstore/sign.py
susperius: In case it helps: the model signing use case would contain JSON as payload. However, as things stand now rekor solely supports in-toto statements.

PR #1068, opened 2024-07-18
tuf, trust: allow custom TUF base dir

### LINKED ISSUE #1067: Allow TUF initialization outside of `$HOME`
We currently assume `$HOME` exists and write to cache/local directories within it:

https://github.com/sigstore/sigstore-python/blob/8b06ae988495317665e6ce58124d49c4d3f887da/sigstore/_internal/tuf.py#L40-L55

This is not super reliable: some newer versions of Debian (and probably other distros) create system accounts without a home directory by default, meaning that this causes extremely opaque errors like:

```
[Errno 13] Permission denied: '/nonexistent'       
```

To work around this, we should allow API users to configure a different directory to initialize the TUF metadata and targets directories within. We could also potentially expose this as a CLI option, although the benefits there are more slight (since most CLI users should be operating within functional user environments).

woodruffw:
This adds `base_dir` kwargs where appropriate, trickling down to the TUF `TrustUpdater`'s initialization. When supplied, TUF will initialize into appropriate subdirectories within `base_dir`, rather than attempting to initialize under the current `$HOME`.

~~Needs tests.~~

Closes #1067.

### DIFF
--- a/sigstore/_internal/trust.py
+++ b/sigstore/_internal/trust.py
@@ -268,38 +268,45 @@ class TrustedRoot:
         cls,
         url: str,
         offline: bool = False,
+        base_dir: Path | None = None,
     ) -> TrustedRoot:
         """Create a new trust root from a TUF repository.
 
-        If `offline`, will use trust root in local TUF cache. Otherwise will
-        update the trust root from remote TUF repository.
+        If `offline`, TUF will use the trust root in the local cache. Otherwise
+        TUFF will update the trust root from remote TUF repository.
+
+        If `base_dir` is given, TUF will use it for appropriate local
+        subdirectories. If not given, appropriate subdirectories within
+        the user's data and cache directories will be used instead.
         """
-        path = TrustUpdater(url, offline).get_trusted_root_path()
+        path = TrustUpdater(
+            url, offline=offline, base_dir=base_dir
+        ).get_trusted_root_path()
         return cls.from_file(path)
 
     @classmethod
     def production(
         cls,
         offline: bool = False,
+        base_dir: Path | None = None,
     ) -> TrustedRoot:
         """Create new trust root from Sigstore production TUF repository.
 
-        If `offline`, will use trust root in local TUF cache. Otherwise will
-        update the trust root from remote TUF repository.
+        See `from_tuf` for the behavior of `offline` and `base_dir`.
         """
-        return cls.from_tuf(DEFAULT_TUF_URL, offline)
+        return cls.from_tuf(DEFAULT_TUF_URL, offline=offline, base_dir=base_dir)
 
     @classmethod
     def staging(
         cls,
         offline: bool = False,
+        base_dir: Path | None = None,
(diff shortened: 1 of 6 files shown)


woodruffw:
NB: This is not part of a public API yet.

woodruffw:
This is good to go in terms of private API changes. I'm not 100% sure how best to expose this as a public API yet: we could either plumb new kwargs into `SigningContext.production()` and `Verifier.production()` (etc.) _or_ make `TrustedRoot` into a public API.

The latter might make more sense, given that `trusted_root` is also a public parameter in APIs like `SigningContext.__init__`.

woodruffw:
> Do we _need_ to expose this as a public API? Or can we just pick a sufficient directory from within this library instead?

The tricky part is picking a sufficient directory 🙂 -- right now we pick directories under `$HOME` by default, but that's causing problems for PyPI (because `$HOME` is `/nonexistent` for default system accounts). 

The other option would be to choose a consistent name under `$TMPDIR` or similar, but I wasn't sure about that. How does that sound to you?

di:
> The other option would be to choose a consistent name under `$TMPDIR` or similar, but I wasn't sure about that. How does that sound to you?

Yeah, this is what I was imagining.

woodruffw:
Sounds good, I'll change to that!

(With that, it may not even make sense to use `platformdirs` any more at all -- everything can go in `$TMPDIR`.)

di:
Another thought: we could just set or recommend setting the `$XDG_` environment variables for anyone that needs to configure this: https://platformdirs.readthedocs.io/en/latest/api.html#unix-linux

I _think_ that should work without requiring any changes here at all?

woodruffw:
> Another thought: we could just set or recommend setting the `$XDG_` environment variables for anyone that needs to configure this: [platformdirs.readthedocs.io/en/latest/api.html](https://platformdirs.readthedocs.io/en/latest/api.html#unix-linux)
> 
> I _think_ that should work without requiring any changes here at all?

I think that would work, but for use in PyPI that might be a little funky: I think we'd need to run the whole `warehouse` WSGI application under those `XDG_*` settings, or just inject them for that one critical bit. That might result in some surprising behavior with the other dependencies (but I don't actually know that, just speculating 🙂)

woodruffw:
Thinking this through a little more: `$TMPDIR` will work some of the time, but poses some problems:

1. On a multi-user system, we need to make sure not to accidentally clobber/share TUF repos between users. We could do this by keying the `$TMPDIR` path with `os.getuid` or similar, but this will require a workaround for Windows.
2. Using `$TMPDIR` for this safely requires making sure everything is `0700`, which shouldn't be _hard_ but adds another layer of security considerations.

In the context of Warehouse neither of these is a serious issue (since it's single user + rootless), but it might be for other installers.

So yeah, I think the best way to go here is to create `/tmp/share` and `/tmp/cache` in the Warehouse container and redirect `XDG_*` to those. That should be minimally invasive and require no changes here.

PR #1077, opened 2024-07-31
API: make _StatementBuilder public

### LINKED ISSUE #1076: Convert in-toto `Statement` to Sigstore DSSE Statement via a single call
**Description**

For model signing, we're producing in-toto statements for serialization of a model (e.g., https://github.com/sigstore/model-transparency/pull/266 -- if looking at it before merging, look at the last commit only). When signing, we need to convert [in-toto Statements](https://github.com/in-toto/attestation/blob/fc43990bbbff21b78d5e1d3aa419b393a7ef893d/python/in_toto_attestation/v1/statement.py#L9) to [Sigstore Statements](https://github.com/sigstore/sigstore-python/blob/575124adaaa42f70230e0f9d355bb7fb358fa848/sigstore/dsse.py#L81)

One option is to serialize the in-toto one to bytes and then use the Sigstore one's constructor to unparse. But this requires too much data transformations to and from strings, so I'd avoid it if possible.

The other is to use [`_StatementBuilder`](https://github.com/sigstore/sigstore-python/blob/575124adaaa42f70230e0f9d355bb7fb358fa848/sigstore/dsse.py#L133). But this is private -- not a big issue, we could make it public. But, it still [dumps a json to then pass the results to the constructor](https://github.com/sigstore/sigstore-python/blob/575124adaaa42f70230e0f9d355bb7fb358fa848/sigstore/dsse.py#L176-L186).

What I'd like to propose is a `classmethod` on `Statement` that would just convert directly from in-toto Statement, if that would be possible.

woodruffw:
Adds the public `StatementBuilder` API, created by making `_StatementBuilder` public. Also adds an alternative construction route for `Statement`, allowing it to be directly constructed from an internal (still private) `_Statement`, avoiding an unnecessary serde round-trip.

Note: This makes `StatementBuilder` public, but explicitly doesn't commit it to SemVer just yet; I don't expect it to change very much, but in the event it does I don't want us locked into design decisions for the entire 3.x lifetime 🙂 

Closes #1076.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ All versions prior to 0.9.0 are untracked.
 ### Added
 
 * API: `dsse.StatementBuilder` has been added. It can be used to construct an
-  in-toto `Statement` for subsequent enveloping and signing
+  in-toto `Statement` for subsequent enveloping and signing.
+  This API is public but is **not considered stable until the next major
+  release.**
   ([#1077](https://github.com/sigstore/sigstore-python/pull/1077))
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+### Added
+
+* API: `dsse.StatementBuilder` has been added. It can be used to construct an
+  in-toto `Statement` for subsequent enveloping and signing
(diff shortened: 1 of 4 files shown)


woodruffw:
CC @mihaimaruseac for visibility 🙂

mihaimaruseac (approved):
Thank you! This looks exactly what we'd need for model_transparency

di (approved):

PR #1089, opened 2024-08-08
add `fix-bundle` plumbing command

### LINKED ISSUE #1088: Some 0.1 bundles fail to verify
See https://github.com/python/cpython/issues/122785.

I did some preliminary triage on this: the cause of the problem is that some `v0.1` bundles contain an `inclusionProof` without a `checkpoint`, which represents the source of signed time from the transparency log. This happened because versions of `sigstore-python` before 2.x didn't include the `checkpoint` in the bundle, as the `inclusionPromise.signedEntryTimestamp` was considered sufficient as a source of signed time.

We became more strict about this starting with 3.x, as part of requiring an inclusion check that's stronger than just a promise. 

To fix this, we have a couple of options:

1. Synchronous: we can relax our signed time requirements to allow inclusion promises to be sufficient again. This would allow the current bundles on `python.org` to be verified as-is, at the cost of making bundle verification malleable w/r/t an unauthenticated field (the bundle's version)
    * The severity of this malleability is debatable (in this case both the proof and promise are signed by the same Rekor keys), but it's arguably a bad precedent to set and could become a more serious security problem in the future. Verification malleability like this is a classic footgun 🙂 
2. Synchronous: we could perform an online lookup of the checkpoint, to allow inclusion proof verification. This would allow the current bundles on `python.org` to be verified as-is, at the cost of a network roundtrip (i.e., these old bundles could not be verified when the user explicitly sets `--offline`).
3. Asynchronous: we could perform a one-time update on `python.org`'s bundles, to add the missing `checkpoint` field to each. This would **not** change the signature or any other cryptographic components, since it's purely an envelope level change, but it _would_ require updating each.
    * To accomplish this, we could pursue the basic idea in https://github.com/sigstore/sigstore-python/issues/718 and implement a `sigstore plumbing bundle upgrade` (or similar) command that takes a bundle as input and updates to the latest version + fixes any inconsistencies.

CC @sethmlarson

woodruffw:
~~WIP; needs documentation and feedback.~~

This command enables users to fix older bundles
that were malformed by older versions of sigstore-python.

This changeset also includes a new integration test hierarchy, with a starter test for the new `fix-bundle` command based on a publicly observed malformed bundle.

CC @sethmlarson 

Closes #1088.

See https://github.com/python/cpython/issues/122785.


haydentherapper:
Just to cross reference, we were thinking about building a similar capability in https://github.com/sigstore/cosign/issues/3794, to upgrade a bundle.

sethmlarson:
Confirming that I was able to get Sigstore CLI to verify all existing CPython Sigstore bundles if this process is run on them.

di (approved):

PR #1110, opened 2024-09-05
Add minimum version to interrogate dependency

facutuesca:
Before, `setuptools` was included in the `lint` dependencies due to a bug in `interrogate`. Since that bug was [fixed](https://github.com/econchick/interrogate/issues/164#issuecomment-2041231552), we remove `setuptools` and add a constraint to the `interrogate` version to ensure we use a version with the fix.

cc @woodruffw


woodruffw (approved):

woodruffw:
Thanks @facutuesca!

Issue #1111, opened 2024-09-05
CLI: Full DSSE support


woodruffw:
Right now, the `sigstore` CLI only has limited support for DSSE bundles: we support verifying them in `sigstore verify`, but not signing/generating via the CLI.

Some things we'll want to do:

- [x] #1115: We should support DSSE bundle generation and signing via the CLI. This will require some design thought on how best to integrate this with the current `sigstore sign` subcommand; two ideas:
    * `sigstore sign --attest` enables DSSE generation/signing, rather than the default `hashedrekord`
    * A new subcommand hierarchy, like `sigstore attest ...`, to mirror `cosign attest`
- [x] #1116: When verifying DSSE bundles with `sigstore verify ...` subcommands, we should support emitting the bundle's interior statement in a structured manner, for further processing.
    * One idea for doing this is to add a `--format=json` or similar flag across all of the verification subcommands, and have the subcommands emit only JSON in that case, e.g. `{"valid": true, "statement": {...}}`.
- [x] #1124: Generally improve our integration-level testing of the sigstore CLI

CC @facutuesca

facutuesca:
I see that DSSE supports multiple artifacts (“subjects”) inside the envelope, each with a name and digest. But the current `sigstore sign` (non-DSSE) command will, given multiple input files, generate multiple bundle files, one for each input file.

Do we want the same behavior when using DSSE? That is, multiple input files -> multiple output bundles, even if a single DSSE envelope supports having multiple subjects? Or do we want to output a single sigstore bundle, that contains the DSSE envelope with multiple subjects?

woodruffw:
> Do we want the same behavior when using DSSE? That is, multiple input files -> multiple output bundles, even if a single DSSE envelope supports having multiple subjects? Or do we want to output a single sigstore bundle, that contains the DSSE envelope with multiple subjects?

The former -- I think we want to maintain the 1:1 relationship between input and output files for now, even though DSSE bundles technically allow a 1:N relationship. My rationale for that is that fewer clients support that 1:N pattern, and the 1:1 pattern is more immediately applicable to use cases like Python packaging 🙂

facutuesca:
Do we want to use the same default predicate type for the in-toto statement as `cosign`? i.e: `https://cosign.sigstore.dev/attestation/v1`

woodruffw:
[...]

PR #1115, opened 2024-09-10
Add `sigstore attest` CLI subcommand to sign using DSSE envelopes

facutuesca:
#### Summary
Part of https://github.com/sigstore/sigstore-python/issues/1111. 
Adds a `sigstore attest` CLI subcommand to sign using DSSE. The command is very similar to `sigstore sign`, but it takes two new options:
- `predicate-type`: One of `[ https://slsa.dev/provenance/v0.2, https://slsa.dev/provenance/v1 ]`. The predicate type for the in-toto statement being signed over.
- `predicate`: The path to the file containing the predicate JSON. The predicate's contents must match the predicate type specified in `predicate-type`.

```sh
sigstore attest --predicate-type https://slsa.dev/provenance/v0.2 --predicate ./path/to/predicate.json  FILE
```

Since the logic for both `sigstore sign` and `sigstore attest` is mostly the same, this PR extracts that shared logic into a separate function `_sign_common()`, which is then called by `_sign()` and `_attest()` in order to not duplicate that logic.

In order to test it, I have uploaded a SLSA v0.2 provenance predicate [here](https://gist.github.com/facutuesca/8568c176dc1978debf128b882b3c4bcf), originally generated by GitHub for [this](https://github.com/sigstore/sigstore-python/releases/tag/v3.2.0) `sigstore-python` release.
Once downloaded, it can be used to sign over a file with the following command:

```sh
sigstore attest --predicate-type https://slsa.dev/provenance/v0.2 --predicate ./path/to/predicate.json  FILE
```

#### Release Note

* CLI: The `sigstore attest` subcommand has been added. This command is
  similar to `cosign attest` in that it signs over an artifact and a
  predicate using a DSSE envelope. This commands requires the user to pass
  a path to the file containing the predicate, and the predicate type.
  Currently only the SLSA Provenance v0.2 and v1.0 types are supported.

#### Documentation
Updated the README

cc @woodruffw


sigstore/_cli.py
```diff
@@ -228,6 +238,102 @@ def _parser() -> argparse.ArgumentParser:
...
+        "--predicate",
+        metavar="FILE",
+        type=Path,
+        required=True,
+        help="Path to the predicate file",
+    )
+    dsse_options.add_argument(
+        "--predicate-type",
+        metavar="TYPE",
+        choices=PREDICATE_TYPES_CLI_MAP,
```
woodruffw, line 267: Out of curiosity: do other clients use these shorthand names? I'm wondering if we should just use the predicate URLs directly as choices here -- that makes the CLI slightly less ergonomic, but avoids adding one more custom convention 🙂

sigstore/_cli.py
```diff
@@ -228,6 +238,102 @@ def _parser() -> argparse.ArgumentParser:
...
+            "Write default outputs to the given directory (conflicts with --signature, --certificate"
+            ", --bundle)"
+        ),
+    )
+    output_options.add_argument(
+        "--overwrite",
+        action="store_true",
+        default=_boolify_env("SIGSTORE_OVERWRITE"),
+        help="Overwrite preexisting signature and certificate outputs, if present",
+    )
```
woodruffw, line 335: Given that this is a brand new command, I'm kind of tempted to remove/avoid the need to do anything except emit bundles 🙂 -- I _think_ the only options we need here are `--bundle` and `--overwrite`.

sigstore/_cli.py
facutuesca: Yes, I got these from `cosign`:
```
    --type='custom':
        specify a predicate type
        (slsaprovenance|slsaprovenance02|slsaprovenance1|link|spdx|spdxjson|cyclonedx|vuln|openvex|custom)
        or an URI
```

sigstore/_cli.py
facutuesca: The advantage of having these is that it allows us to reuse most of the code that was already written for `sigstore sign`. But I can remove them and refactor things around a bit.

sigstore/_internal/predicate.py
woodruffw, line 1: Thinking out loud: conceptually this belongs under the `dsse` module. We could do it now or later, but I think we eventually want the following structure:

```
dsse/
    __init__.py # current stuff
    _predicate.py # the current predicate.py file
```

sigstore/_cli.py
facutuesca: Ah my bad, I forgot I had changed them. I changed them to `slsaprovenance0_2` and `slsaprovenance1_0` because `slsaprovenance1` seems ambiguous. But yeah, these are not standard in any way

woodruffw: Great! Makes perfect sense to use them, then.

CC @loosebazooka @haydentherapper for visibility -- this is maybe too much of an implementation detail to standardize, but something for other clients to be aware of 🙂

sigstore/_internal/predicate.py
facutuesca: Should we do `sigstore/_dsse/` ? Or `sigstore/_internal/dsse` ?

sigstore/dsse/_predicate.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+class ConfigSource(BaseModel):
+    """
+    The ConfigSource object used by Invocation in v0.2
+    """
+
+    uri: Optional[StrictStr]
+    digest: Optional[DigestSetSource]
+    entry_point: Optional[StrictStr] = Field(None, alias="entryPoint")
```
woodruffw, line 75: Nitpick: Rather than having to explicitly declare each `alias` here and below, maybe we can define a common base model with a `model_config` that tweaks alias generation?

Something like:

```python
from pydantic.config import ConfigDict
from pydantic.alias_generators import to_camel

class _PredicateBase(BaseModel):
    model_config = ConfigDict(alias_generator=to_camel)

class ConfigSource(_PredicateBase):
    ...
```

...I think something like that will work, as will save us some manual typing 🙂

sigstore/_cli.py
haydentherapper: Seems reasonable to me to specify the predicate type as its full URL and not deal with shorthand.

sigstore/dsse/_predicate.py
facutuesca: Good idea! I also fixed a couple of bugs I found:
- `slsa-framework/slsa-github-generator` incorrectly generates predicates using `buildInvocationID` (note the final uppercase D, should be `buildInvocationId` instead). In order to fix that, I added a validation alias that allows for both options. The serialization alias will use the camelCase one set by the config you suggested.
- I was dumping the model without specifying it should output the camelCase aliases rather than our snake_case field names. Fixed by passing `by_alias=True` to `model_dump()`
- Our model dump also included the optional fields that were set to `None`. However, that's how we model them, those fields are just absent in the source predicate. So now we pass `exclude_none=True` to `model_dump`, to exclude those fields from the dump.

facutuesca: Mmm given the last point, I wonder if instead of dumping our Predicate model, we should instead just validate it and then use the exact input the user gave us to build the in-toto statement. 
Otherwise we might end up modifying it when doing the JSON -> Model -> JSON roundtrip.

sigstore/_cli.py
woodruffw: I think this still needs to be done -- IMO it's okay if end up being able to create a reusable `_sign_common`, since these extra flags are about the same size as the helper anyays (and we don't want to support them).

facutuesca: Fixed! See last commit

sigstore/dsse/_predicate.py
facutuesca: @woodruffw WDYT about this?

woodruffw: @facutuesca I think that's a good idea! In principle the each JSON layout shouldn't need to be stable, but in practice users might expect that (and as you've pointed out it avoids a potential source of ser/de errors on our side).

README.md
```diff
@@ -179,6 +180,58 @@ Output options:
 ```
 <!-- @end-sigstore-sign-help@ -->
 
+
+### Signing with DSSE envelopes
```
woodruffw, line 184: Thinking out loud: we should probably add some documentation about when users might want to do this versus "normal" signing. But we can do that in a follow-up 🙂

sigstore/dsse/_predicate.py
```diff
@@ -0,0 +1,219 @@
...
+    model_validator,
+)
+from pydantic.alias_generators import to_camel
+
+from sigstore.dsse import Digest
+
+PREDICATE_TYPE_SLSA_v0_2 = "https://slsa.dev/provenance/v0.2"
+PREDICATE_TYPE_SLSA_v1_0 = "https://slsa.dev/provenance/v1"
+
+SUPPORTED_PREDICATE_TYPES = [PREDICATE_TYPE_SLSA_v0_2, PREDICATE_TYPE_SLSA_v1_0]
```
woodruffw, line 36: Flagging for a follow-up: maybe we can make this an `enum.Enum` instead? That will reduce the number of unique imports we need to do, and allow us to define "supported" as "present in the enum".

woodruffw (approved):
Thanks @facutuesca, LGTM! I've flagged two things for follow-ups 🙂

woodruffw (approved):

PR #1116, opened 2024-09-10
Print in-toto statement when verifying DSSE

facutuesca:
#### Summary
Part of https://github.com/sigstore/sigstore-python/issues/1111. This changes the `sigstore verify` CLI command so that when verifying a bundle containing a DSSE envelope, if verification succeeds the inner in-toto statement is printed to the user.

```sh
$ sigstore verify identity README.md --cert-identity me@example.com --cert-oidc-issuer https://issuer.example.com
OK: README.md
{"_type":"https://in-toto.io/Statement/v1","subject":[{"name":"README.md","digest":{"sha256":"033be49064ed2a5f50bf81950f38741a8c550bc8076447452152c7b9d28728bc"}}],"predicateType":"slsaprovenance0_2","predicate":{"builder":{"id":"https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@refs/tags/v2.0.0"},"build_type":"https://github.com/slsa-framework/slsa-github-generator/generic@v1","invocation":{"config_source":{"uri":"git+https://gi.....
....
```

This is done because `sigstore-python` only verifies the subjects of the DSSE envelope match the artifacts being verified, it does not do any verification on the predicate of the statement. This should be done by the user, which is why we print the statement after verification succeeds, so that the user has easy access to it.

#### Release Note

* CLI: The `sigstore verify` command now outputs the inner in-toto statement
  when verifying DSSE envelopes. If verification is successful, the output
  will be "OK: $FILENAME" followed by the inner in-toto statement. This allows
  the user to see the statement's predicate, which `sigstore-python` does not
  verify and should be verified by the user.

cc @woodruffw

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+### Changed
+
+* CLI: The `sigstore verify` command now outputs the inner in-toto statement
+  when verifying DSSE envelopes. If verification is successful, the output
+  will be "OK: $FILENAME" followed by the inner in-toto statement. This allows
+  the user to see the statement's predicate, which `sigstore-python` does not
+  verify and should be verified by the user.
+
 ## [3.2.0]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
-### Changed
+### Added
 
 * CLI: The `sigstore verify` command now outputs the inner in-toto statement
   when verifying DSSE envelopes. If verification is successful, the output
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,9 @@ All versions prior to 0.9.0 are untracked.
 
 * CLI: The `sigstore verify` command now outputs the inner in-toto statement
   when verifying DSSE envelopes. If verification is successful, the output
-  will be "OK: $FILENAME" followed by the inner in-toto statement. This allows
-  the user to see the statement's predicate, which `sigstore-python` does not
-  verify and should be verified by the user.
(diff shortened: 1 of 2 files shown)


sigstore/_cli.py
```diff
@@ -811,8 +811,10 @@ def _verify_identity(args: argparse.Namespace) -> None:
         )
 
         try:
-            _verify_common(verifier, hashed, bundle, policy_)
+            statement = _verify_common(verifier, hashed, bundle, policy_)
             print(f"OK: {file}")
```
woodruffw, line 815: Let's move this to `stderr`:

```suggestion
            print(f"OK: {file}", file=sys.stderr)
```

CHANGELOG.md
```diff
@@ -8,6 +8,14 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+### Changed
```
woodruffw, line 11: This should be considered a semver addition 🙂 

```suggestion
### Added
```

CHANGELOG.md
```diff
@@ -8,6 +8,14 @@ All versions prior to 0.9.0 are untracked.
...
+
+* CLI: The `sigstore verify` command now outputs the inner in-toto statement
+  when verifying DSSE envelopes. If verification is successful, the output
+  will be "OK: $FILENAME" followed by the inner in-toto statement. This allows
```
woodruffw, line 15: Nit: I think we can remove this part, since `OK:` will be on stderr now.

woodruffw:
Thanks @facutuesca, looks good! Just a few small comments.

sigstore/_cli.py
facutuesca: fixed!

CHANGELOG.md
facutuesca: fixed

woodruffw (approved):
Thanks @facutuesca, nice work!

PR #1121, opened 2024-09-13
Attestation CLI command improvements

facutuesca:
Follow up to https://github.com/sigstore/sigstore-python/pull/1115#discussion_r1752718866 and https://github.com/sigstore/sigstore-python/pull/1115#discussion_r1759058534.

This PR:
- Changes `sigstore attest` to sign over the predicate file's contents, rather than our Pydantic model JSON dump. See the comments in the code for the rationale.
- Changes the implementation to use an `enum.Enum` for the predicate types
- Forbids unexpected fields in predicate JSON during model validation

cc @woodruffw

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -16,12 +16,13 @@ from __future__ import annotations
 
 import argparse
 import base64
+import json
 import logging
 import os
 import sys
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict, NoReturn, Optional, TextIO, Union
+from typing import Any, Dict, NoReturn, Optional, TextIO, Union
 
 from cryptography.hazmat.primitives.serialization import Encoding
 from cryptography.x509 import load_pem_x509_certificate
@@ -42,7 +43,6 @@ from sigstore._utils import sha256_digest
 from sigstore.dsse import StatementBuilder, Subject
 from sigstore.dsse._predicate import (
     SUPPORTED_PREDICATE_TYPES,
-    Predicate,
     PREDICATE_TYPE_SLSA_v0_2,
     PREDICATE_TYPE_SLSA_v1_0,
     SLSAPredicateV0_2,
@@ -591,7 +591,7 @@ def main(args: list[str] | None = None) -> None:
 
 
 def _sign_common(
-    args: argparse.Namespace, output_map: OutputMap, predicate: Predicate | None
+    args: argparse.Namespace, output_map: OutputMap, predicate: dict[str, Any] | None
 ) -> None:
     """
     Signing logic for both `sigstore sign` and `sigstore attest`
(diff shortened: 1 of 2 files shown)


sigstore/dsse/_predicate.py
```diff
@@ -30,10 +31,15 @@
 
 from sigstore.dsse import Digest
 
-PREDICATE_TYPE_SLSA_v0_2 = "https://slsa.dev/provenance/v0.2"
-PREDICATE_TYPE_SLSA_v1_0 = "https://slsa.dev/provenance/v1"
 
-SUPPORTED_PREDICATE_TYPES = [PREDICATE_TYPE_SLSA_v0_2, PREDICATE_TYPE_SLSA_v1_0]
+class PredicateType(enum.Enum):
```
woodruffw, line 35: This makes `str(PredicateType)` behave nicely, I believe 🙂 

```suggestion
class PredicateType(str, enum.Enum):
```

sigstore/_cli.py
```diff
@@ -704,16 +696,23 @@ def _attest(args: argparse.Namespace) -> None:
...
+            # type.
+            if args.predicate_type == PredicateType.SLSA_v0_2.value:
+                SLSAPredicateV0_2.model_validate(predicate)
+            elif args.predicate_type == PredicateType.SLSA_v1_0.value:
+                SLSAPredicateV1_0.model_validate(predicate)
             else:
                 _invalid_arguments(
                     args,
-                    f'Unsupported predicate type "{args.predicate_type}". Predicate type must be one of: {SUPPORTED_PREDICATE_TYPES}',
+                    f'Unsupported predicate type "{args.predicate_type}". Predicate type must be one of: {[p.value for p in PredicateType]}',
```
woodruffw, line 712: With the suggestion below, I think this can just be:

```suggestion
                    f'Unsupported predicate type "{args.predicate_type}". Predicate type must be one of: {list(PredicateType)}',
```

(but you should sanity check that 😅)

sigstore/_cli.py
```diff
@@ -277,10 +275,10 @@ def _parser() -> argparse.ArgumentParser:
     dsse_options.add_argument(
         "--predicate-type",
         metavar="TYPE",
-        choices=SUPPORTED_PREDICATE_TYPES,
+        choices=[p.value for p in PredicateType],
         type=str,
         required=True,
-        help=f"Specify a predicate type ({', '.join(SUPPORTED_PREDICATE_TYPES)})",
+        help=f"Specify a predicate type ({', '.join((p.value for p in PredicateType))})",
```
woodruffw, line 281: Same comment below, I think these can be just `PredicateType` or `list(PredicateType)` as appropriate.

sigstore/dsse/_predicate.py
facutuesca: doesn't that interfere with iterating over the enum values? Since [p for p in PredicateType] will behave differently if it's a subclass from `str` first rather than `PredicateType`.
I force-pushed an amend of the last commit, can you check if the way I wrote it is idiomatic?

sigstore/_cli.py
facutuesca: Fixed

sigstore/dsse/_predicate.py
facutuesca: Fixed!

woodruffw (approved):
LGTM, thanks @facutuesca!

PR #1124, opened 2024-09-13
Add CLI integration tests for attest subcommand

facutuesca:
This PR starts adding integration tests for the CLI subcommands. Part of https://github.com/sigstore/sigstore-python/issues/1111

cc @woodruffw


test/integration/cli/conftest.py
```diff
@@ -12,17 +12,89 @@
...
+)
 
 from sigstore._cli import main
+from sigstore.oidc import _DEFAULT_AUDIENCE
 
 _ASSETS = (Path(__file__).parent.parent.parent / "assets/integration").resolve()
 assert _ASSETS.is_dir()
 
 
+def _has_oidc_id():
```
woodruffw, line 33: This and below are duplicated from the unit-level conftest, right? I think this is fine for now, but we should probably think about deduping this at some point. Maybe some kind of `test/{common,unit,integration}` layout where `test.common` contains common helpers?

woodruffw (approved):
LGTM, thank you @facutuesca!

PR #1125, opened 2024-09-16
Add support for verifying digests to CLI verify commands

facutuesca:
#### Summary
This PR changes the `sigstore verify` CLI command so that it accepts digests as inputs (in addition to files).
Concretely:

```shell
# Before
sigstore verify identity ..... path/to/artifact.zip  # can only verify an artifact

# After
sigstore verify identity ..... path/to/artifact.zip  # can verify an artifact
sigstore verify identity ... --bundle b.sigstore.json sha256:aaabbbb.....  # can also verify an artifact's digest
```

This is useful in scenarios where the user doesn't want/need to download the artifact, but has access to its digest.

Related to https://github.com/sigstore/sigstore-conformance/issues/157

#### Release Note

* CLI: The `sigstore verify` command now supports verifying digests. This means
  that the user can now pass a digest like `sha256:aaaa....` instead of the
  path to an artifact, and `sigstore-python` will verify it as if it was the
  artifact with that digest.

#### Documentation

Example added to the README

cc @woodruffw

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,7 +247,7 @@ usage: sigstore verify identity [-h] [-v] [--certificate FILE]
                                 [--signature FILE] [--bundle FILE] [--offline]
                                 --cert-identity IDENTITY --cert-oidc-issuer
                                 URL
-                                FILE [FILE ...]
+                                FILE_OR_DIGEST [FILE_OR_DIGEST ...]
 
 optional arguments:
   -h, --help            show this help message and exit
@@ -262,7 +262,8 @@ Verification inputs:
                         multiple inputs (default: None)
   --bundle FILE         The Sigstore bundle to verify with; not used with
                         multiple inputs (default: None)
-  FILE                  The file to verify
+  FILE_OR_DIGEST        The file path or the digest to verify. The digest
+                        should start with the 'sha256:' prefix.
 
 Verification options:
   --offline             Perform offline verification; requires a Sigstore
@@ -290,7 +291,7 @@ usage: sigstore verify github [-h] [-v] [--certificate FILE]
                               [--cert-identity IDENTITY] [--trigger EVENT]
                               [--sha SHA] [--name NAME] [--repository REPO]
                               [--ref REF]
-                              FILE [FILE ...]
+                              FILE_OR_DIGEST [FILE_OR_DIGEST ...]
 
 optional arguments:
   -h, --help            show this help message and exit
@@ -305,7 +306,8 @@ Verification inputs:
                         multiple inputs (default: None)
   --bundle FILE         The Sigstore bundle to verify with; not used with
                         multiple inputs (default: None)
-  FILE                  The file to verify
+  FILE_OR_DIGEST        The file path or the digest to verify. The digest
+                        should start with the 'sha256:' prefix.
 
 Verification options:
   --offline             Perform offline verification; requires a Sigstore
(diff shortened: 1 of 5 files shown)


sigstore/hashes.py
```diff
@@ -55,3 +55,9 @@ def _as_prehashed(self) -> Prehashed:
         if self.algorithm == HashAlgorithm.SHA2_256:
             return Prehashed(hashes.SHA256())
         raise Error(f"unknown hash algorithm: {self.algorithm}")
+
+    def __str__(self) -> str:
+        """
+        Returns a str representation of this `Hashed`.
+        """
+        return f"{self.algorithm.name}:{self.digest.hex()}"
```
woodruffw, line 63: Small unit test for this, please 🙂

woodruffw (dismissed):
LGTM, nice work @facutuesca! One small request for a unit test.

README.md
```diff
@@ -421,6 +423,19 @@ $ python -m sigstore verify identity foo.txt bar.txt \
...
+
+`sigstore-python` supports verifying digests directly, without requiring the artifact to be
+present. The digest should be prefixed with the `sha256:` string:
+
+```console
+# looks for foo.txt.sigstore
+$ python -m sigstore verify identity sha256:ce8ab2822671752e201ea1e19e8c85e73d497e1c315bfd9c25f380b7625d1691 \
+    --cert-identity 'hamilcar@example.com' \
+    --cert-oidc-issuer 'https://github.com/login/oauth'
+    --bundle 'foo.txt.sigstore.json'
```
di, line 436: Either the comment or the `--bundle` value is wrong here I think?

sigstore/_cli.py
```diff
@@ -149,12 +165,25 @@ def _add_shared_verify_input_options(group: argparse._ArgumentGroup) -> None:
         default=os.getenv("SIGSTORE_BUNDLE"),
         help=("The Sigstore bundle to verify with; not used with multiple inputs"),
     )
+
+    def file_or_digest(arg: str) -> Hashed | Path:
+        if arg.startswith("sha256:"):
+            digest = bytes.fromhex(arg[len("sha256:") :])
+            if len(digest) != 32:
+                raise ValueError()
```
di, line 173: Where will this be caught? We might want to provide a more informative error message here.

README.md
facutuesca: yup, comment was wrong, fixed!

sigstore/_cli.py
facutuesca: `argparse` [catches](https://docs.python.org/3/library/argparse.html#type) it:
> The argument to type can be any callable that accepts a single string. If the function raises [ArgumentTypeError](https://docs.python.org/3/library/argparse.html#argparse.ArgumentTypeError), [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), or [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError), the exception is caught and a nicely formatted error message is displayed. No other exception types are handled.

The only problem is that it ignores the error message, so whatever message I put in `ValueError("mymsg")`, the shown error is always:
```
sigstore verify identity: error: argument FILE_OR_DIGEST: invalid file_or_digest value: 'sha256:3072b971b74902ac37ee452649efbaaeca6cd80c93cef5e27c1f491d7aa2ab443'
```

di: Hmm, would be nice to have a more specific message but this should be fine!

sigstore/hashes.py
facutuesca: Done, see last commit

woodruffw (approved):

PR #1140, opened 2024-09-24
workflows: various CQA fixes

woodruffw:
A bunch of small quality fixes to our workflows.

* Adds explicit `persist-credential:` settings everywhere, disabling the (default) credential persistence as much as possible.
* Moves all permissions into steps and lowers the default outer permissions where explicit.
* Replaces a shell interpolation of `github.ref` with `GITHUB_REF`.


jku (approved):
Nice work, thanks

PR #1152, opened 2024-10-02
_cli: files always take precedence over digests

woodruffw:
This fixes a small edge case where a user supplies `sha256:hash.jsonl` or similar (such as produced by default by `gh attestation`) and the `sigstore verify` subcommands interpret it as an (invalid) hash rather than a file input.

The new behavior is to always interpret the input as a path if a file at that path is extant, and to otherwise interpret it as a hash.

CC @facutuesca


facutuesca (approved):
...that was an edge case that I wasn't expecting at all.

LGTM! I'll make a note to add a test for it when we write the CLI `verify` tests

jku (approved):

Issue #1153, opened 2024-10-03
enable codecov (or similar)?


gliptak:
**Description**

would you be interested in a PR uploading coverage results to http://codecov.io (or similar hosted location)?

https://github.com/codecov/example-python/blob/main/.github/workflows/ci.yml

might require generating and setting up a secret

https://about.codecov.io/tool/github/

https://codecov.io/gh/

woodruffw:
> would you be interested in a PR uploading coverage results to http://codecov.io (or similar hosted location)?

I've had a lot of problems with CodeCov's reliability/uptime on other projects, so I'd prefer not to use a third-party service for this.

Long term, the goal is to get to 100% test coverage here -- when that happens, we can enable `--fail-under 100` and gate PRs on that, with no third-party coverage service required.

Medium term, I think we could detect coverage regressions by persisting the current coverage level somewhere (GH actions artifacts?) and comparing against it. I remember seeing some prior art for that somewhere, but I'd need to find it again.

TL;DR: Thanks for the offer, but CodeCov in particular is not desired at the moment. We'd be interested in coverage regression tracking more generally, but that requires a bit of design thought.

PR #1155, opened 2024-10-04
Add coverage percentage to GHA

gliptak:
#### Summary

step towards https://github.com/sigstore/sigstore-python/issues/1153


woodruffw:
I appreciate that you're trying to help with the coverage issue, but could you hold off on this for a bit? It's going to require some design thought and discussions between the current maintainers, like I said in https://github.com/sigstore/sigstore-python/issues/1153#issuecomment-2392121575

We already have coverage reporting in GHA, which gets rendered in a more human-friendly way than just printing it to the action's log. For example: https://github.com/sigstore/sigstore-python/actions/runs/11185891030?pr=1158#summary-31099929269

PR #1174, opened 2024-10-15
_cli: add `plumbing update-trust-root`

### LINKED ISSUE #1172: Add a command to update `trusted_root.json` without verifying anything
**Description**

I'm sorry if I'm missing something but FWICS the only way to update `trusted_root.json` right now is to actually perform a verification. Could you please consider adding a "pure" update command, that would only update `trusted_root.json` and indicate whether the update was successful via exit status?

This would be part of Gentoo sigstore deployment, where we'd like to separately update the trust root in an online environment, and then deploy it to offline environments.

woodruffw:
This adds a new plumbing-level command: `update-trust-root`. When run, `sigstore plumbing update-trust-root` will perform a TUF update for the appropriate production or staging trust root, depending on whether the top-level `--staging` flag is used.

Closes #1172.

CC @mgorny if you can, it'd be great to have you trial this locally to confirm it meets your needs! In particular it'd be useful to know if you'd prefer it to be "louder" in terms of outputs, i.e. print something on `stdout`/`stderr`.

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -40,6 +40,7 @@ from sigstore._internal.fulcio.client import ExpiredCertificate
 from sigstore._internal.rekor import _hashedrekord_from_parts
 from sigstore._internal.rekor.client import RekorClient
 from sigstore._internal.trust import ClientTrustConfig
+from sigstore._internal.tuf import DEFAULT_TUF_URL, STAGING_TUF_URL, TrustUpdater
 from sigstore._utils import sha256_digest
 from sigstore.dsse import StatementBuilder, Subject
 from sigstore.dsse._predicate import (
@@ -573,6 +574,14 @@ def _parser() -> argparse.ArgumentParser:
         help="Overwrite the input bundle with its fix instead of emitting to stdout",
     )
 
+    # `sigstore plumbing update-trust-root`
+    plumbing_subcommands.add_parser(
+        "update-trust-root",
+        help="update the local trust root to the latest version via TUF",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        parents=[parent_parser],
+    )
+
     return parser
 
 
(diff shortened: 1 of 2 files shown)


mgorny:
Yeah, seems to work fine. My personal weak preference is for more verbose output, particularly if it could help us debug issues in the future. Thanks!

woodruffw:
> Yeah, seems to work fine. My personal weak preference is for more verbose output, particularly if it could help us debug issues in the future. Thanks!

No problem! And makes sense -- I'll mess around with potential outputs here.

sigstore/_cli.py
```diff
@@ -1222,3 +1232,16 @@ def _fix_bundle(args: argparse.Namespace) -> None:
...
+    # Simply creating the TrustedRoot in online mode is enough to perform
+    # a metadata update.
+    if args.staging:
+        trusted_root = TrustedRoot.staging(offline=False)
+    else:
+        trusted_root = TrustedRoot.production(offline=False)
+
+    _console.print(
+        f"Trust root updated: {len(trusted_root.get_fulcio_certs())} Fulcio certificates"
+    )
```
woodruffw, line 1247: It's not the best/most useful output, but it should be a visually obvious sign that the step hasn't quietly failed.

Another option here is to dump the trust root as JSON. Not sure how I feel about that, since I don't want people to assume a stable interface on this subcommand quite yet 😅

mgorny: I think that would be excessive verbose and/or might confuse people as to whether it was written to the file or whether the user needs to redirect the output. Perhaps it could print cache path too, like `written to ...`.

jku: I'm maybe not following which file you are talking about but I think exposing  `trusted_root.json` location is the only thing that makes sense (as users can't do much with TUF files): that is easily available in all TrustedRoot constructors so you could store the path in TrustedRoot for potential console output later.

jku (approved):
Seems ok for a first iteration of a "plumbing" command. The  fulcio cert count seems a bit useless so that could be the trusted root location instead

I suppose one option would be to add a file hash to the output: it's not friendly to humans but would be a little more auditable

sigstore/_cli.py
woodruffw: Yeah, I was thinking `trusted_root.json`.

PR #1179, opened 2024-10-22
_cli: don't warn on bare .sigstore if cert/sig is used

woodruffw:
This should fix #1178.

CC @sethmlarson

PR #1192, opened 2024-10-25
_cli: fix warning check

woodruffw:
This fixes the check added in https://github.com/sigstore/sigstore-python/pull/1179 by ensuring that we don't bypass the legacy bundle discovery behavior by accident when trying to suppress the warning.

I'll do a bugfix release after this.


jku (approved):
yes that seems more correcter

Issue #1194, opened 2024-10-25
Post-mortem for regression in 3.5.0 


woodruffw:
Release 3.5.0 had a minor regression which we only noticed after release: handling of "legacy" sigstore bundles (i.e. `.sigstore` instead of `.sigstore.json`) was broken by an overly broad check on `.crt`/`.sig` inputs.

No other verification flows were affected, including any flows where a user passes the verification materials explicitly rather than discovering them via file suffixes. 

## Resolution

We released 3.5.1 with a fix.

## Improvement items

We should have an integration test that ensures we don't regress on this CLI behavior again.

woodruffw:
CC @facutuesca could you try and extend our current integration tests to cover this? Should be as simple as ensuring that `sigstore verify ... foo.txt` continues to work when `foo.txt.sigstore` (and only `.sigstore`, not `.sigstore.json`) is present.

PR #1198, opened 2024-10-28
Fix warning for CLI verification of legacy bundles

facutuesca:
This PR fixes a check in the CLI command `sigstore verify` where we should have printed a warning while verifying a legacy bundle (`*.sigstore`) if no bare materials (`*.crt` and `*.sig`) are provided.

The warning was never printed, due to an error in the check that made it always return `False`.

This PR also adds a regression test for the bug fixed in https://github.com/sigstore/sigstore-python/pull/1192, checking that the CLI correctly verifies a legacy bundle and that it logs the expected warning.

This closes https://github.com/sigstore/sigstore-python/issues/1194.

cc @woodruffw

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* Fixed a CLI parsing bug introduced in 3.5.1 where a warning about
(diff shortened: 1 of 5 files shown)


woodruffw (dismissed):
Nice work @facutuesca! Mind adding a CHANGELOG entry, since this fixes my broken fix?

woodruffw (approved):

Issue #1182, opened 2024-10-23
[Meta] Support for Timestamp Protocol verification 


DarkaMaul:
Improve the support of `sigstore-python` for signing and verification APIs and CLIs for accepting signed time from a TSA versus an artifact transparency log (or both).

**Description**

We developed a new [`rfc3161-client`](https://github.com/trailofbits/rfc3161-client) to perform the parsing and generation of the Timestamp Request/Response objects.

This set of changes aims to integrate this new client into `sigstore-python`. We'll split the contributions into several PRs to make their review slightly easier.

- [x] Write the `TimestampVerificationData` models
- [x] Update `TrustedRoot` to provide the Timestamps Authority information. 
⚠️ This field is planned for removal ( https://github.com/sigstore/root-signing/issues/1268 , https://github.com/sigstore/root-signing-staging/pull/156 )
- [x] Add the Timestamp verification mechanism itself.
- [ ] Add the possibility to sign the bundle with a TSA
- [ ] ~Add a CLI parameter `require-timestamp` to trigger the verification~

/cc @woodruffw @facutuesca

jku:
Can I get a short explanation for the use cases:

> Support for Timestamp Protocol verification
vs
> This field is planned for removal ([ [targets v11] What to do with the GitHub TSA in trusted_root.json root-signing#1268 , Signing event: sign/update-targets-1 root-signing-staging#156](https://github.com/sigstore/root-signing/issues/1268) )


If the public good instance is not going to support this, is it worth the additional code and dependency in this client?

DarkaMaul:
## Rationale

The idea is to use a Timestamp Authority to attest or verify _when_ the signature has been generated.

The current approach offers a way of verifying the signature time using the timestamp on the inclusion proof. However, this is unreliable because the time here is not verifiable. To quote the spec : 

> Timestamping. Currently, the Transparency Service includes a timestamp in its response to the Signer. This timestamp comes from the Transparency Service's internal clock, which is not externally verifiable or immutable.

Using a dedicated Time Source (such as a TSA) separates concerns between the timestamp and logging. A TSA provides a cryptographically verifiable, trusted timestamp designed for this purpose (following RFC 3161). This is particularly important for the "hybrid model" of certificate verification described in the spec, where signatures must have a valid timestamp while every certificate in the chain up to the root is valid.

And as an excellent point, it allows us to be more closely aligned with the spec:
> For this reason, a Signer SHOULD get their signatures timestamped.

[...]

PR #1200, opened 2024-10-31
Add CertificateAuthority

DarkaMaul:
#### Summary

Following up on #1182, this PR adds a `CertificateAuthority` class to the trust and uses it to retrieve both Fulcio certificates and Timestamping Authorities.


sigstore/_internal/trust.py
```diff
@@ -217,6 +220,102 @@ def __str__(self) -> str:
...
+    """
+
+    def __init__(self, inner: _CertificateAuthority):
+        """
+        Construct a new `CertificateAuthority`.
+
+        @api private
+        """
+        self._inner = inner
+        self._leaf: Certificate | None = None
```
woodruffw, line 235: Hmm, where is the concept of a "leaf" in a `CertificateAuthority` coming from? The message itself only defines `cert_chain` as being a list of root and intermediate CAs, so there's no "leaf" certificate per se (except in the sense that one of the intermediates is the last one).

I think we should drop this and its associated property:

```suggestion
```

sigstore/_internal/trust.py
```diff
@@ -217,6 +220,102 @@ def __str__(self) -> str:
...
+        Root certificate of the authority.
+        """
+        return self._root
+
+    @property
+    def intermediates(self) -> list[Certificate]:
+        """
+        List of intermediates certificates.
+        """
+        return self._intermediates
```
woodruffw, line 293: Related: I think we can remove both of these, and allow access only through the `certificates(...)` method. FWICT they're only used in tests anyways.

woodruffw (commented):
Thanks @DarkaMaul! This is a good start, but I think it can be simplified quite a bit -- we don't need to be so stringent about classifying the different members of the `CertificateAuthority` message, since they all end up in a trust "pool" anyways for chain building purposes 🙂

DarkaMaul:
> Thanks @DarkaMaul! This is a good start, but I think it can be simplified quite a bit -- we don't need to be so stringent about classifying the different members of the `CertificateAuthority` message, since they all end up in a trust "pool" anyways for chain building purposes 🙂

I've updated the code to remove the distinction between all the different certificates 👍

woodruffw (approved):
Nice work!

PR #1206, opened 2024-11-07
Timestamp Authority Verification

DarkaMaul:
#### Summary

This PR verifies signed timestamp within a bundle using Timestamp Authorities provided within the Trusted Root.

#### Release Note

**Added**:

- Signed timestamps embedded in a bundle are now automatically verified.

#### Documentation

The documentation is already written in [Verifying Timestamp](https://docs.sigstore.dev/cosign/verifying/timestamps/)

/cc @woodruffw @facutuesca


jku (commented):
> The documentation is already written in [Verifying Timestamp](https://docs.sigstore.dev/cosign/verifying/timestamps/)

At least for the PR/issue I don't think it's quite true. It's not obvious to me how this feature is going to be used.
* Is there spec for what clients SHOULD/MUST do with timestamps in bundles? If not how is the policy decided?
* how do we end up with bundles with timestamps? Does the signer need to actively decide to include them?
* if there are timestamps in the bundle, how does the verifier get the cert chain? (theoretically from the trusted root sure,  but is there a practical path there? is this just for private instances?)
* who sets the verification threshold policy? If there can be a variable number of timestamps in the bundle and the trusted root can contain a variable number of TSAs, how can you choose a reasonable value for threshold?

I realize these are not really questions that relate to the PR but ... I did try to start a discussion in the issue earlier already. I left some actual code comments but those are more to make sure I understand the intent correctly.

I think this feature would really benefit from a design doc (by which I mean just a detailed comment in the issue describing how it's all going to be used in the future): I don't think this client should just implement things because cosign does. Cosign is famous for kitchen sinks (with 6 water temp knobs).

sigstore/verify/verifier.py
```diff
@@ -154,6 +231,30 @@ def _verify_common_signing_cert(
...
+                msg = (
+                    "No Timestamp Authorities have been provided to validate this "
+                    "bundle but it contains a signed timestamp"
+                )
+                raise VerificationError(msg)
+
+            verified_timestamp = self._verify_timestamp_authority(bundle)
+            # The threshold is set to (1) by default but kept as a variable to allow
+            # this value to change
+            if len(verified_timestamp) < self.verify_timestamp_threshold:
```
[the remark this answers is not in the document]
DarkaMaul: Great catch here: that was indeed a mistake.

I'm also adding a test to catch this type of issue (only a single timestamp is verified, but two are required).

sigstore/verify/verifier.py
```diff
@@ -76,6 +82,8 @@ def __init__(self, *, rekor: RekorClient, trusted_root: TrustedRoot):
         ]
         self._trusted_root = trusted_root
 
+        self.verify_timestamp_threshold: int = 1
```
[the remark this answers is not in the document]
DarkaMaul: In `cosign`, this is provided as a command line argument (one of many).
In [`sigstore-go`](https://github.com/sigstore/sigstore-go/blob/e92142f0734064ebf6001f188b7330a1212245fe/cmd/sigstore-go/main.go#L114), this is set to 1 directly.

I would lean on the `constant` side of things here, as the end-user currently has no way of modifying this value.

DarkaMaul:
> * Is there spec for what clients SHOULD/MUST do with timestamps in bundles? If not how is the policy decided?

According to the Sigstore Client [Spec](https://docs.google.com/document/d/1kbhK2qyPPk8SLavHzYSDM8-Ueul9_oxIMVFuWMWKz0E/edit?tab=t.0#heading=h.g11ovq2s1jxh), it should be decided by the policy. 

> If the verification policy uses the Timestamping Service, the Verifier MUST verify the timestamping response using the Timestamping Service root key material, as described in [Spec: Timestamping Service](https://docs.google.com/document/u/0/d/1FoRHXejIhXwEai0RS3iRsN1HfCV16fJOp582Vl8KA7A/edit), with the raw bytes of the signature as the timestamped data. The Verifier MUST then extract a timestamp from the timestamping response. If verification or timestamp parsing fails, the Verifier MUST abort.

The current implementation follows a minimal approach - verifying timestamps only when they are provided. While this allows for a non-disruptive initial integration that maintains compatibility with existing workflows, I agree we should make our policy more explicit here.

(Note: the document linked here is restricted and I don't have access to it) 

> * how do we end up with bundles with timestamps? Does the signer need to actively decide to include them?

Currently, timestamps are included through explicit opt-in by the signer. For example, using cosign, you would specify a timestamp server URL:

```shell
cosign sign-blob --yes --bundle bundle.sigstore --new-bundle-format=true \
--timestamp-server-url="http://localhost:3000/api/v1/timestamp" \  # here
--oidc-issuer "https://oauth2.sigstage.dev/auth" --fulcio-url "https://fulcio.sigstage.dev" --rekor-url "https://rekor.sigstage.dev" ./bundle.txt
```

While timestamps currently require explicit opt-in, we could consider making them opt-out in the future. As a next step, I plan to implement signing a bundle with timestamping support in `sig store-python` as a follow-up PR.

> * if there are timestamps in the bundle, how does the verifier get the cert chain? (theoretically from the trusted root sure,  but is there a practical path there? is this just for private instances?)

You raise a valid point about the certificate chain verification path. While theoretically, the chain can be sourced from the trusted root (and there are fields specified for it in the bundle format), we need to ensure they are used within the community.

> * who sets the verification threshold policy? If there can be a variable number of timestamps in the bundle and the trusted root can contain a variable number of TSAs, how can you choose a reasonable value for threshold?

The verification threshold is a parameter specified in the spec. In practice, `sigstore-go` currently uses a fixed value of 1. This simplifies the initial implementation, though we should document this decision and its implications more explicitly.

> I realize these are not really questions that relate to the PR but ... I did try to start a discussion in the issue earlier already. I left some actual code comments but those are more to make sure I understand the intent correctly.
I think this feature would really benefit from a design doc (by which I mean just a detailed comment in the issue describing how it's all going to be used in the future): I don't think this client should just implement things because cosign does. Cosign is famous for kitchen sinks (with 6 water temp knobs).

You're absolutely right, and I apologize for missing the earlier discussion in the issue. Your point about not simply mirroring cosign's features without clear design rationale is well-taken. Let's continue the discussion there.

sigstore/dsse/__init__.py
```diff
@@ -228,6 +228,17 @@ def __eq__(self, other: object) -> bool:
...
 
+    @property
+    def signature(self) -> bytes:
+        """Return the decoded bytes of the Envelope signature."""
+        if len(self._inner.signatures) == 0:
+            return b""
+
+        signature_bytes = self._inner.signatures[0].sig
+        if not signature_bytes:
+            return b""
```
woodruffw, line 239: Rather than forwarding these invalid states, we should enforce the invariant that a `dsse.Envelope` always has exactly one non-empty signature member.

This should be done in the `Envelope` constructor -- there's an example of this pattern here:

https://github.com/trail-of-forks/sigstore-python/blob/5f23327c1520b757670f6e3aeccda1da66c4805f/sigstore/models.py#L444

sigstore/verify/verifier.py
```diff
@@ -154,6 +234,30 @@ def _verify_common_signing_cert(
...
+                )
+                raise VerificationError(msg)
+
+            verified_timestamp = self._verify_timestamp_authority(bundle)
+            # The threshold is set to (1) by default but kept as a variable to allow
+            # this value to change
+            if verified_timestamp < VERIFY_TIMESTAMP_THRESHOLD:
+                msg = (
+                    f"Not enough Timestamp validated to meet the Validation "
+                    f"Threshold ({verified_timestamp}/{VERIFY_TIMESTAMP_THRESHOLD})"
```
woodruffw, line 257:
```suggestion
                    f"Not enough timestamps validated to meet the validation "
                    f"threshold ({verified_timestamp}/{VERIFY_TIMESTAMP_THRESHOLD})"
```

sigstore/verify/verifier.py
```diff
@@ -108,6 +118,76 @@ def _from_trust_config(cls, trust_config: ClientTrustConfig) -> Verifier:
...
+                    "Unable to verify Timestamp because not in CA time range."
+                )
+            else:
+                _logger.debug(
+                    "Unable to verify Timestamp because no validity provided."
+                )
+
+        return False
+
+    def _verify_timestamp_authority(self, bundle: Bundle) -> int:
```
woodruffw, line 165: I think this needs to return the verified timestamp, not the _number_ of verified timestamps.

(It needs to return the verified timestamp, because that becomes the source of trusted time for the rest of the verification flow.)

sigstore/dsse/__init__.py
DarkaMaul: To keep changes contained, this is part of an other PR : #1211

sigstore/verify/verifier.py
DarkaMaul: Great catch - I've reworked a bit this part to have a more informative return type and integrated the certificate chain validity check.

sigstore/_internal/timestamp.py
woodruffw, line 1: NB: I moved this to `sigstore._internal` since these don't need to be public APIs quite yet.

sigstore/verify/verifier.py
```diff
@@ -108,6 +119,156 @@ def _from_trust_config(cls, trust_config: ClientTrustConfig) -> Verifier:
...
+        Verify a Signed Timestamp using the TSA provided by the Trusted Root.
+        """
+        cert_authorities = self._trusted_root.get_timestamp_authorities()
+        for certificate_authority in cert_authorities:
+            certificates = certificate_authority.certificates(allow_expired=True)
+
+            builder = VerifierBuilder()
+            for certificate in certificates[:-1]:
+                builder.add_intermediate_certificate(certificate)
+            builder.add_root_certificate(certificates[-1])
```
woodruffw, line 135: This is a minor thing, but technically we can do `add_root_certificate` for all of these, since these are all in the "root of trust" regardless of whether they're intermediates or not. That'll result in a slightly shorter chain when verifying.

woodruffw (approved):
Thanks @DarkaMaul, nice work!

PR #1216, opened 2024-11-15
Sign Bundle with a Timestamp Authority

DarkaMaul:
Final bit of https://github.com/sigstore/sigstore-python/issues/1182

#### Summary

This PR introduces the possibility to create a bundle with a timestamp signed by a TimestampAuthority.

### Added

- If a `Timestamp Authority` URL has been provided in the SigningConfig, the bundle are now automatically generated with a signed timestamp.  

/cc @woodruffw @facutuesca

PR #1224, opened 2024-11-25
CHANGELOG: record #1216

woodruffw:
NFC.


jku (approved):
LGTM.

Do we have documentation for signed timestamp verification (detailing the policy decisions)? I'm a little worried that at some point in future some clients start including timestamps (maybe multiple timestamp) in bundles and other clients start verifying them but there is no obviously correct policy and the result is we end up with verify failing when it maybe should not.

PR #1225, opened 2024-11-26
Update Sigstore Timestamp using dependabot

DarkaMaul:
#### Summary

This PR is a follow-up from #1216 and update (Sigstore) Timestamp Authority to be picked up by dependabot ( following a comment from @woodruffw )

I've tested the approach at https://github.com/DarkaMaul/test-dep, and that appears to be working.


.github/workflows/ci.yml
```diff
@@ -61,7 +61,8 @@ jobs:
       - name: test (timestamp-authority)
         if: ${{ matrix.conf.os == 'ubuntu-latest' }}
         run: |
-          wget https://github.com/sigstore/timestamp-authority/releases/download/${SIGSTORE_TIMESTAMP}/timestamp-server-linux-amd64 -O /tmp/timestamp-server
+          export SIGSTORE_TIMESTAMP_VERSION=$(grep "github.com/sigstore/timestamp-authority" .github/go.mod | awk '{print $2}')
```
DarkaMaul, line 64: FLAG: I'm unsure of the implications here - is this a dangerous pattern in a workflow ?

jku: looks ok to me.

export seems unnecessary though

.github/go.mod
```diff
@@ -0,0 +1,11 @@
+module sigstore/sigstore-python
+
+go 1.18
```
jku, line 3: this is very old by golang standards, probably unsupported for years by now...

Likely dependabot doesn't actually use golang at all so maybe that does not matter?

DarkaMaul: I've bumped the version here just in case and used the one in the latest ubuntu-runner.

PR #1236, opened 2024-12-02
fulcio: remove detached SCT support

### LINKED ISSUE #850: Removed support for detached SCTs
Per https://github.com/sigstore/fulcio/issues/1499: Fulcio may be removing detached SCT support soon, so sigstore-python can follow along.

woodruffw:
This fully removes detached SCT support, meaning
that any Fulcio instance must use embedded SCTs.

CC @haydentherapper for sanity-checking here, since this follows your original RFC on Fulcio.

Closes #850.

xref: https://github.com/sigstore/fulcio/issues/1499


sigstore/_internal/fulcio/client.py
```diff
@@ -238,22 +132,12 @@ def post(
...
-        if resp.json().get("signedCertificateEmbeddedSct"):
-            sct_embedded = True
-            try:
-                certificates = resp.json()["signedCertificateEmbeddedSct"]["chain"][
-                    "certificates"
-                ]
-            except KeyError:
-                raise FulcioClientError("Fulcio response missing certificate chain")
-        else:
-            sct_embedded = False
```
haydentherapper, line 250: For instances that don’t use a CT log, the certificate chain is written to signedCertificateDetachedSct. To continue to support these local deployments, you’d need to still read the chain from either field, but only check the SCT if embedded. 

Cc @codysoyland who added that for Sigstore-go

jku: Is it not a problem if the server can just "downgrade" to detached sct like that? 

Or are we supposed to always require embedded SCT if the trust root contains at least one CT log, but otherwise be ok with detached SCT?

woodruffw: Right now sigstore-python doesn't allow non-CT BYO uses at all, so allowing `signedCertificateDetachedSct` while removing detached SCT support would mean a larger refactor 😅 

Given that, I _think_ we're OK to remove this outright, since `signedCertificateDetachedSct`-without-CT is a use case we already didn't support. But I'm curious if this matches @jku's understanding 🙂

haydentherapper: > Is it not a problem if the server can just "downgrade" to detached sct like that?
> 
> Or are we supposed to always require embedded SCT if the trust root contains at least one CT log, but otherwise be ok with detached SCT?

For Sigstore-go, it’s a threshold configured by the user. For cosign, SCTs are required unless opted out explicitly by the user.

If you require CT now, then agreed this would not be a breaking change.

jku: @woodruffw sure, that sounds good.

sigstore/_internal/fulcio/client.py
```diff
@@ -264,35 +148,12 @@ def post(
...
-                sct = DetachedFulcioSCT.model_validate_json(base64.b64decode(sct_b64))
-            except Exception as exc:
-                # Ideally we'd catch something less generic here.
-                raise FulcioClientError from exc
+        try:
+            # The SignedCertificateTimestamp should be accessed by the index 0
+            sct = _get_precertificate_signed_certificate_timestamps(cert)[0]
+
+        except UnexpectedSctCountException as ex:
+            raise FulcioClientError(ex)
```
jku, line 156: itching to refactor these methods a bit, using them is overly complicated and it's still incomplete (you could get  ValueError in addition to this unnecessary internal error)... but that's not really related to this PR

jku (approved):
Looks correct to me

sigstore/_internal/fulcio/client.py
woodruffw: Yeah, I'd love to have these cleaned up. Let's do a follow-up for that 🙂

PR #1247, opened 2024-12-10
fix: require an inclusion promise when log integration time is used

woodruffw:
See https://github.com/sigstore/protobuf-specs/pull/442: this strengthens the inclusion promise requirement to make it mandatory when the log integration time is used as the common verification time.

This should be a non-breaking change in principle, although in practice there might be some real-world bundles that have the inclusion promises stripped from them (similar to what we saw with bundles that had inclusion proofs but no signed checkpoints). The conformance test should determine this, since it includes all CPython bundles.

Edit: conformance passes, indicating that we're OK in terms of existing bundles having inclusion promises.

CC @haydentherapper


sigstore/models.py
```diff
@@ -543,6 +546,14 @@ def _verify(self) -> None:
             if not log_entry.inclusion_proof.checkpoint:
                 raise InvalidBundle("expected checkpoint in inclusion proof")
 
+            if (
```
haydentherapper, line 549: Do you fall back to current time if no signed timestamp is provided? 
We do this in sigstore-go and Cosign (https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/signed_entity.go#L725-L727), though it's not particularly useful because it's unlikely that a verification takes place during the certificate's limited validity window.

haydentherapper (approved):

sigstore/models.py
woodruffw: > Do you fall back to current time if no signed timestamp is provided? We do this in sigstore-go and Cosign ([sigstore/sigstore-go@`main`/pkg/verify/signed_entity.go#L725-L727](https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/signed_entity.go?rgh-link-date=2024-12-10T21%3A50%3A01Z#L725-L727)), though it's not particularly useful because it's unlikely that a verification takes place during the certificate's limited validity window.

At the moment we hard-fail during verification if we can't obtain at least one signed timestamp:

https://github.com/sigstore/sigstore-python/blob/9ee7ac259bcce37f95db13aa7d84a407c8b99a17/sigstore/verify/verifier.py#L324-L326

woodruffw: I think we could tweak that to fall back on the current time, although I'm a little murky on whether that's a better or worse UX 😅

haydentherapper: One motivation is for private deployments that have their own long-lived code signing certificates, but I know that isn't something you've tackled as part of this library. I don't think this is a requirement to add to a library.

woodruffw: > but I know that isn't something you've tackled as part of this library. I don't think this is a requirement to add to a library.

Makes sense! I think I'll leave it out for now, then, but I'll pin this to revisit.

PR #1251, opened 2024-12-11
bump rfc3161-client

woodruffw:
This bumps rfc3161-client, which will both remove a spurious `maturin` runtime dep *and* make us compatible with older versions of `cryptography`, preventing resolution hell.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
   "rich ~= 13.0",
   "rfc8785 ~= 0.1.2",
   # NOTE(dm): Under very active development, so strictly pinned.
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -36,8 +36,7 @@ dependencies = [
   "requests",
   "rich ~= 13.0",
   "rfc8785 ~= 0.1.2",
-  # NOTE(dm): Under very active development, so strictly pinned.
```
woodruffw, line 39: NB: Removed the `NOTE` here since this is now in beta and point releases won't be breaking.

jku (approved):
Nice

PR #1259, opened 2024-12-13
Minor refactoring of Fulcio SCT handling

### LINKED ISSUE #1258: refactor sct handling
There's multiple small annoyances in the way SCTs are managed internally: a bit of refactoring likely makes this simpler without affecting public API or CLI.

I have branch with some improvements, will post PR

jku:
#### Summary

Simplify SCT management in fulcio client and the sites that need to call verify_sct()
* SCTs are only needed inside `verify_sct()`: Make them an implementation detail and stop exposing them in FulcioCertificateSigningResponse (which is internal, so this is not an API change)
* The only functional change is that `FulcioCertificateSigningResponse` may now exist without a valid SCT: In practice Signer works just like before since we call `verify_sct()`  on the cert immediately after getting the response
* Also remove an unused exception

There are two commits but I recommend reviewing the PR as a whole.
Fixes #1258.

#### Release Note

N/A: No public API changes or CLI changes are expected.

### DIFF
--- a/sigstore/_internal/sct.py
+++ b/sigstore/_internal/sct.py
@@ -148,32 +148,27 @@ def _get_issuer_cert(chain: List[Certificate]) -> Certificate:
     return issuer
 
 
-class UnexpectedSctCountException(Exception):
-    """
-    Number of percerts scts is wrong
-    """
-
-    pass
-
-
-def _get_precertificate_signed_certificate_timestamps(
+def get_signed_certificate_timestamp(
     certificate: Certificate,
-) -> PrecertificateSignedCertificateTimestamps:
-    # Try to retrieve the embedded SCTs within the cert.
+) -> SignedCertificateTimestamp:
+    """Retrieve the embedded SCT from the certificate.
+
+    Raise ValueError if certificate does not contain exactly one SCT
+    """
     try:
-        precert_scts_extension = certificate.extensions.get_extension_for_class(
+        timestamps = certificate.extensions.get_extension_for_class(
             PrecertificateSignedCertificateTimestamps
         ).value
     except ExtensionNotFound:
         raise ValueError(
-            "No PrecertificateSignedCertificateTimestamps found for the certificate"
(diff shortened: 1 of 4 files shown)


jku:
Looks ok to me even with fresh eyes, marking ready for review.

@javanlacerda do you want to have a look?

jku:
I have no idea how to remove the old failed GCB check from the results: 
* this looks like a GitHub inconsistency (the commit was not yet in the git server instance that GCB was using on the first run)
* for some reason the check is uniquely named so subsequent passing checks don't override the result

di:
> I have no idea how to remove the old failed GCB check from the results

Probably not necessary since the check is not required, but I agree that's annoying!

woodruffw:
Yeah, weird -- I was thinking it might be a Pulimi thing but it's not configured there either. Hopefully it just expires at some point, but agreed with @di about it being a nonblocker 🙂

sigstore/_internal/sct.py
```diff
@@ -148,32 +148,28 @@ def _get_issuer_cert(chain: List[Certificate]) -> Certificate:
...
+        raise VerificationError(
+            "Certificate does not contain a signed certificate timestamp extension"
         )
 
-    if len(precert_scts_extension) != 1:
-        raise UnexpectedSctCountException(
-            f"Unexpected embedded SCT count in response: {len(precert_scts_extension)} != 1"
+    if len(timestamps) != 1:
+        raise VerificationError(
+            f"Expected one certificate timestamp, found {len(timestamps)}."
```
woodruffw, line 169: Nitpick: no periods in exceptons

```suggestion
            f"Expected one certificate timestamp, found {len(timestamps)}"
```

woodruffw (dismissed):
LGTM, thanks @jku! One small nitpick 🙂

jku:
rebased and fixed the suggested exception message

jku:
rebase on main

woodruffw (approved):
LGTM, thanks @jku!

PR #1283, opened 2025-01-13
ci: fix offline tests on ubuntu-latest

woodruffw:
~~WIP.~~

This "fixes" the CI by effectively undoing the namespace default changes that Ubuntu made between 22.04 and 24.04. Doing this should have no security impact in practice, since we're already running with full root access on these runners.

The other option here would be to remove `unshare` entirely, and continue to test the offline tests without actually enforcing their offline-ness. However, I think it's good to keep this around as a backstop, at least for as long as we can 🙂


jku (approved):
LG (enough) TM

PR #1293, opened 2025-01-29
ci: remove dependabot + gomod, always fetch latest

woodruffw:
Instead of using Dependabot to test for an update (which was unreliable, per #1288), this always fetches the latest TSA build.

I'm not positive this is a great idea, however -- an alternative that doesn't involve blindly pulling the latest release would be to set up a release-tracking workflow that files an issue for us whenever a new TSA release is made. That would be slightly more manual, but would also avoid opaque CI changes.


jku (approved):
The whole workflow step is indeed one big hack but let's go with this: maybe it'll be fine until there is a real TSA.

I'll update the branch: I think you'll be able to "approve" then?

jku:
hmm, looks like my approval is still valid after updating.  I'll merge.

PR #1300, opened 2025-02-10
Update embedded TUF root

jku:
This makes client a little snappier on first launch. This is the result of:

```
  python -m sigstore sign README.md # just to update local TUF cache
  cp ~/.local/share/sigstore-python/tuf/https%3A%2F%2Ftuf-repo-cdn.sigstore.dev/root.json \
      sigstore/_store/prod/root.json
  cp ~/.cache/sigstore-python/tuf/https%3A%2F%2Ftuf-repo-cdn.sigstore.dev/trusted_root.json \
      sigstore/_store/prod/trusted_root.json
```
This uses sigstore-python to update the local TUF cache during the signing, then embeds the new good TUF root (and trusted root) in the source dir

The above is also a good way for others to verify this PR is correct (just run python-sigstore from current main first, only then switch to PR branch)


woodruffw (approved):
Thanks, LGTM! I confirmed locally as well as against <https://tuf-repo-cdn.sigstore.dev/>.

Random thought: it'd be nice to have `make bump-embedded-root` or similar to automate this in the future. We could then stick it in the CI on a `cron` trigger 🙂

PR #1301, opened 2025-02-10
Makefile: Add update-embedded-root rule

jku:
* Add `make update-embedded-root` rule
* Add workflow that files an issue if root is not up-to-date

I'm not sure if the workflow is really needed but I decided to include it. I think automatically making a PR might be a bad idea (since there are test suite tests for this we'd be trusting CI 100% and that doesn't sound right in this case).

--- 

Makefile rule uses the "plumbing" command to ensure the newest root has been downloaded and verified. Then it copies the newest TUF root and the trusted_root.json into the sources. The benefit here is that one does not need to manually find the cache directories when an update should be done.

Makefile rule hard codes XDG_DATA_HOME and XDG_CACHE_HOME for simplicity.

The workflow adds a new CI-dependency (github-script) but I believe the currently used actions do not provide the capabilities needed here.

I've created the "embedded-root-update" label manually in this project already so this should just work.

### DIFF
--- a/.github/workflows/check-embedded-root.yml
+++ b/.github/workflows/check-embedded-root.yml
@@ -1,0 +1,63 @@
+name: Check embedded root
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '13 13 * * 3'
+
+jobs:
+  check-embedded-root:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+
+    steps:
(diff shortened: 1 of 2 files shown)


woodruffw (approved):
LGTM, thanks @jku! For full generality I suppose we could also override the XDG vars and do this in a scratch directory, but I'm also content with leaving it as is 🙂

woodruffw (approved):
LGTM, thanks @jku! For full generality I suppose we could also override the XDG vars and do this in a scratch directory, but I'm also content with leaving it as is 🙂

PR #1319, opened 2025-03-09
Apply new ruff rules

DimitriPapadopoulos:
#### Summary

Apply additional ruff rules, only when they improve readability. “Improvement” is debatable, of course, which is why I have a commit per rule, so that some rules can be skipped.

These new rules are not enforced at this point.

#### Release Note

Not really worth a changelog entry, since these are mostly automatic changes.

#### Documentation

No need to change the documentation.


woodruffw (approved):
Thanks @DimitriPapadopoulos! IIRC we were avoiding some of the newer-style type hints because of problems with Pydantic, but those may no longer be relevant. I'll let the CI shake out any issues here.

woodruffw:
Yep, my memory was correct:

```
 E   TypeError: Unable to evaluate type annotation 'StrictStr | None'. If you are making use of the new typing syntax (unions using `|` since Python 3.10 or builtins subscripting since Python 3.9), you should either replace the use of new syntax with the existing `typing` constructs or install the `eval_type_backport` package.
make: *** [Makefile:82: test] Error 4
```

I'd be okay with either adding `eval_type_backport` or reverting that particular `ruff` rule. PR author's choice :slightly_smiling_face:

sigstore/dsse/_predicate.py
```diff
@@ -17,7 +17,7 @@
 """
 
 import enum
-from typing import Any, Dict, List, Literal, Optional, TypeVar, Union
+from typing import Any, Literal, Optional, Self, Union
```
jku, line 20: Self is too new unfortunately, python >= 3.11

DimitriPapadopoulos: I believe ruff wouldn't suggest `Self` if it weren't supported by supported Python versions:
https://github.com/sigstore/sigstore-python/blob/aa8e4587384c3be3be31aa73135c5ed59e9ad168/pyproject.toml#L46

Let me see the documentation for [custom-type-var-for-self (PYI019)](https://docs.astral.sh/ruff/rules/custom-type-var-for-self/#custom-type-var-for-self-pyi019):
> ### [Fix behaviour and safety](https://docs.astral.sh/ruff/rules/custom-type-var-for-self/#fix-behaviour-and-safety)
> 
> The fix replaces all references to the custom type variable in the method's header and body with references to `Self`. The fix also adds an import of `Self` if neither `Self` nor `typing` is already imported in the module. If your [`target-version`](https://docs.astral.sh/ruff/settings/#target-version) setting is set to Python 3.11 or newer, the fix imports `Self` from the standard-library typing module; otherwise, the fix imports `Self` from the third-party [`typing_extensions`](https://typing-extensions.readthedocs.io/en/latest/) backport package.

I think I might have used `--preview` which might be broken. I have manually changed:
```python
from typing import Self
```
to:
```python
from typing_extensions import Self
```
Hopefully this will fix this issue.

DimitriPapadopoulos:
> I'd be okay with either adding `eval_type_backport` or reverting that particular `ruff` rule. PR author's choice 🙂

I've removed that rule. According to [non-pep604-annotation-union (UP007)](https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/#non-pep604-annotation-union-up007):
> This rule is enabled when targeting Python 3.10 or later (see: [`target-version`](https://docs.astral.sh/ruff/settings/#target-version)). By default, it's also enabled for earlier Python versions if `from __future__ import annotations` is present, as `__future__` annotations are not evaluated at runtime. If your code relies on runtime type annotations (either directly or via a library like Pydantic), you can disable this behavior for Python versions prior to 3.10 by setting [`lint.pyupgrade.keep-runtime-typing`](https://docs.astral.sh/ruff/settings/#lint_pyupgrade_keep-runtime-typing) to `true`.

jku (approved):
nice, thanks!

PR #1320, opened 2025-03-09
ruff settings: select → extend-select

DimitriPapadopoulos:
#### Summary

Instead of explicitly selecting `E` and `F` rules, use ruff default rules, which include `F` rules and `E` rules compatible with formatters:
	https://docs.astral.sh/ruff/settings/#lint_extend-select

According to the Scientific Python Library Development Guide, `W` rules are not needed when using a formatter:
	https://learn.scientific-python.org/development/guides/style/#ruff

Also enforce `UP` ruff rules.

Requires #1319.

#### Release Note

I might be mistaken, but probably not needed.

#### Documentation

N/A

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,8 +121,6 @@ plugins = ["pydantic.mypy"]
 exclude_dirs = ["./test"]
 
 [tool.ruff.lint]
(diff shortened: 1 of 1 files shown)


woodruffw (approved):
Thanks!

PR #1321, opened 2025-03-10
Apply ruff/pygrep-hooks rule PGH003

DimitriPapadopoulos:
#### Summary

```
PGH003 Use specific rule codes when ignoring type issues
```
Fix `type: ignore` annotations that suppress all type warnings, as opposed to targeting specific type warnings.

#### Release Note

Irrelevant for end users.

#### Documentation

N/A


woodruffw:
Thanks, looks good. I'm somewhat surprised our `mypy` settings didn't catch this -- can we enable this rule so that future bare ignores are caught earlier?

DimitriPapadopoulos:
`PGH003` is a ruff rule, but it will also be caught by the MyPy setting `enable_error_code = ["ignore-without-code"]` in #1322.

I may actually enforce more ruff rules, including `PGH`, in a later PR.

woodruffw (approved):

Issue #1040, opened 2024-06-07
Refactor: replace `requests` with `urllib3`


woodruffw:
`urllib3` has everything we need from `requests` (and is the library beneath it), including connection pooling. It's much more actively maintained and has a more consistent API, so we should switch to it.

woodruffw:
> * urllib3 defaults to system certificates while requests defaults to certifi

Yep, this is a "plus" IMO: the fact that PyPI is a CA bundle delivery mechanism (for `certifi`) has always been one of those "don't look but this is very scary" realities about HTTPS in Python 🙂

jku:
I think that sounds very reasonable.

@sethmlarson if you have the time, would you mind commenting on this one? 
* are we overlooking any potential problems when switching from certifi to system certs
* should we (as library) mention truststore in docs?
* should we (as CLI app) use truststore explicitly?

jku:
The PR linked above makes this urllib3 change for our tuf traffic.

WRT proxy env variables: I added the component (that I kind of think should exist in urllib3) in python-tuf: https://github.com/theupdateframework/python-tuf/blob/develop/tuf/ngclient/_internal/proxy.py. It's pretty small so it's not out of the question to just copy paste here if we do want to support these env vars.

PR #1323, opened 2025-03-11
Upgrade to tuf 6.0

jku:
This is not really urgent but I wanted to open this before the dependabot update comes in.

I think the potentially annoying part here is the certificate source: it's not ideal that some parts of sigstore-python will now be using certifi certs and the tuf client will be using system certs... but the choice made here is in line with #1040 

### Details 

The big changes in tuf 6.0 are dependencies:
* no longer uses requests but directly urllib3
* no longer uses certifi

The included code change in this PR is not strictly necessary but is potentially a little more secure (tuf client no longer trusts the metadata cache but always initializes the Updater from the embedded root).

The test changes are related to
* assumptions we make on the tuf metadata cache content
* our tests mocking a feature that was changed in TUF (RequestsFetcher got changed to Urllib3Fetcher)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
   # NOTE(ww): Both under active development, so strictly pinned.
   "sigstore-protobuf-specs == 0.3.2",
   "sigstore-rekor-types == 0.0.18",
-  "tuf ~= 5.0",
+  "tuf ~= 6.0",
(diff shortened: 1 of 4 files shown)


woodruffw (approved):
LGTM! Agreed about the root program stuff not being ideal, but IMO it's acceptable pending #1040.

PR #1350, opened 2025-04-11
_internal/trust: Fix bug in rekor key lookup

### LINKED ISSUE #1349: multiple rekor keys should be allowed
There's currently a safety check for exactly one rekor key in sigstore-python: this is incorrect as logs will be retired (after which they'll remain valid verification logs) and once sigstore/rekor-tiles is integrated, we  will actually have two active logs in staging/prod that clients can choose from for a while.

jku:
Rekor keyring can (and in future will) have multiple keys: logs not only get sharded but once rekor-tiles is integrated in the public good instance, there will be two writable logs for a while.

As far as I can tell all calling code is already capable of handling the keyring.

Fixes #1349

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+* Fixed issue where a trust root with multiple rekor keys was not considered valid:
+  Now any rekor key listed in the trust root is considered good to verify entries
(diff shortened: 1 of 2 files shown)


woodruffw (approved):
LGTM!

woodruffw:
@jku actually, could you add a small CHANGELOG entry too? I imagine users of older versions might hit this once Sigstore shards and a CL entry will point them in the right direction.

woodruffw (approved):

PR #1276, opened 2025-01-10
deps: bump protobuf-specs

woodruffw:
Very WIP. This includes underlying changes to the proto models themselves, so I need to go through various internal usages and correct them.

On the plus side, this makes our typechecking of each proto model's internals much more correct.

PR #1358, opened 2025-04-29
More proto updates

jku:
This is built on top of #1276: I'm fine with merging that first or closing that one and merging all the proto changes at once in this PR.

This PR upgrades us to the most recent protobuf-specs. Changes include (on top of 1276):
* A new signingconfig version is supported, and a `SigningConfig` class has been added to handle parsing the data in it
* I'm removing support for SigningConfig 0.1 since that was not actually usable: I don't think anyone will mind (staging currently has 0.1 but we do not use it and 0.2 should be available in staging by next week)

I'd like to handle followups in other PRS, mainly I'm thinking of:
* rekorclient is still a bit of a mess
* the signingconfig service selectors are not fully supported yet
* oidc provider is still hard coded when running CLI
* signingconfig from TUF is not yet used: There is still a discussion on what the specific TUF targetpath is going to be

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ All versions prior to 0.9.0 are untracked.
 * API: Make Rekor APIs compatible with Rekor v2 by removing trailing slashes
   from endpoints ([#1366](https://github.com/sigstore/sigstore-python/pull/1366))
 
+### Changed
+
+* `--trust-config` now requires a file with SigningConfig v0.2, and is able to fully
+  configure the used Sigstore instance [#1358]/(https://github.com/sigstore/sigstore-python/pull/1358)
+
 ## [3.6.2]
 
 ### Fixed
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
   "rfc8785 ~= 0.1.2",
   "rfc3161-client >= 0.1.2,< 1.1.0",
   # NOTE(ww): Both under active development, so strictly pinned.
-  "sigstore-protobuf-specs == 0.3.2",
+  "sigstore-protobuf-specs == 0.3.5",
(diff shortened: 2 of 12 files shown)


jku:
FYI @ramonpetgrave64: 
* this should work with new signing config format ....
* but still doesn't make oidc provider selection use the signing config -- I believe that's a minor patch, CLI just needs a small refactor

BTW, now that I had a look at it, I think the current RekorClient is a bit of a mess: it probably needs a redesign even without rekor-tiles.

woodruffw:
Thanks @jku! I'll review now.

> This is built on top of #1276: I'm fine with merging that first or closing that one and merging all the proto changes at once in this PR.

We can close #1276; you've made much more progress than I've had time to 🙂 

> * I'm removing support for SigningConfig 0.1 since that was not actually usable: I don't think anyone will mind (staging currently has 0.1 but we do not use it and 0.2 should be available in staging by next week)

Sounds good to me!

woodruffw (dismissed):
LGTM! Thanks a ton for taking over the old PR and pushing this through + making the `SigningConfig` changes. If I'm not mistaken we shouldn't need a CHANGELOG entry here, since this should all be internal API changes 🙂

jku:
> If I'm not mistaken we shouldn't need a CHANGELOG entry here, since this should all be internal API changes 🙂

The one thing I now started thinging about is `--trust-config`: I just dropped the signingconfig 0.1 support in this PR so if anyone was using --trust-config already then the files they were previously using will not work anymore... I will add a Changelog mention

jku:
@woodruffw or @di a final review would be welcome (I only added a changelog line after last review)

jku:
rebased, fixed conflicts

woodruffw (approved):
LGTM, thanks @jku!

PR #1363, opened 2025-05-05
Use trust config everywhere

### LINKED ISSUE #1347: Use signingconfig from "--trust-config"
Currently I believe we don't use all of the information in the signingconfig part of `--trust-config` (or the signingconfig from TUF). It would be useful for #1340 if we did as we can then try to provide a "sigstore-in-a-box" test setup that comes with a trustedroot and signingconfig.

Basically, if we have signingconfig we should select the used services from there when signing 
* interactive oidc url: this needs work
* fulcio url (EDIT: this is already covered when using --trust-config)
* rekor urls (EDIT: this is already covered when using --trust-config)

In addition:
* --production and --staging should actually use the signing config from TUF

To fix these issues:
* `_internal.trust` has to contain a SigningConfig implementation and should use a method provided by `_internal.tuf` to fetch it (special casing is currently needed for production since there is no signingconfig yet). I have an almost ready branch for this.
* #1276 is sort of a blocker: signingconfig has had changes and it would probably make sense to only support the current version

jku:
Use TrustConfig throughout the project

This looks complete to me now. I'm keeping it a draft while 
* #1358 merges 
* I want to review the API a bit: Maybe the internal components (like SigningConfig) don't actually have to know about ClientTrustConfig and could just get a SigningConfig instead...
* I will think about changelog entry

Apologies for the seemingly large PR. It's mostly just the assets getting updated: this update is done to get test coverage for signingconfig handling.

Fixes #1347, fixes #1347

* Use `TrustConfig.signing_config` to get the OIDC url
* Provide the `from_tuf()` method for TrustConfig and not Trustedroot
* Use TrustConfig as the initialization source where that makes sense (OIDC, SigningContext)
* Refactor embedded asset lookup: directories are now based on repository URLs so there's less hard coded if clauses for Public Good instance
* Update the embedded staging assets and the testing staging assets: This means the staging signing config is used to test the code in unit tests


jku:
> I want to review the API a bit: Maybe the internal components (like SigningConfig) don't actually have to know about ClientTrustConfig and could just get a SigningConfig and/or TrustedRoot instead...

It looks a bit messy but I think it's fine:
* Verifier takes a `TrustedRoot` as argument: this makes some sense since it only needs the trust root
* SigningContext takes a `ClientTrustConfig` since it at least currently requires data from both signing config and trust root
* Issuer now just takes a url as argument (calling `SigningConfig.get_oidc_url()` is simple enough, no need to have a special constructor)

jku:
CC @ramonpetgrave64

jku:
ah, I forgot that if I update the staging test assets I need to handle the TUF timestamp expiring... I think that's just a minor monkeypatch but it does mean this is not yet ready for review

jku:
sigh, now the tests are failing because timestamp-authority is doing a release and our tests are expecting it to be available already because the tags are there...

I will continue on monday

jku:
I've marked this ready for review. It's a big patch but only 200 lines are code changes (rest is assets). I can try to split this but it might not make it more easier to manage.

sigstore/oidc.py
```diff
@@ -271,20 +268,6 @@ def __init__(self, base_url: str) -> None:
...
-        Returns an `Issuer` configured against Sigstore's production-level services.
-        """
-        return cls(DEFAULT_OAUTH_ISSUER_URL)
-
-    @classmethod
-    def staging(cls) -> Issuer:
-        """
-        Returns an `Issuer` configured against Sigstore's staging-level services.
-        """
-        return cls(STAGING_OAUTH_ISSUER_URL)
```
woodruffw, line 286: NB: This is a breaking API change, so we'll need a CHANGELOG entry documenting its removal + to plan it for a new major release (I'm okay with doing that major release whenever 🙂)

sigstore/sign.py
```diff
@@ -324,29 +324,7 @@ def __init__(
...
-    @classmethod
-    def staging(cls) -> SigningContext:
-        """
-        Return a `SignerContext` instance configured against Sigstore's staging-level services.
-        """
-        return cls(
-            fulcio=FulcioClient.staging(),
-            rekor=RekorClient.staging(),
-            trusted_root=TrustedRoot.staging(),
-        )
```
woodruffw, line 346: Same note about public API + breaking changes as above!

test/unit/internal/oidc/test_issuer.py
```diff
@@ -33,4 +33,4 @@ def test_get_identity_token_bad_code(monkeypatch):
     monkeypatch.setattr("builtins.input", lambda _: "hunter2")
 
     with pytest.raises(IdentityError, match=r"^Token request failed with .+$"):
-        Issuer.staging().identity_token(force_oob=True)
+        Issuer("https://oauth2.sigstage.dev/auth").identity_token(force_oob=True)
```
ramonpetgrave64, line 36: Add docstring explaining why this raises an exception.

sigstore/_internal/fulcio/client.py
```diff
@@ -39,8 +39,6 @@
 
 _logger = logging.getLogger(__name__)
 
-DEFAULT_FULCIO_URL = "https://fulcio.sigstore.dev"
-STAGING_FULCIO_URL = "https://fulcio.sigstage.dev"
```
ramonpetgrave64, line 43: Also remove from default URLs from `sigstore-python/sigstore/_internal/rekor/client.py`

sigstore/_internal/trust.py
```diff
@@ -386,9 +387,7 @@ def get_tsa_urls(self) -> list[str]:
         but may return more in future.
         """
         url = self._get_valid_service_url(self._inner.tsa_urls)
```
ramonpetgrave64, line 389: For rekor urls, I think `get_valid_service_url()` needs updating to potentially support a major version 2

- https://github.com/sigstore/sigstore-python/blame/baa95acc60082ab37d22813890085d384c8a6d37/sigstore/_internal/trust.py#L344

ramonpetgrave64 (changes requested):

sigstore/_internal/trust.py
jku: sure, agreed. Or maybe rather rekor and tsa cases should not use `_get_valid_service_url()` at all and should do their own things that actually return more meaningful results in their cases: both should return multiple results after all

sigstore/sign.py
jku: yeah, I should probably hide the similar methods in RekorClient now as well (they are still used by the `fix-bundle` special case but I could make them more private)

ramonpetgrave64 (approved):

CHANGELOG.md
```diff
@@ -31,8 +31,25 @@ All versions prior to 0.9.0 are untracked.
...
+  comes from the TUF repository [#1363](https://github.com/sigstore/sigstore-python/pull/1363)
+
+### Removed
+ * API:
+  * `Issuer.production()` and `Issuer.staging()` have been removed: Use
+    `Issuer()` instead with relevant URL. The current public good production and
+    staging URLs are available via the `ClientTrustConfig` object.
+    [#1363](https://github.com/sigstore/sigstore-python/pull/1363)
+  * `Signingcontext.production()` and `Signingcontext.staging()` have been removed:
+    Use `Signingcontext.from_trust_config()` instead.
```
woodruffw, line 50:
```suggestion
  * `SigningContext.production()` and `SigningContext.staging()` have been removed:
    Use `SigningContext.from_trust_config()` instead.
```

sigstore/_cli.py
```diff
@@ -1167,7 +1158,27 @@ def _verify_common(
...
-def _get_identity(args: argparse.Namespace) -> Optional[IdentityToken]:
+def _get_trust_config(args: argparse.Namespace) -> ClientTrustConfig:
+    """
+    Return the client trust configuration (Sigstore service URLs, key material and lifetimes)
+
+    The configuration may come from explicit argument (--trust-config) or from the TUF
+    repository of the used Sigstore instance.
+    """
+    # Some commands do not have offline argument, parse carefully
+    offline = args.offline if "offline" in args else False
```
woodruffw, line 1169: Nitpick: I think this is a little more idiomatic:

```suggestion
    # Not all commands provide --offline
    offline = getattr(args, "offline", False)
```

(We could also make `--offline` a global argument to avoid this entirely -- I would have no objection to that 🙂)

sigstore/_internal/trust.py
```diff
@@ -575,6 +536,65 @@ def from_json(cls, raw: str) -> ClientTrustConfig:
...
+        try:
+            sc_path = updater.get_signing_config_path()
+            inner_sc = _SigningConfig().from_json(Path(sc_path).read_bytes())
+        except TUFError as e:
+            # TUF repo may not have signing config yet: hard code values for prod:
+            if url == DEFAULT_TUF_URL:
+                embedded = read_embedded("signing_config.v0.2.json", url)
+                inner_sc = _SigningConfig().from_json(embedded)
+            else:
+                raise e
```
woodruffw, line 588: Just confirming my understanding: we're doing this because the production TUF instance doesn't have the v2 signing config yet, right? So we'll be able to remove this once it's deployed there?

woodruffw: (And if that's right, is there a tracking issue/PR we can xref here?)

jku: correct. I will file an issue

jku (commented):
I should update make target update-embedded-root since I'm changing `_store` directory names 

EDIT: fixed

sigstore/_store/https%3A%2F%2Ftuf-repo-cdn.sigstage.dev/root.json
woodruffw, line 1: I don't _love_ that we now have URL-encoded filenames checked into the project, but I don't have a great alternative here. Maybe we could trim the `https://` prefix, although that could be a waste of time?

woodruffw (approved):
LGTM, thanks @jku!

I left one non-blocking comment about the URL-encoded filenames, but we can merge if you don't want to bother with that.

sigstore/_store/https%3A%2F%2Ftuf-repo-cdn.sigstage.dev/root.json
jku: yeah, it's not great.

I think I'll leave the https there (we don't currently test with a local http server but if someone ever wants to, assuming https  would be a blocker). Not a strong opinion though, can do that change anyone really doesn't like it.

Issue #1365, opened 2025-05-06
Trailing slash in "/entries" path


ramonpetgrave64:
[Client support for Rekor V2: sigstore-python](https://github.com/sigstore/rekor-tiles/issues/289)

**Description**

For the endpoint to post new entries, the client will have a trailing slash. This does not seem to be according to rekor V1 spec, but will still work. The spec is the same in rekorV2, but currently does not work in rekor V2 servers.

- https://github.com/sigstore/rekor/blob/140f622a85d9eeb72677f6ab7de21744954eb4cc/pkg/generated/client/entries/entries_client.go#L97
- https://github.com/sigstore/rekor-tiles/blob/364926072f047df0cd4a9c3285c52e2f7f2d5dc9/api/proto/rekor_service.proto#L32-L37

PR #1366, opened 2025-05-06
no trailing slash for post to /entries

ramonpetgrave64:

#### Summary

Resolves https://github.com/sigstore/sigstore-python/issues/1365

by stripping the any trailing slash from the POST request.

#### Release Note

Removed trailing "/" in POST requests to "/entries".

#### Documentation

None


ramonpetgrave64:
@jku

jku (commented):
I think you're right here, thanks.

* I'd make the fix in `RekorLog.entries()` though: no need to rstrip() if we never add the slash in the first place. The other use cases of url look like they will keep working just fine
* Could also do the same fix of not adding the trailing slash in `RekorEntries.retrieve()`  and `RekorClient.log()` --- the situations look identical there (I don't think these end points get used in normal client use so they're not important but IMO makes sense to fix all the cases)

WDYT?

jku (changes requested):
Please add a line into CHANGELOG as well.

(I'm marking this "request changes" for the changelog: feel free to make your own decision on the suggestions in previous comment)

woodruffw:
Hey @ramonpetgrave64, thanks for the PR!

> * I'd make the fix in `RekorLog.entries()` though: no need to rstrip() if we never add the slash in the first place. The other use cases where the url is used look like they will keep working just fine

Yeah, I think I'd prefer a "never add the slash" approach here versus "add it and then remove it sometimes" -- IMO that'll be easier to read and follow.

(In terms of doing that -- I think we can just modifying the underlying `urljoin` calls that build up these different routes.)

ramonpetgrave64:
`urljoin` doesn't work like other "join" functions. Removing the slashes altogether would make everything more complicated when we need to add the slashes. If you both still prefer, I could do it the way you suggest.

- https://stackoverflow.com/questions/10893374/python-confusions-with-urljoin

woodruffw:
> `urljoin` doesn't work like other "join" functions. Removing the slashes altogether would make everything more complicated when we need to add the slashes. If you both still prefer, I could do it the way you suggest.

Ah yeah, I forgot about the wonky join behavior even when there's no leading or trailing `/`. Sigh.

Given that, I think maybe we'd be better off uisng f-strings here and building up the URL that way, to avoid having to strip things post facto. Alternatively, we could use the [rfc3986](https://rfc3986.readthedocs.io/en/latest/index.html) package (which we probably already indirectly depend on), since it has `URIBuilder.add_path` and `URIBuilder.extend_path` for exactly this purpose.

ramonpetgrave64:
@woodruffw [rfc3986](https://rfc3986.readthedocs.io/en/latest/index.html) isn't currently a dependency, so I changed to using f-strings.

sigstore/_internal/rekor/client.py
```diff
@@ -112,7 +112,7 @@ def entries(self) -> RekorEntries:
         Returns a `RekorEntries` capable of accessing detailed information
         about individual log entries.
         """
-        return RekorEntries(urljoin(self.url, "entries/"), session=self.session)
+        return RekorEntries(urljoin(self.url, "entries"), session=self.session)
```
woodruffw, line 115: Sorry, I might have been unclear -- I was saying that we could remove `urljoin` entirely and do this:

```suggestion
        return RekorEntries(f"{self.url}/entries", session=self.session)
```

(And similarly for all other `urljoin` uses below.)

ramonpetgrave64: removed urljoin

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+* Fixed the POST path for "/entries" to not have a trailing slash.
```
woodruffw, line 11: Needs a subcategory + backlink:

```suggestion
### Fixed

* API: Make Rekor APIs compatible with Rekor v2 by removing trailing slashes
  from endpoints ([#1366](https://github.com/sigstore/sigstore-python/pull/1366))
```

woodruffw (approved):
Thanks @ramonpetgrave64!

@jku will need to re-review since I've tweaked the CHANGELOG, but this LGTM.

Issue #1372, opened 2025-05-08
TSA request message digest defaults to sha256


ramonpetgrave64:
[Client support for Rekor V2: sigstore-python](https://github.com/sigstore/rekor-tiles/issues/289)

**Description**

The request to the Timestamp Authority defaults to sha512,

- https://github.com/sigstore/sigstore-python/blob/2199d9b5d4c6aa1c6f5c6b63a18299fdedef27b0/sigstore/_internal/timestamp.py#L95-L97
- https://github.com/trailofbits/rfc3161-client/blob/c331b83648355f57d529c1a7d3f4616637a8aa0f/src/rfc3161_client/base.py#L80-L81

but then when verifying, assumes sha256.

- https://github.com/sigstore/sigstore-python/blob/2199d9b5d4c6aa1c6f5c6b63a18299fdedef27b0/sigstore/verify/verifier.py#L188-L190

We should either explicitly request sha256, or implement the example logic to determine the what the correct message digest was. Or, wait for a fix in rfc3163-client's `verify()` that accepts a message, instead of a digest.

 - https://github.com/trailofbits/rfc3161-client/blob/c331b83648355f57d529c1a7d3f4616637a8aa0f/README.md?plain=1#L93-L97
 - https://github.com/trailofbits/rfc3161-client/issues/130

**Version**

sigstore-python v3.6.2

Using TSA timestamp.sigstage.dev

jku:
Well that seems like an obvious bug -- do we not have a test where we verify a timestamp produced/requested by sigstore-python?

> We should either explicitly request sha256, or implement the example logic to determine the what the correct message digest was

I assume sigstore specs specify  (or should specify ) what the correct choice here is? We want the bundles produced by sigstore-python to be verifiable by other clients (and vice versa): making sigstore-python work with sigstore-python is only a partial win...

PR #1373, opened 2025-05-08
Specify sha256 in TSA request

ramonpetgrave64:

#### Summary

Resolves https://github.com/sigstore/sigstore-python/issues/1372

Makes the TSA request specify sha256 for the message digest, ~~since verification currently assumes sha256.~~ Verification shouldn't assume any specific algorithm, but I think for signing requests this library should specify an algorithm, for the sake of predictability. 

#### Release Note

* TSA: Changed the Timestamp Authority requests to explicitly use sha256 for message digests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All versions prior to 0.9.0 are untracked.
 ### Fixed
 
 * TSA: Changed the Timestamp Authority requests to explicitly use sha256 for message digests.
+  [#1371](https://github.com/sigstore/sigstore-python/pull/1371)
 
 * Fixed the certificate calidity period check for Timestamp Authorities (TSA).
   Certificates need not have and end date, while still requiring a start date.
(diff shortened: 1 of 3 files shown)


jku:
> Makes the TSA request specify sha256 for the message digest, since verification currently assumes sha256.

I mentioned this in the issue but for completeness will repeat: I think we should fix verification directly since we can't just support timestamps made by sigstore-python so can't enforce which algorithms we will see. https://github.com/trailofbits/rfc3161-client/pull/144 may be something usable?

ramonpetgrave64:
@jku I agree. Verification shouldn't assume any specific algorithm, but I think for signing requests this library should specify an algorithm, for the sake of predictability.

jku:
> @jku I agree. Verification shouldn't assume any specific algorithm, but I think for signing requests this library should specify an algorithm, for the sake of predictability.

Oh yeah that is a better reason for this PR than "verification currently assumes sha256".  SGTM.

jku (approved):
I agree, we should be choosing the algorithm explicitly when creating a timestamp. SHA-256 sounds fine to me and we can always bump that up (preferably after verification is ok with it).

I'll file a new issue for the verification hash being hard coded so we don't forget

PR #1382, opened 2025-05-12
Fix pydantic in `_to_rekor()` for optional inclusion proofs

ramonpetgrave64:
#### Summary

pydantic and betterproto: avoid assigning a var of type Optional[InclusionPromise] to TransparencyLogEntry.inclusion_promise
    
 ```
      File "/usr/local/google/home/rpetgrave/src/ssp/sigstore-python/sigstore/models.py", line 262, in _to_rekor
        tlog_entry = rekor_v1.TransparencyLogEntry(
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/local/google/home/rpetgrave/src/ssp/.venv/lib/python3.12/site-packages/pydantic/_internal/_dataclasses.py", line 120, in __init__
        s.__pydantic_validator__.validate_python(ArgsKwargs(args, kwargs), self_instance=s)
    pydantic_core._pydantic_core.ValidationError: 1 validation error for TransparencyLogEntry
    inclusion_promise
      Inp
```

#### Release Note

Avoid pydantic's instantiation issues with `TransparencyLogEntry` when `InclusionPromise` is not present.

#### Documentation

None

### CODE REVIEW

**Comment by ramonpetgrave64** [comment]
@jku @woodruffw

File: CHANGELOG.md [hunk 1]
```diff
@@ -10,6 +10,8 @@ All versions prior to 0.9.0 are untracked.
 
 ### Fixed
 
+* Avoid pydantic's instantiation issues with `TransparencyLogEntry` when `InclusionPromise` is not present.
+
```
jku, line 14: I don't think this is really related to pydantic (although I appreciate the neater handling in `_to_rekor()`). Isn't the issue that our code just won't work without an inclusion promise currently?

test/unit/test_models.py
```diff
@@ -42,6 +42,20 @@ def test_missing_inclusion_proof(self):
                 inclusion_promise=None,
             )
 
+    def test_missing_inclusion_promise_round_trip(self, signing_bundle):
+        """
```
jku, line 46: Shouldn't this use a bundle without an inclusion promise and test with that? I'm guessing deserializing that bundle will still fail, based on looking at `Bundle._verify()`.  Maybe you are planning to fix that in a future PR?  It would make sense to test a complete bundle round trip here but let me know what you think...

I think signing with current staging and just manually removing the inclusion promise (and maybe also integration time) from the bundle file should work?

jku (changes requested):
Making Bundle work without an inclusion promise makes sense, however 
*  this seems to be only part of that -- I think fixing and testing the whole Bundle round trip in this PR makes sense (at least `Bundle._verify()` seems to need work), WDYT?
* I don't think this is a pydantic issue, just us not being prepared for no inclusion promise

test/unit/test_models.py
jku: scratch that, I didn't read that fully

test/unit/test_models.py
```diff
@@ -42,6 +42,20 @@ def test_missing_inclusion_proof(self):
                 inclusion_promise=None,
             )
 
+    def test_missing_inclusion_promise_round_trip(self, signing_bundle):
```
jku, line 45: ok, a second try at this comment after more reading:

Is it possible to make this test work over a Bundle "round trip" -- either using `Bundle.from_parts()`  or creating a real Bundle file in assets that does not have inclusion promise (and even no integration time)?

I imagine the latter may be doable by just signing with current staging and removing the inclusion promise and integration time fields from the file.

ramonpetgrave64: Also removed the `integrated_time` in the existing test.

I want to add this test to test_verifier.py, but it depends on the changes from #1381, but maybe this test should instead be submitted in that PR.

```python
@@ -222,6 +222,19 @@ class TestVerifierWithTimestamp:
             null_policy,
         )
 
+    def test_vierifier_verify_no_inclusion_promise_and_integrated_time(self, verifier, asset, null_policy):
+        bundle = Bundle.from_json(
+            asset("tsa/bundle.txt.sigstore").read_bytes())
+        _dict = bundle.log_entry._to_rekor().to_dict()
+        del _dict["inclusionPromise"]
+        del _dict["integratedTime"]
+        bundle._log_entry = LogEntry._from_dict_rekor(_dict)
+        verifier.verify_artifact(
+            asset("tsa/bundle.txt").read_bytes(),
+            bundle,
+            null_policy,
+        )
+
```

CHANGELOG.md
ramonpetgrave64: removed mention of pydantic.

jku (approved):
This seems correct. The test is a little lacking but is handled in #1381

jku:
Sigh, the merge rules here are really annoying... 
* I can't merge out-of-date branches
* If I update the branch I think I'll be the latest committer, so then can't approve again

I'll test that and update the branch anyway.

jku:
oh, merging main does not invalidate the review -- I did not know that.

So I can merge this even though I'm the last committer :shrug:

PR #1385, opened 2025-05-15
Verifier: Use correct Timestamp hash algorithm

jku:
DRAFT until rfc3161-client has a release. This is not urgent, I'm just filing to keep track.

---

Don't assume sha256. Use verify_message() instead: it looks up the correct hash from the the timestamp response.

TODO: this requires unreleased rfc3161-client.


woodruffw (approved):
LGTM, thanks @jku!

(Needs deconflict but otherwise good to go.)

woodruffw (approved):
Thanks!

PR #1387, opened 2025-05-16
[DRAFT] Rekor v1 and v2 support via signing-config

ramonpetgrave64:

#### Summary

Adds Rekor V2 support, while being compatible with Rekor v1, swicthable with the signing-config, when signing and verify an artifact.

Lots of files in this PR:

- New python protos, generated from rekor-tiles, copied into sigstore-python. We may not need all these files. And thy maybe belong in [trailofbits/rekor-types](https://github.com/trailofbits/sigstore-rekor-types), or even sigstore/protobuf-specs.
  - https://github.com/sigstore/rekor-tiles/pull/303
    - and pending https://github.com/sigstore/rekor-tiles/pull/305
  - related: https://github.com/trailofbits/sigstore-rekor-types/pull/168
- trust configs
  - with rekor.sigstage.dev rekor v1: trust_config.json
  - with a local rekorv2: trust_config_v2.json
    - with a new release of rekor-tiles, v2 should be also deployed to rekor.sigstage.dev for testing
 - changes from pedning PRs
   -  #1370
   - #1381 
   - #1382

### Testing

Unit tests pass ~~locally with python 3.12.9. CI tests fails on python 3.9~~. And linters are not yet expected to pass.

Signing and verifying both work.

Invoke with 

```shell
sigstore --trust-config ./trust_config.json sign README.md --overwrite -v

sigstore --trust-config ./trust_config.json verify identity --bundle README.md.sigstore.json --cert-oidc-issuer https://accounts.google.com --cert-identity <your cert identity> README.md --verbose
```

```shell
sigstore --trust-config ./trust_config_v2.json sign README.md --overwrite -v

sigstore --trust-config ./trust_config_v2.json verify identity --bundle README.md.sigstore.json --cert-oidc-issuer https://accounts.google.com --cert-identity <your cert identity> README.md --verbose
```

### TODO

Lots of items, so the last few items regarding testing will probably come in a separate PR.

- [x] make it work in CI on all python versions
  - [ ] maybe need to edit the generator in rekor-tiles.
- [ ] signing and verifying with dsse
- [ ] find a home for the types
- [ ] tests for all the branches that switch between v1 and v2
  - [ ] example v2: bundles and v2 TLEs and v2 canonicalized bodies
  - [ ] client endpoint switching
  - [ ] parsing switching
- [ ] find a way to programmatically detect the key type, even though we do hardcode a specific type for signing and verifying
- [ ] use new changes in rfc3161 client from https://github.com/sigstore/sigstore-python/pull/1399
- [ ] are major versions 1 and 2 the only acceptable versions? https://github.com/sigstore/sigstore-python/issues/1390, https://github.com/sigstore/sigstore-python/issues/1388
  - [ ] refactor the rekor, fulcio, etc, clients to accept a `v1.Service` in their constructors, which has both `url` and `major_api_version`. https://github.com/sigstore/sigstore-python/issues/1396
  - [ ] integration tests against v2, at rekor.sigstage.dev, or with the new scaffodling/setup-sigstage-env Action
  - [ ] test verifying bundles from other clients, perhaps only v2 bundles, for now
  - [ ] Decide if the user may also switch to rekor V2 via another CLI option, other than trust-config.


sigstore/sign.py
```diff
@@ -355,7 +409,10 @@ def _from_trust_config(cls, trust_config: ClientTrustConfig) -> SigningContext:
         signing_config = trust_config.signing_config
         return cls(
             fulcio=FulcioClient(signing_config.get_fulcio_url()),
-            rekor=RekorClient(signing_config.get_tlog_urls()[0]),
+            rekor=RekorClient(
+                signing_config.get_tlog_urls()[0],
+                signing_config._inner.rekor_tlog_urls[0].major_api_version,
+            ),
```
jku, line 415: my initial idea was to make get_tlog_urls() take a majorversion argument: so this code would do something like

```
logs=signing_config.get_tlogs(2)
if logs:
 rekor = Rekor2Client(v2_logs[0])
 else:
  logs = signing_config.get_tlogs(1)
  rekor = RekorClient(v2_logs[0])
 
```

obviously you can make it work with a single RekorClient as well, that's just to show the idea

sigstore/sign.py
```diff
@@ -293,6 +310,43 @@ def sign_artifact(
             ),
         )
 
+        # Create the proposed hashedrekord entry
+        if self._signing_ctx._rekor.major_api_version == REKOR_V1_API_MAJOR_VERSION:
```
jku, line 314: What do you think about this:  RekorClient provides a `build_proposed_entry()` that takes artifatc_signature, b64_cert and hashed_input (and dsse/hashedrekord selection) as argument -- then this code (and the dsse code) could just call `self._signing_context._rekor.build_proposed_entry()`

ramonpetgrave64: That's a good idea, especially since this patch is building the request body.

sigstore/sign.py
ramonpetgrave64: There is no `get_tlogs ()`, did you mean `get_tlog_urls()`?

That could work if we change `get_tlog_urls(self) -> list[str]` to also return a `v1.Service`.

`get_tlog_service(self, api_major_version: int) -> v1.Service`, and maybe also change the name to `get_tlog_service`, since we're currently only supporting `ServiceSelector.ANY`

ramonpetgrave64:
@jku Yes, like we discussed offline, it perhaps would be nicer if the V2 changes would not edit the current `RekorClient` class. I'll have to see how viable the auto-generated v2 client `RekorBase` is. 

- https://github.com/sigstore/sigstore-python/pull/1387/files#diff-d7b419ac40c89309564411de463dfcde20b004790450a98f3077da9e1f60c459R267

If not, I also mentioned that subclassing could make more sense. Abstract the common methods of V1 and V2 into another base class. If that, I don't think removing V1 support in the future after the estimated 18 months would be difficult.

sigstore/sign.py
jku: * Yes, I mean get_tlog_urls()
* returning a list of urls is enough in the proposal since this code knows which major version it asks for from get_log_urls()
* I think returning a list is correct -- SigningConfig does not support other serviceselectors **yet**, but we can fix that later and then the return value is a naturally a list

ramonpetgrave64: Ok, and you prefer the caller of `get_tlog_urls(self, api_major_version: int) -> list[str]` prefer the known highest `api_major_version`?

sigstore/sign.py
ramonpetgrave64: Changed for just hashedrekord. dsse to come later.

sigstore/sign.py
ramonpetgrave64: I see we're meant to use the highest that the client supports

https://github.com/sigstore/rekor-tiles/pull/255/files#diff-eb568acf84d583e4d3734b07773e96912277776bad39c560392aa33ea2cf2210R373-R375

jku: right: Once all public good instances have v2 running we can just drop support for v1 altogether IMO but for now I see it as the signing code responsibility to ask for the urls/services for a specific version in order of preference: this means that  all of the complexity of Services (like service selectors, operators etc) can remain an implementation detail of `trust.py` -- the signing code just knows which versions it supports and asks for urls.

jku:
closing as housekeeping: I believe #1414 is closer to current SOTA

PR #1400, opened 2025-05-20
[DRAFT] Add RekorV2Client

ramonpetgrave64:

#### Summary

Adds a new `RekorV2Client` class, for use as a library, not yet within the CLI.

With the same pending TODOs in https://github.com/sigstore/sigstore-python/pull/1387

- We are embedding the auto-generated types until we can find a home for them

### Testing

- Unit tests that call live instances of RekorV2, only "alpha" for now
- Lints are not yet expected to pass, mainly due to type hints on the copy-pasted auto-generated types.

### TODOS

- [ ] add more client methods: `get_tile`, `get_entry_bundle`, `get_checkpoint`.
- [ ] enable testing against more instances: staging, prod, local
- [ ] add extra assert in test after https://github.com/sigstore/sigstore-python/pull/1370
- [ ] use these types in sigstore/protobuf-specs when ready (hopefully this week of May 19, 2025)
- [ ] remove workaround where the `...V_0_0_2` is changed to `...V002`
- [x] prefer sending the certificate, rather than the public key to Rekor

#### Release Note

* Added a `RekorV2Client` for posting new entries to a Rekor V2 instance.

#### Documentation

TODO


sigstore/_internal/rekor/client_v2.py
```diff
@@ -0,0 +1,187 @@
...
+        return cls(
+            DEFAULT_REKOR_URL,
+        )
+
+    @classmethod
+    def staging(cls) -> RekorV2Client:
+        """
+        Returns a `RekorV2Client` populated with the default Rekor staging instance.
+        """
+        return cls(STAGING_REKOR_URL)
```
jku, line 166: I think we do not need these (as long as #1363 merges): the rekor v1 client still needs them for the "fix-bundle" special case but that should not apply to this new code.

ramonpetgrave64: removed.

jku:
Note to self:
* I **think** nothing in either Client really needs the "rekor types" that the _build_* methods return: `create_entry()` uses them but only to generate the json payload
* so it looks like the `_build_*()` methods could just return the relevant json payload
* this would allow making the create_entry method signatures actually same and not  just pretend they are similar

jku:
todo for this pr:
* sort out the protobuf situation (https://github.com/sigstore/rekor-tiles/issues/338): preferably get the proto code generated in protobuf-specs, release that. Alternatively update and polish the generated code here -- this would be a temporary  thing
* testing (I'm looking at this next): 
  * I think the current tests for the `_build_*()` methods are not amazing: they are complicated and basically only repeat the content of the methods in the test and I think rely on an external service (OIDC) for no reason: the value is questionable... I think I will remove them for now. 
  * The create_entry() test seems usable -- we'll just have to get it actually running on CI. Using staging rekor seems most practical
  * the support for multiple environments in the tests is probably overkill -- unit tests should just verify the code works

jku:
closing for #1422 (the branch is the same, but now in origin so staging tests run on CI)

PR #1407, opened 2025-05-22
trust: Support operator field, support multiple services

jku:
* Change the SigningConfig API so that it returns actual clients (like RekorClient): this may seem unusual but makes sense because SigningConfig in practice must know which service/versions this client supports, this means the code is simplest if it directly returns correct clients (e.g. when rekor has two separate clients for v1 and v2). This allows keeping version, operator, selector mode, etc as SigningConfig implementation details: Otherwise some of those have to start leaking outside
* There is one exception to previous change: get_oidc_url() still returns string, not Issuer object. I could make this change as well to be consistent, it just requires a small refactor (because currently Issuer makes a http request on construction and that seems bad in general, but especially for testing)
* Support operator field: This is used to ensure we only return one service version per operator
* Support all serviceselector modes
* There is a tweak to to the production signing config placeholder in _store: this is required since the code now considers a "ANY" selector with no services listed an error (I think that's reasonable: "ALL" could mean 0 but "ANY" can't really)

Getting `_get_valid_services()` right (and simple) was surprisingly tricky, but I think it's there now -- thanks to ramon for feedback. I think the decision to keep this code contained in trust.py is correct: it's not trivial and sign.py is complicated enough already.

@ramonpetgrave64 I think this fits your plans well:
* When rekor2 client is added, we just make sure `get_tlogs()` returns either v1 client or v2 depending on major_version
* signing code can just use the client regardless of version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,9 @@ All versions prior to 0.9.0 are untracked.
   * ClientTrustConfig now provides methods `production()`, `staging()`and `from_tuf()`
     to get access to current client configuration (trusted keys & certificates,
     URLs and their validity periods). [#1363](https://github.com/sigstore/sigstore-python/pull/1363)
+  * SigningConfig now has methods that return actual clients (like `RekorClient`) instead of
+    just URLs. The returned clients are also filtered according to SigningConfig contents.
+    [#1407](https://github.com/sigstore/sigstore-python/pull/1407)
 * `--trust-config` now requires a file with SigningConfig v0.2, and is able to fully
   configure the used Sigstore instance [#1358]/(https://github.com/sigstore/sigstore-python/pull/1358)
 * By default (when `--trust-config` is not used) the whole trust configuration now
--- a/sigstore/_internal/trust.py
+++ b/sigstore/_internal/trust.py
@@ -340,21 +340,31 @@ class SigningConfig:
         except ValueError:
             raise Error(f"unsupported signing config format: {self._inner.media_type}")
 
-        # currently not supporting other select modes
-        # TODO: Support other modes ensuring tsa_urls() and tlog_urls() work
-        if self._inner.rekor_tlog_config.selector != ServiceSelector.ANY:
-            raise Error(
-                f"unsupported tlog selector {self._inner.rekor_tlog_config.selector}"
-            )
-        if self._inner.tsa_config.selector != ServiceSelector.ANY:
-            raise Error(f"unsupported TSA selector {self._inner.tsa_config.selector}")
-
         # Create lists of service protos that are valid & supported by this client
+        # Limit the TSA and tlog lists using the service selector config
         self._tlogs = self._get_valid_services(
             self._inner.rekor_tlog_urls, REKOR_VERSIONS
         )
+        if not self._tlogs:
+            raise Error("No valid Rekor transparency log found in signing config")
+        if self._inner.rekor_tlog_config.selector == ServiceSelector.EXACT:
+            if len(self._tlogs) < self._inner.rekor_tlog_config.count:
+                raise Error(
+                    "Not enough Rekor transparency logs found in signing config"
+                )
+            self._tlogs = self._tlogs[: self._inner.rekor_tlog_config.count]
+        elif self._inner.rekor_tlog_config.selector == ServiceSelector.ANY:
+            self._tlogs = self._tlogs[:1]
+
         self._tsas = self._get_valid_services(self._inner.tsa_urls, TSA_VERSIONS)
+        if self._inner.tsa_config.selector == ServiceSelector.EXACT:
+            self._tsas = self._tsas[: self._inner.tsa_config.count]
+        elif self._inner.tsa_config.selector == ServiceSelector.ANY:
+            self._tsas = self._tsas[:1]
+
         self._fulcios = self._get_valid_services(self._inner.ca_urls, FULCIO_VERSIONS)
+        if not self._fulcios:
+            raise Error("No valid Fulcio CA found in signing config")
         self._oidcs = self._get_valid_services(self._inner.oidc_urls, OIDC_VERSIONS)
 
     @classmethod
@@ -397,9 +407,6 @@ class SigningConfig:
         """
         Returns the rekor transparency logs that client should sign with.
         """
-
-        if not self._tlogs:
-            raise Error("No valid Rekor transparency log found in signing config")
         return [RekorClient(tlog.url) for tlog in self._tlogs]
 
     def get_fulcio(self) -> FulcioClient:
@@ -407,8 +414,6 @@ class SigningConfig:
         Returns url for the fulcio instance that client should use to get a
         signing certificate from
         """
-        if not self._fulcios:
-            raise Error("No valid Fulcio CA found in signing config")
         return FulcioClient(self._fulcios[0].url)
 
     def get_oidc_url(self) -> str:
--- a/sigstore/_internal/trust.py
+++ b/sigstore/_internal/trust.py
@@ -342,25 +342,27 @@ class SigningConfig:
 
         # Create lists of service protos that are valid & supported by this client
         # Limit the TSA and tlog lists using the service selector config
-        self._tlogs = self._get_valid_services(
-            self._inner.rekor_tlog_urls, REKOR_VERSIONS
-        )
-        if not self._tlogs:
+        tlogs = self._get_valid_services(self._inner.rekor_tlog_urls, REKOR_VERSIONS)
+        if not tlogs:
             raise Error("No valid Rekor transparency log found in signing config")
         if self._inner.rekor_tlog_config.selector == ServiceSelector.EXACT:
-            if len(self._tlogs) < self._inner.rekor_tlog_config.count:
+            if len(tlogs) < self._inner.rekor_tlog_config.count:
                 raise Error(
                     "Not enough Rekor transparency logs found in signing config"
                 )
-            self._tlogs = self._tlogs[: self._inner.rekor_tlog_config.count]
+            self._tlogs = tlogs[: self._inner.rekor_tlog_config.count]
         elif self._inner.rekor_tlog_config.selector == ServiceSelector.ANY:
-            self._tlogs = self._tlogs[:1]
+            self._tlogs = tlogs[:1]
+        else:
+            self._tlogs = tlogs
 
-        self._tsas = self._get_valid_services(self._inner.tsa_urls, TSA_VERSIONS)
+        tsas = self._get_valid_services(self._inner.tsa_urls, TSA_VERSIONS)
         if self._inner.tsa_config.selector == ServiceSelector.EXACT:
-            self._tsas = self._tsas[: self._inner.tsa_config.count]
+            self._tsas = tsas[: self._inner.tsa_config.count]
         elif self._inner.tsa_config.selector == ServiceSelector.ANY:
-            self._tsas = self._tsas[:1]
+            self._tsas = tsas[:1]
+        else:
+            self._tsas = tsas
 
         self._fulcios = self._get_valid_services(self._inner.ca_urls, FULCIO_VERSIONS)
         if not self._fulcios:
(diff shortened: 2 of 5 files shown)


jku:
oops, apparently I did not test with production :)

sigstore/_internal/trust.py
```diff
@@ -355,55 +366,65 @@ def from_file(
...
+            logs_by_operator[service.operator].append(service)
+
+        # return a list of services but make sure we only return logs of one version per operator
+        result: list[Service] = []
+        for logs in logs_by_operator.values():
+            logs.sort(key=lambda s: s.major_api_version)
+            max_version = logs[-1].major_api_version
+
+            while logs and logs[-1].major_api_version == max_version:
+                result.append(logs.pop())
```
jku, line 392: oh, I forgot to check the selector here: will fix. 
* If ANY, we can break both loops

jku: I decided to support all of the modes since it was so close already

sigstore/_internal/trust.py
```diff
@@ -355,55 +376,60 @@ def from_file(
...
-            return service.url
-        return None
 
-    def get_tlog_urls(self) -> list[str]:
+            logs_by_operator[service.operator].append(service)
+
+        # return a list of services but make sure we only return logs of one version per operator
+        result: list[Service] = []
+        for logs in logs_by_operator.values():
+            logs.sort(key=lambda s: s.major_api_version)
```
ramonpetgrave64, line 398:
```suggestion
            logs.sort(key=lambda s: s.major_api_version, reverse=True)
```

sigstore/_internal/trust.py
```diff
@@ -323,28 +333,39 @@ def __init__(self, inner: _SigningConfig):
...
-        # TODO: Support other modes ensuring tsa_urls() and tlog_urls() work
-        if self._inner.rekor_tlog_config.selector != ServiceSelector.ANY:
-            raise Error(
-                f"unsupported tlog selector {self._inner.rekor_tlog_config.selector}"
-            )
-        if self._inner.tsa_config.selector != ServiceSelector.ANY:
-            raise Error(f"unsupported TSA selector {self._inner.tsa_config.selector}")
+        # Create lists of service protos that are valid & supported by this client
+        # Limit the TSA and tlog lists using the service selector config
+        self._tlogs = self._get_valid_services(
```
ramonpetgrave64, line 345: I think it would be safer to not yet set the instance variable. Wait until we decide on a on the set of Services.

sigstore/_internal/trust.py
```diff
@@ -323,28 +333,39 @@ def __init__(self, inner: _SigningConfig):
...
+                )
+            self._tlogs = self._tlogs[: self._inner.rekor_tlog_config.count]
+        elif self._inner.rekor_tlog_config.selector == ServiceSelector.ANY:
+            self._tlogs = self._tlogs[:1]
+
+        self._tsas = self._get_valid_services(self._inner.tsa_urls, TSA_VERSIONS)
+        if self._inner.tsa_config.selector == ServiceSelector.EXACT:
+            self._tsas = self._tsas[: self._inner.tsa_config.count]
+        elif self._inner.tsa_config.selector == ServiceSelector.ANY:
+            self._tsas = self._tsas[:1]
```
ramonpetgrave64, line 363: Consider abstracting this into a separate function, like

```python
def _eveluluate_selector(services, selector)
```

sigstore/sign.py
```diff
@@ -332,12 +332,10 @@ def from_trust_config(cls, trust_config: ClientTrustConfig) -> SigningContext:
         """
         signing_config = trust_config.signing_config
         return cls(
-            fulcio=FulcioClient(signing_config.get_fulcio_url()),
-            rekor=RekorClient(signing_config.get_tlog_urls()[0]),
+            fulcio=signing_config.get_fulcio(),
+            rekor=signing_config.get_tlogs()[0],
```
ramonpetgrave64, line 336: This is cool. Potentially we sign and entry with multiple different tlogs, and put them into a single bundle.

ramonpetgrave64 (changes requested):

sigstore/_internal/trust.py
jku: I'll need to add yet another else-branch in that case but I suppose it is more readable

sigstore/_internal/trust.py
jku: I can handle it in `_get_valid_services()` if I provide the selectorconfig maybe I'll try that

sigstore/_internal/trust.py
jku: this got resolved differently by following your later suggestion

jku:
As for tests to make sure the service selection really handles more complicated cases, I suggest I file a new issue for those and do them later: 
* new code is not worse than what we had (since it passess existing tests and works with prod and staging signingconfigs)
* this is basically a blocker for ramons PRs

jku:
Added the changelog entry I had forgot: it's not a great changelog entry but I think we'll need to rewrite the changelog after all the rekor v2 work anyway to make it useful: as long as we have all changes listed it should be doable.

sigstore/_internal/trust.py
```diff
@@ -355,55 +372,73 @@ def from_file(
...
+            max_version = logs[-1].major_api_version
+
+            while logs and logs[-1].major_api_version == max_version:
+                result.append(logs.pop())
+
+        # limit the list based on ServiceConfiguration
+        if not config or config.selector == ServiceSelector.ALL:
+            return result
+
+        count = config.count if config.selector == ServiceSelector.EXACT else 1
```
ramonpetgrave64, line 407: Add a comment saying this for when the selector is ANY or undefined.

ramonpetgrave64 (changes requested):

sigstore/_internal/trust.py
```diff
@@ -355,55 +372,73 @@ def from_file(
...
+            logs_by_operator[service.operator].append(service)
+
+        # build a list of services but make sure we only include logs of one version per operator
+        result: list[Service] = []
+        for logs in logs_by_operator.values():
+            logs.sort(key=lambda s: -s.major_api_version)
+            max_version = logs[-1].major_api_version
+
+            while logs and logs[-1].major_api_version == max_version:
+                result.append(logs.pop())
```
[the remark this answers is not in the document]
jku: yes, I thought this was the idea. Maybe you are right and that has no value...

sigstore/_internal/trust.py
```diff
@@ -355,55 +372,73 @@ def from_file(
...
-        return None
 
-    def get_tlog_urls(self) -> list[str]:
+            logs_by_operator[service.operator].append(service)
+
+        # build a list of services but make sure we only include logs of one version per operator
+        result: list[Service] = []
+        for logs in logs_by_operator.values():
+            logs.sort(key=lambda s: -s.major_api_version)
+            max_version = logs[-1].major_api_version
```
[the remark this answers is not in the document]
jku: oh, so your original review comment wasn't about a bug , you just wanted a different direction...  That was dumb of me to assume

Thanks, I'll think about if I revert or do what you suggest here.

sigstore/_internal/trust.py
jku: I suppose I should handle those separately since undefined should be an error ... I really don't get why it needs to be there...

sigstore/_internal/trust.py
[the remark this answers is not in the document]
jku: Yeah you are right, I read the spec incorrectly

jku:
ramon found multiple bugs: I 'll take that as indication that I need to add the tests in this PR and not later...

sigstore/_internal/trust.py
```diff
@@ -323,28 +334,34 @@ def __init__(self, inner: _SigningConfig):
...
+
+        self._fulcios = self._get_valid_services(
+            self._inner.ca_urls, FULCIO_VERSIONS, None
+        )
+        if not self._fulcios:
+            raise Error("No valid Fulcio CA found in signing config")
+
+        self._oidcs = self._get_valid_services(
+            self._inner.oidc_urls, OIDC_VERSIONS, None
+        )
```
ramonpetgrave64, line 364: Add a comment or docstring saying we only support one CA and one OIDC. 

But maybe we should require that the signing config explicitly be `ANY` for CA and OIDC, or just log a warning if the user supplied a config with `ALL` for CA and OIDC.

sigstore/sign.py
ramonpetgrave64, line 336: Add a docstring saying this method will use only one tlog and one fulcio.

ramonpetgrave64 (changes requested):

sigstore/_internal/trust.py
jku: IIRC the spec does not contain ServiceConfiguration for those two.

I think the only potential use for multiple services here would be redundancy (as in, if this URL is 404, try the other one) but even that seems a little theoretical

sigstore/sign.py
jku: filed #1412

test/unit/internal/test_trust.py
```diff
@@ -56,12 +77,117 @@ def test_good(self, asset):
...
+                [_service_v1_op1, _service_v1_op2],
+            ),
+        ],
+    )
+    def test_get_valid_services(self, services, versions, config, expected_result):
+        result = SigningConfig._get_valid_services(services, versions, config)
+
+        assert len(result) == len(expected_result)
+        for s1, s2 in zip(result, expected_result):
+            assert s1.url == s2.url
```
ramonpetgrave64, line 166: List equality should work here, or use `set(result) == set(expected_result)`, if the sorting order is unpredictable.

```suggestion
        assert result == expected_result
```

test/unit/internal/test_trust.py
```diff
@@ -56,12 +77,117 @@ def test_good(self, asset):
...
+    @pytest.mark.parametrize(
+        "services, versions, config, expected_result",
+        [
+            (
+                [_service_v1_op1],
+                [1],
+                ServiceConfiguration(ServiceSelector.ALL),
+                [_service_v1_op1],
+            ),
+            (  # multiple services, same operator: expect 1 service in result
```
ramonpetgrave64, line 105: Non-blocker: consider settting the id for the test cases, like
```python
pytest.param(
    [_service_v1_op1, _service2_v1_op1],
    [1],
    ServiceConfiguration(ServiceSelector.ALL),
    [_service2_v1_op1], 
    id="multiple services, same operator: expect 1 service in result"
)
```

ramonpetgrave64 (approved):

test/unit/internal/test_trust.py
jku: hmm, I was sure some of the protobuf types had issues with `__eq__` but you seem to be right

sigstore/_internal/trust.py
```diff
@@ -356,54 +373,73 @@ def from_file(
...
-        signing certificate from
+        Returns a Fulcio client to get a signing certificate from
         """
-        url = self._get_valid_service_url(self._inner.ca_urls)
-        if not url:
-            raise Error("No valid Fulcio CA found in signing config")
-        return url
+        return FulcioClient(self._fulcios[0].url)
 
     def get_oidc_url(self) -> str:
```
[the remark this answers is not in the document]
woodruffw: This design looks reasonable to me; doing it in a followup sounds good!

woodruffw (approved):
LGTM, thanks @jku!

(One nitpick about the CHANGELOG, but nonblocking.)

CHANGELOG.md
```diff
@@ -50,6 +50,9 @@ All versions prior to 0.9.0 are untracked.
   * ClientTrustConfig now provides methods `production()`, `staging()`and `from_tuf()`
     to get access to current client configuration (trusted keys & certificates,
     URLs and their validity periods). [#1363](https://github.com/sigstore/sigstore-python/pull/1363)
+  * SigningConfig now has methods that return actual clients (like `RekorClient`) instead of
+    just URLs. The returned clients are also filtered according to SigningConfig contents.
+    [#1407](https://github.com/sigstore/sigstore-python/pull/1407)
```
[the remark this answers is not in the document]
jku: yeah, I was thinking of this as well while adding the entry. In a way SigningConfig semi-public since the signing_config property in ClientTrustConfig is currently public -- and ClientTrustConfig is in practice exposed through the sign and verify modules...

I think the intent of the API now is that ClientTrustConfig is public (so users can e.g. call `from_production()`) but the internals of ClientTrustConfig are not... so in that way this is not an API change but it's debatable.

I might leave it as is for now: We need to rewrite the changelog anyway for next release (combine related changes, try to make a coherent log).

PR #1414, opened 2025-05-23
sign and verify with Rekorv2

ramonpetgrave64:

#### Summary

This PR, in favor of #1387

Pending #1400  

Enables signing and verifying with Rekorv2 via a trust-config.

### Testing

Unit tests should pass, but are currently using somewhat hacky method of paramaterizing fixtures to use either staging or the embedded tust_config.

#### TODO:

 - [ ] a less hacky fixture setup
 - [ ] tests for cli/attest/py
 - [ ] documentation
 - [ ] tests for cli/verify, with an embedded bundle


sigstore/verify/verifier.py
```diff
@@ -495,15 +553,58 @@ def verify_artifact(
...
+                            algorithm=bundle._inner.message_signature.message_digest.algorithm,
+                            digest=bundle._inner.message_signature.message_digest.digest,
+                        ),
+                        signature=v2.Signature(
+                            content=bundle._inner.message_signature.signature,
+                            verifier=v2.Verifier(
+                                public_key=v2.PublicKey(
+                                    raw_bytes=bundle.signing_certificate.public_key().public_bytes(
+                                        encoding=serialization.Encoding.DER,
+                                        format=serialization.PublicFormat.SubjectPublicKeyInfo,
```
jku, line 588: this does not look right: if the bundle contains a signing certificate, we should expect the signature to contain that (and not just the public key of the certificate), right?

jku:
I think I have an updated version in #1432 --  @ramonpetgrave64 there might still be something we want here, but I'm going to close this one for clarity for now.

sigstore/verify/verifier.py
jku: I've used the certificate in my PR, I think it looks ok

PR #1422, opened 2025-06-04
Add Rekor v2 client

jku:
This is a continuation of Ramons #1400 , I'm moving it here in hopes of getting staging tests running.

The PR adds a new RekorV2Client for adding entries to a rekor-tiles log.
* The plan is that a followup PR starts using this client in the CLI when signingconfig contains a rekor v2 instance
* The existing RekorClient is slightly modified so that there is a common abstraction that the signing code can use
   * the abstraction contains `create_entry()` and methods for constructing  the requests for dsse/hashedrekord
   * name of the ABC is currently `RekorLogSubmitter` -- we could name it RekorClient but obviously would then need to rename the original V1 client...
* Note that the client does not include functionality to lookup entries from the log (since a sigstore client does not strictly speaking need to do that). The feature could still be added
* testing is very light but it does show that the client can add entries to a rekor-tiles log (the staging instance)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -130,7 +130,7 @@ jobs:
         # Confusingly, this action also supports updating releases, not
         # just creating them. This is what we want here, since we've manually
         # created the release that triggered the action.
-        uses: softprops/action-gh-release@da05d552573ad5aba039eaac05058a918a7bf631 # v2.2.2
+        uses: softprops/action-gh-release@72f2c25fcb47643c292f7107632f7a47c1df5cd8 # v2.3.2
         with:
           # smoketest-artifacts/ contains the signatures and certificates.
           files: |
--- a/.github/workflows/scorecards-analysis.yml
+++ b/.github/workflows/scorecards-analysis.yml
@@ -52,6 +52,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@fca7ace96b7d713c7035871441bd52efbe39e27e # v3.28.19
+        uses: github/codeql-action/upload-sarif@ce28f5bb42b7a9f2c824e633a3f6ee835bab6858 # v3.29.0
         with:
           sarif_file: results.sarif


sigstore/_internal/rekor/client_v2.py
```diff
@@ -0,0 +1,173 @@
...
+                digest=hashed_input.digest,
+                signature=v2.Signature(
+                    content=signature,
+                    verifier=v2.Verifier(
+                        x509_certificate=common_v1.X509Certificate(
+                            raw_bytes=certificate.public_bytes(
+                                encoding=serialization.Encoding.DER
+                            )
+                        ),
+                        key_details=DEFAULT_KEY_DETAILS,  # type: ignore[arg-type]
```
jku, line 113: why are we using a hard coded key type here?

sigstore/_internal/rekor/client_v2.py
```diff
@@ -0,0 +1,195 @@
...
+    A generic error in the Rekor client.
+    """
+
+    def __init__(self, http_error: requests.HTTPError):
+        """
+        Create a new `RekorClientError` from the given `requests.HTTPError`.
+        """
+        if http_error.response is not None:
+            try:
+                error = rekor_types.Error.model_validate_json(http_error.response.text)
```
jku, line 188: this doesn't make a lot of sense and is the only use of rekor_types here

test/unit/conftest.py
```diff
@@ -216,7 +216,7 @@ def ctx_cls():
     return ctx_cls, IdentityToken(token)
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
```
jku, line 219: I think this is leftover from the earlier test setup... I will have a look

sigstore/_internal/rekor/client_v2.py
```diff
@@ -0,0 +1,166 @@
...
+
+        integrated_entry = resp.json()
+        _logger.debug(f"integrated: {integrated_entry}")
+        return LogEntry._from_dict_rekor(integrated_entry)
+
+    @staticmethod
+    def _get_key_details(certificate: Certificate) -> common_v1.PublicKeyDetails:
+        """Determine PublicKeyDetails from a certificate
+
+        We know that sign.Signer only uses secp256r1 so do not support anything else"""
```
woodruffw, line 95: For consistency with the other docstrings:

```suggestion
        """
        Determine PublicKeyDetails from a certificate

        We know that sign.Signer only uses secp256r1 so do not support anything else
        """
```

test/unit/internal/rekor/test_client_v2.py
```diff
@@ -0,0 +1,80 @@
...
+        .predicate(
+            {
+                "Data": "",
+                "Timestamp": "2023-12-07T00:37:58Z",
+            }
+        )
+    ).build()
+
+    with sign_ctx.signer(identity) as signer:
+        bundle = signer.sign_dsse(stmt)
```
ramonpetgrave64, line 59: One thing I wanted to ensure that the certificate, not only the public key, was used in the request. It's no problem to not include that here, though.

ramonpetgrave64 (changes requested):

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ All versions prior to 0.9.0 are untracked.
   [#1402](https://github.com/sigstore/sigstore-python/pull/1402)
 
 
+* Added a `RekorV2Client` for posting new entries to a Rekor V2 instance.
+  [#1400](https://github.com/sigstore/sigstore-python/pull/1400)
```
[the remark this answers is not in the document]
jku: I'll leave it in for now but agree: One item mentioning rekor v2 support (and how it is triggered) should be enough in the end

sigstore/_internal/rekor/__init__.py
```diff
@@ -16,18 +16,85 @@
...
 from sigstore._utils import base64_encode_pem_cert
+from sigstore.dsse import Envelope
 from sigstore.hashes import Hashed
+from sigstore.models import LogEntry
 
 __all__ = [
     "_hashedrekord_from_parts",
 ]
 
+EntryRequest = NewType("EntryRequest", dict[str, Any])
```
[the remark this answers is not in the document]
jku: :+1: thanks for the patches

test/unit/internal/rekor/test_client_v2.py
jku: Yeah tests that really dissect both the request and response would be cool... For the request we should be able to do that without ever contacting rekor or any other service: a test that provides existing assets to the `_build_*()` methods and ensures the request body looks like we expect should do it -- but I'm ok leaving that for followup.

ramonpetgrave64 (approved):

ramonpetgrave64:
Looks like some tests failed after merging. Can you run them again to make sure?

PR #1424, opened 2025-06-05
trust: Fail less hard when unsupported keys are seen

jku:
Currently verification fails immediately if trusted root contains any unsupported keys. I think it makes more sense to warn and continue as it is possible these keys are not required for verification.

Unfortunately I missed this case when I tested the multiple log support in #1350  :(

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,10 @@ All versions prior to 0.9.0 are untracked.
   still required.
   [#1381](https://github.com/sigstore/sigstore-python/pull/1381)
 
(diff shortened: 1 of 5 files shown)


jku:
Marking this a draft while I do another test with the (future) staging  trusted root

woodruffw (approved):
LGTM, thanks @jku!

woodruffw:
> I'm thinking of backporting this fix to a 3.6.x branch:

Sounds good to me -- I see you opened the backport PR, give me a ping on it when it's ready for review 🙂

PR #1425, opened 2025-06-05
Backport 1424, prepare 3.6.3 release

jku:
This is a backport of #1424 into 3.6.x series:
* This backport should make root-signing-staging update (with a ed25519 key) a little easier to manage
* Makes it more likely that all sigstore-python users have the fix once the same happens in production

Current main contains a lot of work (but not yet rekor v2 client integration) so I wouldn't like to release all of that now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -616,7 +616,8 @@ This is a corrective release for [2.1.1].
 
 
 <!--Release URLs -->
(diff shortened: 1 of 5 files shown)


jku:
#1424 was merged: this is a backport of that.

The only differences are in tests (that have changed in main after 3.6.2)

woodruffw (approved):
Thanks @jku!

PR #1432, opened 2025-06-09
Sign with rekorv2

jku:
Start signing and verifying with rekor v2 (when signingconfig / trustedroot instruct to do so).

**Status:** 

* Waiting for #1422 to merge
* This is a continuation of #1414 (although does not use any direct commits from there). The verifier.py changes are mostly Ramons

**Contents:**

* SigningConfig now returns RekorV2Clients when appropriate
* tests: 
  * SigningConfig test are amended to test for different rekor clients
  * A simple rekorv2 signing test is added
  * a test asset is added for verifying rekorv2 signatures, this is used in an existing test
* Verifier: verify_dsse() and verify_artifact() now handle v002 entry types
* A trustconfig for signing with staging rekor v2 is added to assets: It's not used in tests but is handy for manually signing with `--trust-config test/assets/trust_config/staging-but-sign-with-rekor-v2.json`

**TODO:**

* Verifier contains a bit too much manual protobuf building: should clean that up
* Good ideas for better testing are welcome


sigstore/verify/verifier.py
```diff
@@ -491,15 +552,57 @@ def verify_artifact(
...
+                                    )
+                                ),
+                                key_details=self._get_key_details(
+                                    bundle.signing_certificate
+                                ),
+                            ),
+                        ),
+                    )
+                ),
             )
```
jku, line 588: this is the code I'd rather refactor somewhere else -- v1 RekorClient has `_hashedrekord_from_parts()` for the same purpose

sigstore/verify/verifier.py
```diff
@@ -418,34 +436,13 @@ def verify_dsse(
...
-                signature=base64.b64encode(signature.sig).decode(),
-                verifier=base64_encode_pem_cert(bundle.signing_certificate),
-            )
-            for signature in envelope._inner.signatures
-        ]
-        if signatures != entry_body.spec.root.signatures:
-            raise VerificationError("log entry signatures do not match bundle")
+            validate_dsse_v002_entry_body(bundle)
+        else:
+            validate_dsse_v001_entry_body(bundle)
```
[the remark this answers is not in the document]
jku: I _think_ this is now safe to do (as in does not break verifying older entries) but let's make sure...

jku:
* rebased on the rekor v2 client branch
* refactored verifier a bit more
* Added a better failure for unknown  entry types/versions: I *think* this is safe to do with regards to old bundles (the kind_version gets backfilled from the entry content itself during deserialization in those cases)

sigstore/verify/verifier.py
```diff
@@ -635,3 +607,29 @@ def validate_hashedrekord_v002_entry_body(bundle: Bundle) -> None:
...
+    if isinstance(public_key, EllipticCurvePublicKey):
+        if public_key.curve.name == "secp256r1":
+            key_details = cast(
+                v1.PublicKeyDetails,
+                v1.PublicKeyDetails.PKIX_ECDSA_P256_SHA_256,
+            )
+        else:
+            raise ValueError(f"Unsupported EC curve: {public_key.curve.name}")
+
+    # TODO support other keys
```
jku: Do we have blessed list of key types we should expect the signing certificates to contain?

jku: https://github.com/sigstore/architecture-docs/blob/main/algorithm-registry.md is close. I will have a closer look.

jku:
last commit: Made sure we support the same signature algorithms in verification that we support with rekor v1 entries (we should support a bit more but that's another PR)

ramonpetgrave64:
> Good ideas for better testing are welcome 

I think we can get a lot more coverage without writing new tests in #1441

sigstore/verify/verifier.py
```diff
@@ -490,16 +477,175 @@ def verify_artifact(
...
+                    digest=bundle._inner.message_signature.message_digest.digest,
+                ),
+                signature=v2.Signature(
+                    content=bundle._inner.message_signature.signature,
+                    verifier=_v2_verifier_from_certificate(bundle.signing_certificate),
+                ),
             )
+        ),
+    )
+    v2_actual_body = v2.Entry().from_json(base64.b64decode(entry.body))
```
haydentherapper, line 615: Is this canonicalized JSON? Otherwise, the comparison might fail with different JSON ordering.

sigstore/verify/verifier.py
```diff
@@ -490,16 +477,175 @@ def verify_artifact(
...
+    This method decides which signature algorithms are supported for verification
+    (in a rekor v2 entry), see
+    https://github.com/sigstore/architecture-docs/blob/main/algorithm-registry.md.
+    Note that actual signature verification happens in verify_artifact() and
+    verify_dsse(): New keytypes need to be added here and in those methods.
+    """
+    public_key = certificate.public_key()
+
+    if isinstance(public_key, ec.EllipticCurvePublicKey):
+        if isinstance(public_key.curve, ec.SECP256R1):
```
haydentherapper, line 635: To confirm, what set of algorithms was supported previously?

sigstore/verify/verifier.py
ramonpetgrave64: It's the inner canonicalized body returned from Rekor. Here, we parse the values it into our Entry object, so it wouldn't matter if it's canonicalized or not.

sigstore/verify/verifier.py
ramonpetgrave64: It's that same one, also hardcoded in a few other places in sigstore-python.

sigstore/verify/verifier.py
haydentherapper: Thanks, was thinking this was _to_ json rather than _from_. LGTM.

sigstore/verify/verifier.py
jku: #1439 is the follow up

haydentherapper:
@di @woodruffw can you take a pass over this? I’ve reviewed it, it looks functionally good, though I can’t speak to how pythonic it is.

Issue #1413, opened 2025-05-23
Add tests for local containers in CI


ramonpetgrave64:
**Description**

The new Action at https://github.com/sigstore/scaffolding/tree/main/actions/setup-sigstore-env can setup containers within a Workflow run, and it can generate a trusted root ([currently](https://github.com/sigstore/scaffolding/issues/1555) not signing config v0.02) and signing config to make it easy to use from the CLI.

It tracks mainline of the services Fulcio, Rekor, TSA, the new rekor-tiles, and provides a fakeoidc provider we might find useful.

Since it tracks mainline, we might run tests with this Action on a daily timer, just to make sure new changes will be compatible.

jku:
> Since it tracks mainline, we might run tests with this Action on a daily timer, just to make sure new changes will be compatible.

I'll point out that if the test runs in sigstore-python project it's not about "making sure changes are compatible" -- if you wanted to do that, you  should have run the test in the infra projects, _before merging the code_.

The issue with testing here is that it pushes the responsibility to react to failures to sigstore-python: this project does not have a huge amount of developers willing to monitor tests like this... Maybe this does makes sense, but this is the compromise that would be made. We'll want to carefully evaluate what we want to test against -- some possible examples:
1. staging and prod only
1. staging, prod and a local test setup running latest infra releases
1. staging, prod and a local test setup running every infra commit
  
My hunch is that  the last option seems unlikely, but the second one might be interesting?

ramonpetgrave64:
I was just trying this today and it looks like we need some more work on both the set-sigstore-env Action side and in sigstore-python

* The fakeoidc provider might need to contain more fields in it's .well-known-configuration response:
  * `authorization_endpoint` and `token_endpoint`, as expected by sigstore-python
    * https://github.com/sigstore/sigstore-python/blob/d05a629ebb1d2ed861600e99ef56c915cecda4fe/sigstore/oidc.py#L48-L57
  * https://github.com/sigstore/scaffolding/blob/c1697866accfa0fbe21caf7e16ab85f236695428/actions/setup-sigstore-env/fakeoidc/main.go#L37-L40

* Not 100% sure about this one: For compatibility with this local provider, I think we'll also need to try to get the token **before** trying to follow the `authorization_endpoint`.

My current workaround: invoke with --identity-token `curl <host>/token`. We can't just remove `oidcUrls` from the trust config because then the local fulcio won't respect the ambient oidc provider.

PR #1441, opened 2025-06-12
Test with setup-sigstore-env Action

ramonpetgrave64:

Addresses #1413

#### Summary

Runs tests against local rekor and fulcio containers using the sigstore/scaffolding setup-sigstore-env Action.

- Adds a new workflow `.github/workflows/test-with-setup-sigstore-env.yml` that runs scaffolding/setup-sigstore-env for local sigstore service containers.
- Renames the `staging` fixture `preprod`
- parameterizes the `preoprod` fixture to also support a new "local" option that uses the local containers.
  - if the conditions in `.github/workflows/test-with-setup-sigstore-env.yml` are not met, then test insstances that would use use the local containers are skipped.
  -  see https://docs.pytest.org/en/stable/how-to/fixtures.html#fixture-parametrize

## TODO:

- [x] Add the option to more tests
- [x] Consider adding this to the `staging` fixture, which may need renaming to perhaps `"preprod"`. We would paramaterize the fixture, so that every test function that uses the fixture would automatically run twice: once for real staging rekorv1 and again for alpha rekor v2.
- [x] make verification work for the local containers. It seems like an issue verifying the local fulcio certificate chain.
- [x] fix problem where lots of tests are skipped within CI.
  - Tests are skipped in this PR because my PR is from a fork, and so will not have permissions for `id-token: write`. I think a maintainer needs to close and reopen this PR for it work work, or open another PR from my branch.
    - `SKIPPED [33] test/conftest.py:104: skipping test that requires an ambient OIDC credential`
  - They are actually working on another PR in my fork repo.
    - https://github.com/sigstore/sigstore-python/pull/1441
    - https://github.com/ramonpetgrave64/sigstore-python/actions/runs/15643987379/job/44077716123#step:6:25

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
           unshare --map-root-user --net make test T="test/unit" TEST_ARGS="--skip-online -vv --showlocals"
 
       - name: test
-        run: make test TEST_ARGS="-vvv --showlocals"
+        run: make test TEST_ARGS="-vv --showlocals"
 
       # TODO: Refactor this or remove it entirely once there's
       # a suitable staging TSA instance.


jku:
closing for now as no one is actively pushing this:
* I really like the sigstore-in-a-box approach but it needs to be fast enough, both locally and in CI
* I don't think sigstore-python is interested in just testing against main versions of infra components by default -- that's probably too much to ask

Issue #1439, opened 2025-06-11
Support more signing cert signing algorithms in verification


jku:
sigstore-python chooses to use a specific signing algorithm for signing with the signing certificate: this seems like a fine decision.

Even for verifying the signing certificate signature, sigstore clients are not required to support more than one key type... but it would make sense to support the low hanging fruit in https://github.com/sigstore/architecture-docs/blob/main/algorithm-registry.md (IOW the algorithms supported by cryptography but not yet by sigstore.verify: RSA, EdDSA).

This requires some changes in verifier.py:
* `verify_artifact()`
* `verify_dsse()`
* after #1432 also `_v2_verifier_from_certificate()`

PR #1456, opened 2025-06-20
Add function for determining `key_details`

ramonpetgrave64:
#### Summary

Partially addresses https://github.com/sigstore/sigstore-python/issues/1439

Inspired by https://github.com/sigstore/sigstore-python/pull/1432/files#diff-0287921ae266ddac61ace4f5d2ee5fa8e74495f6e3a7c231d3715a31679690b1R622-R651

Adds a function for determining the `key_details` of a certificate.

Note that this does not add direct support for signing with the various key types, but in the future will aid with sending requests to Rekor and verifying Entry body contents.

#### Release Note

* Added a function for determining the `key_details` of a certificate`.


sigstore/_internal/key_details.py
```diff
@@ -0,0 +1,74 @@
...
+            key_details = v1.PublicKeyDetails.PKIX_ECDSA_P256_SHA_256
+        elif isinstance(public_key.curve, ec.SECP384R1):
+            key_details = v1.PublicKeyDetails.PKIX_ECDSA_P384_SHA_384
+        elif isinstance(public_key.curve, ec.SECP521R1):
+            key_details = v1.PublicKeyDetails.PKIX_ECDSA_P521_SHA_512
+        else:
+            raise ValueError(f"Unsupported EC curve: {public_key.curve.name}")
+    elif isinstance(public_key, rsa.RSAPublicKey):
+        if public_key.key_size == 2048:
+            raise ValueError("Unsupported RSA key size: 2048")
```
jku, line 47: this is already handled by the else branch

jku (commented):
The actual method looks pretty good (although maybe it could live in `sigstore/_utils.py` like other random methods like this).

> Addresses https://github.com/sigstore/sigstore-python/issues/1439

This is not true since this only modifies the signing code (I assume the plan was to extend this to support verification as well later on) .

* I think we can re-use the code for signing and verification like it looks like you planned but it's worth noting that 
   1. The signing code does not strictly speaking *need* this support (since there keytype is hard coded and that seems fine)
   1. this re-use saves 7 lines of code
   1. the requirements for signing & verification might end up different: *maybe* we want to support the "not recommended" algorithms for verification but not for signing
* let's at least wait until #1414 so we can plug this into the verification code where it will be actually useful

sigstore/_internal/key_details.py
ramonpetgrave64: fixed.

Issue #1457, opened 2025-06-23
Circular ImportError: sigstore.models and sigstore._internal.rekor dependency loop


SequeI:
When attempting to install and use sigstore-python directly from the Git repository (e.g., main branch or a recent commit), I'm encountering an ImportError due to a circular dependency between sigstore.models and sigstore._internal.rekor.

This issue prevents the successful initialization of the sigstore library, manifesting as soon as sigstore.models is imported.

```
Traceback (most recent call last):
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 148, in _get_module_details
  File "<frozen runpy>", line 112, in _get_module_details
  File "/model-transparency/src/model_signing/__init__.py", line 124, in <module>
    from model_signing import signing
  File "/model-transparency/src/model_signing/signing.py", line 54, in <module>
    from model_signing._signing import sign_sigstore as sigstore
  File "/model-transparency/src/model_signing/_signing/sign_sigstore.py", line 23, in <module>
    from sigstore import models as sigstore_models
  File "/.local/share/hatch/env/virtual/model-signing/-_BIuW3L/model-signing/lib/python3.13/site-packages/sigstore/models.py", line 65, in <module>
    from sigstore._internal.rekor.checkpoint import verify_checkpoint
  File "/.local/share/hatch/env/virtual/model-signing/-_BIuW3L/model-signing/lib/python3.13/site-packages/sigstore/_internal/rekor/__init__.py", line 32, in <module>
    from sigstore.models import LogEntry
ImportError: cannot import name 'LogEntry' from partially initialized module 'sigstore.models' (most likely due to a circular import) (/.local/share/hatch/env/virtual/model-signing/-_BIuW3L/model-signing/lib/python3.13/site-packages/sigstore/models.py)
```

[this](https://github.com/sigstore/sigstore-python/blob/aa695c4d95e4d1af7c526e6fbd92cdcab144112f/sigstore/_internal/rekor/checkpoint.py#L34) seems to be the main issue currently. Found while preparing `model_transparency` for the new release of this python client.

jku:
Agreed, looks like a bug that appeared when RekorV2Client was added

PR #1458, opened 2025-06-23
fix: resolve circular import of models.LogEntry

SequeI:
This PR addresses an ImportError caused by a circular dependency between sigstore.models and the sigstore._internal.rekor package, specifically in its __init__.py file. This issue prevented the successful initialization of the sigstore library at runtime.

The problem occurred because sigstore.models (during its loading) was importing verify_checkpoint from sigstore._internal.rekor.checkpoint, while sigstore._internal.rekor/__init__.py simultaneously attempted to import LogEntry from the still-loading sigstore.models.

Resolves sigstore/sigstore-python#1457

### DIFF
--- a/sigstore/_internal/rekor/__init__.py
+++ b/sigstore/_internal/rekor/__init__.py
@@ -19,8 +19,8 @@ APIs for interacting with Rekor.
 from __future__ import annotations
 
 import base64
+import typing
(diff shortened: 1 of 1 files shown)


jku (approved):
LGTM, thanks!

Issue #1049, opened 2024-06-20
Refactor: remove `betterproto` dep


woodruffw:
Our dependency on `betterproto` is a source of a few issues:

1. We currently depend on the `2.x` series, which is only in beta (currently `2.0.0b6`). This is both non-ideal *and* is incompatible with some Python installers, most notably `uv`: https://github.com/astral-sh/uv/blob/main/PIP_COMPATIBILITY.md#pre-release-compatibility
2. `betterproto`'s handling of the `Struct` WKT is partially broken. I've tried to fix it, but round-tripping a JSON serialized message through a `Struct` is still impossible, which causes all kinds of problems for the (mostly in-toto) definitions that use `Struct`
3. Finally, the models and their APIs themselves aren't particularly Pythonic -- this is more of an issue with protobuf itself than `betterproto`, but results in development friction and impedance mismatches elsewhere that we have to paper over

As such, we should probably remove our dependency on `betterproto` in the medium term. There are a few blockers to doing this:

1. The `Bundle`, etc. models all come from `protobuf-specs`, which is currently built on top of `betterproto` for the Python bindings.
2. There are (probably) a few places we leak references to `betterproto`-generated models in Sigstore's public APIs

(1) is probably addressable either by re-modeling the relevant parts of `protobuf-specs` in pydantic, or by regenerating the Python bindings on top of the JSON Schema for `protobuf-specs`. (2) may require a major reversion.

CC @darkamaul since he noticed this 🙂

woodruffw:
Not immediately useful, but saving this here so I don't forget it: https://github.com/criccomini/proto-schema-parser

mgorny:
I wanted to report that betterproto 2.0.0b7 was released in August (and that it fails tests), but given this issue, is there a point?

woodruffw:
That is indeed good to know! We're still planning on dropping our dependency on it, but we should test against the latest in the mean time. I'll start a PR for that.

mgorny:
Thanks! It would really be appreciated, given that b6 is partially broken with modern pydantic (though I don't think it affects sigstore, but its reproduced with its own test suite).

woodruffw:
https://github.com/sigstore/protobuf-specs/pull/404 has the bump, thanks again! I'll get a version of that package cut and then bump here once it's ready.

jku:
This should now be done in main since #1470 replaces sigstore-protobuf-specs with sigstore-models (and the latter does not depend on betterproto)

PR #1470, opened 2025-07-18
refactor: begin to remove sigstore_protobuf_specs

woodruffw:
Towards #1049.

~~This is very WIP and won't work in CI yet since I'm using a local editable install of the new `sigstore_models` package while I iterate on it.~~

NB: I've also temporarily disabled interrogate because it has some kind of issue with a `cairo` dep.


woodruffw:
I've published https://pypi.org/project/sigstore-models/ to accompany this.

woodruffw:
Thanks @jku!

I have an internal need for this, but there's no huge rush from me -- I can always pin to the `git` reference temporarily 🙂 

(With that being said, I'd love it if we could land this with v4, if you think that's possible -- I think it's be nice to have a clean major break.)

jku:
> These quirks should have no public API implications, but they require some more explicit internal round-trip handling than the sigstore_protobuf_specs APIs did. In particular, we now need to explicitly base64 encode bytes when passing them into the models, and same string-encoding uint64s.

This is the part I haven't yet wrapped my head around. Does pydantic still get some actual value from the serializers that are defined in sigstore-models?

woodruffw:
> This is the part I haven't yet wrapped my head around. Does pydantic still get some actual value from the serializers that are defined in sigstore-models?

Nope, it's more that Pydantic (which we were already using indirectly, just with less direct control) doesn't have a super clean split between "wire type" and "real type" (for lack of a better term).

For example, here's a made-up Protobuf definition:

```proto
message Hello {
    int64 abc = 1;
    bytes def = 2;
}
```

when serialized as ProtoJSON, that turns into a "wire" representation like:

```json
{ "abc": "1234", "def": "aGVsbG8=" }
```

i.e. `int64` gets wrapped as a string for range safety, and `bytes` gets wrapped as a base64 string because JSON has no raw bytes.

To then pull those out of JSON and into useful Python types, we need Pydantic type adapters, e.g.:

https://github.com/astral-sh/sigstore-models/blob/d0dad407dfe19a35aa24cc1c54054dc9d90ecd0e/src/sigstore_models/_core.py#L41C1-L85C2

...which we'd then use like this:

```python
class Hello(BaseModel):
    abc: ProtoU64
    def: ProtoBytes
```

This works really well once the object is actually loaded (since the revealed types for `abc` and `def` become `int` and `bytes` respectively), but it also means that initializing a `Hello` object needs to go through those wrapper types instead of directly though the revealed types. In other words, this *doesn't* work:

```python
Hello(abc=123, def=b"hello")
```

instead, we have to do:

```python
Hello(abc=str(123), def=base64.b64encode(b"hello"))
```

...so we end up with explicit `base64` littered in a few more places than the current models 😞 

There might be a way around this, but it's not super clear from the Pydantic docs. I can look into it some more.

woodruffw:
Hm, looks like Rekor v2 in staging is 503ing:

```
        if http_error_msg:
>           raise HTTPError(http_error_msg, response=self)
E           requests.exceptions.HTTPError: 503 Server Error: Service Unavailable for url: https://log2025-alpha1.rekor.sigstage.dev/api/v2/log/entries
```

jku (dismissed):
yeah, lgtm.

The rekor 503 did not reproduce (I did realize the rekor-tiles http logging does not seem to be enabled while looking at this, I'll follow up on that elsewhere)

jku:
DCO check was still complaining but I will count this as a good faith effort to sign-off-by :)

jas4711:
Hi!  Just a quick side-note from a Debian packager: python-sigstore-protobuf-specs is packaged in Debian but sigstore-models appears to be problematic to package because it uses the UV build system which is rust-based with a lot of dependencies.  People have worked for a year to add it to Debian and it is still not finished.  So making sigstore-models a required dependency may cause some friction for us...  so I would be happy if at least during some time window both would be supported.  I have no idea if this is feasible from a code point of view, just trying to share my view as a Debian packager.

PR #1471, opened 2025-07-22
cli: Add --rekor-version to sign arguments

jku:
This adds a CLI option that should not be needed... but it could be handy if
* SigningConfig already contains rekor v2 (and still contains a  v1 instance)
* user for some reason does not want rekor v2 entries in the bundle

This option only does anything if there are multiple Rekor versions listed in SigningConfig.

The test is changed since the "ANY" selector is now considered to not be
an error if there are 0 services:
* This is not a problem since for both TSAs and tlogs we have a check
  that there is at least one service
* This improves the error message when --rekor-version is used with
  a version that is not found in signingconfig


sigstore/_cli.py
```diff
@@ -288,6 +288,11 @@ def _parser() -> argparse.ArgumentParser:
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
         parents=[parent_parser],
     )
+    attest.add_argument(
+        "--rekor-version",
+        type=int,
+        help="Force the rekor transparency log version (in case there are multiple in signing configuration)",
+    )
```
di, line 295: I think it would be more clear & user-friendly if the default was set here (`1` for < `4.0.0` release and `2` for >`4.0.0` release) rather than selecting the default elsewhere. (It would also indicate to the user that these should be integers)

jku: the issue (that I should have document better here) is that 
* if user sets `--rekor-version X` I expect they *really* want that version -- it's not just a suggestion. This means that if version X rekor instance is not found in signingconfig this should lead to an error (and it does in this PR)
* but otherwise the default is just "highest version that we support that we find in signingconfig" -- so even if we support v2, it's not an error if signingconfig does not contain rekor v2 instances 

I'll think about if I can solve the clarity issue somehow...

di: > but otherwise the default is just "highest version that we support that we find in signingconfig" -- so even if we support v2, it's not an error if signingconfig does not contain rekor v2 instances

I'm suggesting that the "highest version we support" default to v1 before 4.0.0, and change not whether the signingconfig has changed. I still think the user can still choose rekor version 2 in any version (at which point we will check if it's present in the signingconfig, or error) but we should not quietly change the default before a major version bump.

jku: > we should not quietly change the default before a major version bump

I don't think anyone wants that: the changes currently in main are certainly worth a major version bump (both from API changes and from the impact on CLI). If we want to release further 3.x versions it makes sense to make a branch from 3.6.4, so before any rekorv2 support landed.

Current situation as I see it is
* sigstore-python 3.x supports rekor v1 only (applies to both verify and sign)
* sigstore-python 4.0 will support both rekor versions: verify will work for both and signing chooses the version based on signinconfig (and `--rekor-version` if this PR is merged)

If you wanted to make a release that has rekor2 support that just isn't enabled at all without the user enabling a flag, that's possible... but that release should still be called 4.0 because of API changes

di: > If we want to release further 3.x versions it makes sense to make a branch from 3.6.4, so before any rekorv2 support landed.

OK! This makes sense to me then.

jku:
Marking ready for review: I'm still not sure if this is actually useful but happy to include (and improve) it if others feel like it's useful. 

sigstore-python 3.6.5 correctly verifies a signature bundle created with sigstore-python from this branch if `--rekor-version 1` is used.

jku:
any opinions on this @woodruff or @di ?

woodruffw (approved):
LGTM -- no strong opinion, but one thing we _could_ do here is make `--rekor-version` a global argument instead, similar to `--staging` works. I don't have a strong intuition on whether that's a better fit or not.

jku:
> one thing we could do here is make --rekor-version a global argument instead, similar to --staging works

I will merge as is for two reasons:
* I think there is no good reason to limit verification to specific rekor versions -- For signing it makes some limited sense since there are compatibility concerns: are other users able to verify the signatures we produce.
* extending this to verification would make trust root management more complex: this seems even less desirable than making singing config more complex

PR #1482, opened 2025-07-28
verifier: Fix TSA cert chain construction

jku:
Using add_root_certificate() multiple times looks like a bug: it usually ends up working because 
* the API accepts multiple roots and handles roots and intermediates in the same way
* typically the signing cert is actually included in the timestamp itself

However, in the case of timestamps with no embedded certs verification fails with

`Certificates neither found in the answer or in the Verification Options`

This is understandable since the rfc3161-client Verifier thinks it does not have the actual signing certificate (as it's actually added as root certificate). This case should succeed since our certificate chain contains the signing certificate -- so it is not required in the timestamp itself.

FYI @DarkaMaul


jku:
This seems correct to me and fixes the conformance issue in https://github.com/sigstore/sigstore-conformance/pull/230

I could try to add a unit test for this as well but not today: happy to take review opinions on this

sigstore/verify/verifier.py
```diff
@@ -128,9 +128,18 @@ def _verify_signed_timestamp(
...
+                _logger.debug("Unable to verify Timestamp: cert chain is incomplete")
+                continue
+
+            builder = (
+                VerifierBuilder()
+                .tsa_certificate(certificates[0])
+                .add_root_certificate(certificates[-1])
+            )
+            for certificate in certificates[1:-1]:
+                builder = builder.add_intermediate_certificate(certificate)
```
woodruffw, line 142: Flagging, nonblocking: we _could_ add these with `add_root_certificate` instead, since they're semantically root certs ("root" meaning "in the root program," not "is a self-signed certificate"). That would be _slightly_ faster, since our verification would then be one-hop (`TSR -> TSA -> TSA-CA`) rather than having to chain `TSA-CA` up to its "root."

In practice that's probably a marginal performance benefit, however.

PR #1485, opened 2025-07-29
Parallel sign with fixed tests

jku:
This is attempt number 2 for #1468 (that was reverted because windows tests started failing), now with fixed tests.

The second commit contains the new changes -- only write to test specific temp directory in tests.


jku:
The annoying part is that the failing tests are skipped in PRs... I will try enabling the tests in my fork

woodruffw:
Yeah, this testing setup is continuing to show signs of unnecessary stress. I think we're at the point now where we should consider dropping our first-party OIDC usage entirely, and switch just to the public OIDC beacon for tests -- that would eliminate all of the PR test limitations and would also make local runs much easier.

woodruffw (approved):

PR #1489, opened 2025-07-30
Clean up verified time handling

### LINKED ISSUE #1488: Make timestamp/established time policy enforcement clearer
Currently our established time policy is a little weird and I'd like to to make it clear and correct:
 * we don't (yet) have a way to define more complex policies like requiring more than one timestamp : this is fine
 * The default policy is or should be _"require at least one established time from a source listed in trusted_root.json"_ where established time is a TSA timestamp or alternatively a rekor v1 integrated time
* Currently there are two issues:
   * If the bundle contains TSA timestamps, then one of the TSA timestamps is ALSO required to be valid (even if the entry is rekorv1 entry and has a valid integrated time). This means an invalid bundle can be made valid by just removing a timestamp from it
   * If a rekor v2 entry did somehow contain an integrated time, this would be a valid established time for the time policy code ... (maybe this is not technically possible but I'd like to see it in the code )

The first issue currently prevents the following scenario
* sign bundle with `--rekor-version 1` #1471 This leads to a bundle with rekor v1 entry and a TSA timestamp
* verify bundle with previous release 3.6.4 (currently fails with `not enough timestamps validated to meet the validation threshold (0/1)` maybe because the timestamp signing key type was not yet supported?)

I don't know how important this is but it feels like a source of future issues.

jku:
DRAFT While I think about it -- there's a few ways to handle this, not sure this is right.

Try to handle TSA timestamps and rekor v1 integrated time in a sensible manner:
* no special cases for when TSA timestamps are present
* require one verified time by default
* Only allow integrated time to be a verified time if entry is from rekor v1
* VERIFY_TIMESTAMP_THRESHOLD now refers to "number of verified times", not just TSA timestamps
* Tests use a rekor v1 bundle but expect it to be invalid if the timestamp is invalid -- but the integrated time is enough. Fix this by monkeypatching VERIFY_TIMESTAMP_THRESHOLD

Fixes #1488.

This would be back portable to make 3.x branch be able to verify newer bundles (that contain an "extra"  TSA timestamp)

### DIFF
--- a/sigstore/verify/verifier.py
+++ b/sigstore/verify/verifier.py
@@ -215,13 +215,6 @@ class Verifier:
                 raise VerificationError(msg)
 
             timestamp_from_tsa = self._verify_timestamp_authority(bundle)
-            if len(timestamp_from_tsa) < VERIFY_TIMESTAMP_THRESHOLD:
-                msg = (
-                    f"not enough timestamps validated to meet the validation "
-                    f"threshold ({len(timestamp_from_tsa)}/{VERIFY_TIMESTAMP_THRESHOLD})"
-                )
-                raise VerificationError(msg)
-
             verified_timestamps.extend(timestamp_from_tsa)
 
         # If a timestamp from the Transparency Service is available, the Verifier MUST
@@ -232,6 +225,12 @@ class Verifier:
         if (
             timestamp := bundle.log_entry.integrated_time
         ) and bundle.log_entry.inclusion_promise:
+            kv = bundle.log_entry._kind_version
+            if not (kv.kind in ["dsse", "hashedrekord"] and kv.version == "0.0.1"):
+                raise VerificationError(
+                    "Integrated time only supported for dsse/hashedrekord 0.0.1 types"
+                )
+
             verified_timestamps.append(
                 TimestampVerificationResult(
                     source=TimestampSource.TRANSPARENCY_SERVICE,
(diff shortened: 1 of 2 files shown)


woodruffw:
Thanks for poking at this @jku! Let me know when I should do a review.

> * VERIFY_TIMESTAMP_THRESHOLD now refers to "number of verified times", not just TSA timestamps

Perhaps we could make it `SIGNED_TIMESTAMP_THRESHOLD` to make it clear that the check is now (correctly) more generic?

> This would be back portable to make 3.x branch be able to verify newer bundles (that contain an "extra" TSA timestamp)

Yeah, I think that would be ideal!

sigstore/verify/verifier.py
```diff
@@ -232,6 +225,12 @@ def _establish_time(self, bundle: Bundle) -> list[TimestampVerificationResult]:
         if (
             timestamp := bundle.log_entry.integrated_time
         ) and bundle.log_entry.inclusion_promise:
+            kv = bundle.log_entry._kind_version
+            if not (kv.kind in ["dsse", "hashedrekord"] and kv.version == "0.0.1"):
```
jku, line 229: this check will not work when backported to 3.x (as log_entry does not have kindversion there) but the check can be removed there -- only 0.0.1 versions are supported anyway

jku:
> > * VERIFY_TIMESTAMP_THRESHOLD now refers to "number of verified times", not just TSA timestamps
> 
> Perhaps we could make it `SIGNED_TIMESTAMP_THRESHOLD` to make it clear that the check is now (correctly) more generic?

I originally avoided this thinking about the potential backport and the fact that technically this is API... but I think I agree with you:
* the semantics are already changing, better change the name as well
* this is only technically part API, in reality it's more a marker for "this is where you would plug in timestamp policy management if it existed"

jku:
This should be ready. 

It's backportable to 3.x  (with the small caveat mentioned in a code comment) and that makes bundles created with current main using `--rekor-version 1` (#1471) verifiable with a 3.x client.

woodruffw (approved):
LGTM, thanks @jku!

How do you want to do the backport here? I figure it probably makes sense to land before #1470.

jku:
> How do you want to do the backport here? I figure it probably makes sense to land before #1470.

Thanks, I'll merge this now. I'll make the backport a new PR today/tomorrow

PR #1496, opened 2025-08-08
feat: expose ClientTrustConfig as a public class

### LINKED ISSUE #1019: Public API: Expose `ClientTrustConfig`
`ClientTrustConfig` is currently a private type, via `sigstore._internal.trust`. We should consider making it public, so that methods like `Verifier.from_trust_config(...)` can also be made public.

See: https://github.com/sigstore/sigstore-python/pull/1010#discussion_r1603154748

SequeI:
#### Summary

Exposing ClientTrustConfig. Since recent changes to SigningContext.staging/production being moved to ClientTrustConfig, it would be better for this class to be public as it is/will be used in Model-Transparency and such. The API has had massive changes as is, so I think it would be better doing this now for next release.

Resolves #1019 

#### Release Note
Added to changelog.md

The ClientTrustConfig class has been moved from the private _internal package to a new public
module (sigstore.trust). This change formally adds the class to the project's public API,
making it available for use in other projects.


jku:
> Noticed the rekorV2 tests are wishy washy

If you mean that the signing tests using staging rekor2 fail with 50x errors more often then is reasonable: yes this seems to be the case. I'm following up in sigstore/rekor-tiles repo

woodruffw:
@SequeI Could you avoid force-pushing unless absolutely necessary? It causes a notification to me (and presumably anything one else subscribed on the PR) each time you to 🙂 

(It's okay to have a non-clean history on this PR -- we'll clean it up with the actual merge.)

SequeI:
@woodruffw Apologies! Won't do that anymore 

Would it make sense to have a new public module (trust like previously) and just expose trustedRoot and SigningConfig alongside ClientTrustConfig due to it's dependence on it? Putting it into models creates a lot of circular import errors with rekor and rekorv2 unfortunately

SequeI:
Sure :

```
    from sigstore._internal.rekor.client import RekorClient
../../.local/lib/python3.13/site-packages/sigstore/_internal/rekor/client.py:41: in <module>
    from sigstore.models import TransparencyLogEntry
../../.local/lib/python3.13/site-packages/sigstore/models.py:65: in <module>
    from sigstore._internal.trust import SigningConfig, TrustedRoot
../../.local/lib/python3.13/site-packages/sigstore/_internal/trust.py:43: in <module>
    from sigstore._internal.rekor.client_v2 import RekorV2Client
../../.local/lib/python3.13/site-packages/sigstore/_internal/rekor/client_v2.py:41: in <module>
    from sigstore.models import TransparencyLogEntry
E   ImportError: cannot import name 'TransparencyLogEntry' from partially initialized module 'sigstore.models' (most likely due to a circular import) (/home/asiek/.local/lib/python3.13/site-packages/sigstore/models.py)
```

```
test/unit/conftest.py:40: in <module>
    from sigstore._internal.rekor.client import RekorClient
../../.local/lib/python3.13/site-packages/sigstore/_internal/rekor/client.py:41: in <module>
    from sigstore.models import TransparencyLogEntry
../../.local/lib/python3.13/site-packages/sigstore/models.py:65: in <module>
    from sigstore._internal.trust import SigningConfig, TrustedRoot
../../.local/lib/python3.13/site-packages/sigstore/_internal/trust.py:44: in <module>
    from sigstore._internal.rekor.client import RekorClient
E   ImportError: cannot import name 'RekorClient' from partially initialized module 'sigstore._internal.rekor.client' (most likely due to a circular import) (/home/asiek/.local/lib/python3.13/site-packages/sigstore/_internal/rekor/client.py)
```

woodruffw:
> I wasn't sure if this would be in the scope of the PR, or if I should fix this etc, but I'll fix these in this case

Thank you! Yes, I'd consider it in scope, and I appreciate you fixing them 🙂

jku (dismissed):
I've been avoiding reviewing since both of these feel like code smells:
* ClientTrustConfig is moved to models and SigningConfig/TrustedRoot remain in _internal -- all of them are essentially public API now though
* the import within function

That said, I've tried to come up with better solutions and so far have no better ideas: API publicity was discussed already and the import trick seems required if we want to keep the rekor client selection a SigningConfig implementation detail.

so I'm happy with this. @woodruffw do you want to have another look?

woodruffw:
> * ClientTrustConfig is moved to models and SigningConfig/TrustedRoot remain in _internal -- all of them are essentially public API now though

I think we have two options here:

1. We could explicitly mark `ClientTrustConfig.trusted_root` and `ClientTrustConfig.signing_config` as private APIs, i.e. with `@private` like elsewhere in the codebase.
2. We could make these types public as part of `sigstore.models` as well.

I don't have a strong preference between these -- (1) is probably simpler but we'll likely want to do (2) eventually anyways so maybe we should just get it out of the way.

> the import within function

Yeah, I don't love this -- I (personally) find function-scoped imports to be a code smell and much harder to debug (since they modify global import state "late" in execution).

I'm OK with merging here if there's no clean solution, but I'd like to better understand where the import cycle is here so that we can eventually fix it.

PR #1510, opened 2025-08-15
Improve case of possible wrong instance on verify

### LINKED ISSUE #1487: accidentally verifying with wrong sigstore instance looks confusing
If I do `sigstore --staging sign ...` and then do `sigstore verify identity ...` (accidentally without specifying --staging) the error I get is `not enough timestamps validated to meet the validation threshold (0/1)`

It's not incorrect but it would be nice if we could be somehow more informative.

jku:
Fixes #1487
* moves rfc3161-client exception details to debug log level
* Offers a hint if TSA cert chain verification fails: this is a sign that possible the bundle uses a different sigstore instance than the verification expects

### DIFF
--- a/sigstore/_cli.py
+++ b/sigstore/_cli.py
@@ -46,7 +46,7 @@ from sigstore.dsse._predicate import (
     SLSAPredicateV0_2,
     SLSAPredicateV1_0,
 )
-from sigstore.errors import Error, VerificationError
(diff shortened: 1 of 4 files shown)


jku:
Maybe I went a little too far in removing details from the error putput. Could output one line of error details and then a separate hint

PR #1514, opened 2025-08-20
rekor: Fix checkpoint signature lookup

### LINKED ISSUE #1513: issues with multiple checkpoint signatures
There are some issues with C2SP checkpoints as documented in https://github.com/sigstore/sigstore-conformance/pull/241
* at least we should not expect all signatures to be from the log key

jku:
Checkpoint can have multiple signatures from multiple keys. We just want one of them to be the log key.

This looks like it will fix #1513


jku:
@haydentherapper can you confirm this is what we want here: the log must have _a_ signature in the checkpoint, otherwise we don't care about the signatures at this point.

jku:
I'll merge as is. I was debating myself whether I need to add more tests for checkpoint... but conformance tests for this, that should be good enough

PR #1515, opened 2025-08-20
Update embedded TUF root

jku:
* Update the related Makefile rules
  * Include the signingconfig file we now use
  * Add rule for staging (I kept the rules separate so we don't get issues like #1473 for staging content that can change a lot more often)
* Remove TUF repository mock from tests to simplify maintenance (see comment further down)
* Update TUF root for public good instance, both staging and production

The files can be verified by running `make update-embedded-root update-embedded-root-staging` -- if there's no change to the files in git, they are correct.

### DIFF
--- a/test/unit/internal/test_trust.py
+++ b/test/unit/internal/test_trust.py
@@ -17,8 +17,6 @@ import os
 from datetime import datetime, timedelta, timezone
 
 import pytest
-from cryptography.hazmat.primitives.serialization import Encoding, PublicFormat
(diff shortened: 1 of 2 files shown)


jku:
I'm removing the TUF mocking from our tests: it was useful to really verify that every request is expected but maintenance is a bit of a hassle. If other maintainers would rather keep it, I can update the test assets instead (need to copy some more files from the tuf repo and update some expected test results).

PR #1565, opened 2025-10-06
Better error for verifying rekorv2 on 3.6.x

### LINKED ISSUE #1564: consider improving compatibility message in 3.6.x
Currently verifying a signature bundle that contains a rekorv2 entry using sigstore-python  3.6.5 fails with

     VerificationError: not enough sources of verified time

This is not incorrect but it's also not very useful: we should inform the user that client upgrade is required.

I'll have a look, maybe there is an easy improvement here.

jku:
* Improve error message when verifying with 3.6.x but the bundle contains a rekor v2 entry: Current error is technically correct but not very helpful
* Bump version as well (I can drop this if we don't want to make point release)

Fixes #1564.


di (approved):
(we should do the same for the 4.x series as well...)

jku:
> Speaking of cryptography version pin bumps: should we do that in 3.6.6 or no? On one hand we would like library users to upgrade to sigstore 4.0 but I also recognise how very annoying a library pinning another library (especially cryptography) is...

I will make a PR of that but either way works for me.

PR #1575, opened 2025-10-11
CI: Add test that checks that old releases can verify

### LINKED ISSUE #1573: test verifying new bundles with old clients
The test suite sort of naturally includes tests for verifying old bundles with new client... but it would be useful to ensure that bundles created by current client are verified correctly by expected old clients -- I think we don't have any tests like this.

* job 1: 
  * checkout current branch
  * sign
  * upload bundle as artifact
* job2: matrix of older git tags
  * checkout tag
  * download artifact
  * verify bundle

This might be a little more involved (e.g. we don't expect 3.x to verify bundles with rekorv2 entries)

jku:
Making sure that old clients can verify bundles created by current code (and current public good infrastructure) is not too difficult and would be very useful.

This is an initial attempt, keeping it draft for now:
* I don't like that the tested versions are hard coded... but I don't think there is much alternative here, testing all tags is silly. It could be easy to forget to add releases here but I don't have better solutions
* I think we only want to test last patch release of every minor release that we still care about
* I would like to say we still care about v3.5.3 (released Nov 2024) but that release no longer handles current trusted root correctly and we did not patch that in 3.5.x: #1578 is a potential fix for that
* It might be useful to run this on cron as well (since trust root changes affect the results) ... but then this would need a file-an-issue job as well. Filed #1583 to follow up

Example run: https://github.com/jku/sigstore-python/actions/runs/18432104146

Fixes #1573


jku:
I'm marking this ready for review: If #1578 happens, we can add 3.5.3 to the test matrix

.github/workflows/cross-version-verify.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
+          ref: ${{matrix.version}}
+          fetch-tags: true
+      - uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
+        with:
+          python-version: "3.x"
+          cache: "pip"
+          cache-dependency-path: pyproject.toml
```
woodruffw: Noting: at some point I suspect some of our older releases will stop working with the latest Python, if only because of dependency slippage. But that's not a bridge we need to cross with this PR 🙂 

(If this becomes an issue, we can make the matrix parametric on needed Python versions.)

PR #1645, opened 2026-01-07
Stop supporting Python 3.9

jku:
We could just stop testing with it (to unblock the other open PRs), but removing support in pyproject.toml seems more honest: otherwise we just break the support later on.

unblocks #1643, #1644


woodruffw (approved):
Fully agreed, thanks @jku!

PR #1649, opened 2026-01-08
chore: remove an unused nonce from OAuth flow

woodruffw:
Our OAuth flow uses `authorization_code`, in which `nonce` is not required and has no effect.


jku (approved):
Agreed

PR #1659, opened 2026-01-16
Include SCT extension in signature data

cmurphy:
If an SCT includes an extension, its signature is signed over the entire data structure including the extension.

Followup to https://github.com/sigstore/sigstore-python/pull/1657
Relates to https://github.com/sigstore/rekor-tiles/issues/73
Fixes conformance failure in https://github.com/sigstore/sigstore-conformance/actions/runs/21049488308/job/60532184810?pr=319


woodruffw (approved):
Thanks @cmurphy! The conformance test is good enough for merge here IMO, but it'd be great to also have a unit test in tree for this (I think we have other tests for SCTs that lack these extension bytes.)

PR #1727, opened 2026-03-27
chore: ignore macOS .DS_Store files

enlightened88:


di:
Hi, this project does not produce `.DS_Store` files, [you should add this to your global `.gitignore` instead](https://dev.to/lpheller/quickly-ignore-dsstore-files-globally-ecd).

PR #1732, opened 2026-04-07
Re-use rekor/timestamp connections (thread-locally)

### LINKED ISSUE #1730: better connection re-use
When Rekor v2 support was added we felt multi-threading was needed (since otherwise signing multiple artifacts takes forever). As a result, the client now overuses connections when lots of artifacts are signed (because I was unsure about the state of requests Session thread-safety). The plan was to revisit this after #1653.

However, I'm now seeing some issues in a multi-artifact use case where we end up using more connections than is allowed on the host environment... and I have a fairly simple solution:
* bring back Session re-use
* but make the Session thread-local

this way even if there is some unknown thread safety issue, we won't see it. If someone promises that Session is safe (or we use urllib3 directly) it will be easy to just drop threading.local()

I will take a look next week

jku:
Fixes #1730 

* session reuse was removed from rekor client when rekor2 support was added (as it was not clear if Session is thread safe, but we really needed threads for rekor 2 signing to perform)
* This becomes an issue if users sign hundreds of artifacts with the same Signer: in that case we use too many connections/ports
* This PR reintroduces Session reuse, but keeps it thread local -- so even if there is a thread safety issue we should not be affected
* I'm fine with also not using `threading.local()` if it's clear that thread safety is not an issue


jku:
Documenting the practical effect of this PR:
* sigstore-python CLI uses ThreadPoolExecutor to run signing in threads with currently no `max_workers` set: so in practice with this PR the CLI (when called with enough artifacts to sign) uses `min(32, os.cpu_count() + 4)` connections to each host. 32 is still a bit much but should be acceptable.
* API users like pypi-attestations can decide their thread use independently: currently pypi-attestations runs in a single thread so will use a single connection per host

woodruffw (approved):
Thanks @jku! Seems reasonable to me 🙂

Issue #1656, opened 2026-01-15
incompatibility with new CT log


jku:
staging tried to enable https://log2026-1.ctfe.sigstage.dev, a new CT endpoint running [TesseraCT](https://github.com/transparency-dev/tesseract)

for some reason tests only fail sometimes (will have to check if there's any randomness to the ct endpoint selection), but this error has been seen with sigstore-python: 

```
[22:25:39] ERROR    Unexpected trailing extension bytes             errors.py:42
                    For detailed error information, run sigstore                
                    with the `--verbose` flag.                                 
```

links:
* signing event that adds the new ct URL https://github.com/sigstore/root-signing-staging/pull/364
* plan is to at least partially rollback https://github.com/sigstore/root-signing-staging/pull/371

woodruffw:
Ah yeah, that's probably safe for us to remove as a check 🙂

PR #1738, opened 2026-04-09
tests: Update tests to be compatible with current staging

### LINKED ISSUE #1737: A hard coded trustconfig in tests is leading to failures
#1736 , #1734, #1735

Staging just changed the CT log key. This is likely breaking some tests that are hard coding a trust config but still using the staging environment.

Need to either update the config or make the tests not hard code the config -- at least the tsa tests are not necessarily that useful

### LINKED ISSUE #1739: Cross-version verify failure on staging
cross-version-verify.yaml uses old versions to verify a bundle created with current sigstore-python and current sigstore staging infra.
* Tests are currently failing because staging  changed Fulcio to one that uses tesseraCT: #1656 means our old versions are incompatible: >=4.2.0 and >=3.6.7 are compatible
* There is no immediate plan to start using TesseraCT in production (because a few clients were incompatible) but it will happen at some point. Staging will keep using TesseraCT from now on.

I think we should switch cross-version-verify to use production env too, and just acknowledge that old versions do not verify on current staging bundles

jku:
Staging changed the CT log implementation:
* key changed
* #1656 is now in play (so old sigstore-python releases will have issues with new staging bundles)

## Update TSA tests

* We have some localhost TSA tests already that test TSA client specifically
* The removed signing tests are mostly redundant as the rekor v2 tests already require the staging TSA to work and the TSA client tests check for failure modes
* The removed tests are fragile since they require a hard coded TrustConfig but still use most of the staging infra -- currently CI is red because the CT key changed

## Update cross-version tests

* Replace 3.6.6 with 3.6.7 as the latest release in that branch
* Start testing on both prod and staging
* Skip 3.5.6, 4.0.0 and 4.1.0 on staging: They now fail because of #1656

### DIFF
--- a/.github/workflows/cross-version-verify.yaml
+++ b/.github/workflows/cross-version-verify.yaml
@@ -42,8 +42,9 @@ jobs:
       - name: Sign 
         run: |
           touch artifact
-          python -m sigstore --staging sign --bundle artifact-rekor2.sigstore.json --identity-token $(cat oidc-token.txt) --rekor-version=2 artifact
-          python -m sigstore --staging sign --bundle artifact-rekor1.sigstore.json --identity-token $(cat oidc-token.txt) --rekor-version=1 artifact
+          python -m sigstore --staging sign --bundle artifact-staging-rekor2.sigstore.json --identity-token $(cat oidc-token.txt) --rekor-version=2 artifact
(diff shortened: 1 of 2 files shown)


jku:
@DarkaMaul you might have an opinion on whether the signing tests with custom tsa are still required: I think they're not

woodruffw (approved):

Issue #67, opened 2022-05-02
Allow disabling of ambient OIDC credential detection


woodruffw:
Related to #31 and #50: we should support *disabling* ambient credential detection and use, with the same CLI flag that `cosign` uses (`--oidc-disable-ambient-providers` as of https://github.com/sigstore/cosign/pull/1832).

woodruffw:
`cosign` just merged this, so I'll tackle it now.

PR #1744, opened 2026-04-13
Add timeout to Fulcio HTTP requests to prevent indefinite hangs

MichaelMVS:
The FulcioSigningCert.post() and FulcioTrustBundle.get() methods were missing timeout parameters on their session.post() and session.get() calls, which could cause indefinite hangs if the Fulcio server is slow or unresponsive.

Added _CLIENT_TIMEOUT = 30 (consistent with other HTTP timeouts in the sigstore-python codebase) and applied it to both methods.

This mirrors the fix applied in #67 for the Rekor client and #1 for other endpoints in the codebase. Without timeouts, sigstore-python can hang indefinitely when network issues or server problems occur.

### DIFF
--- a/sigstore/_internal/fulcio/client.py
+++ b/sigstore/_internal/fulcio/client.py
@@ -37,6 +37,8 @@ from sigstore._internal import USER_AGENT
 from sigstore._utils import B64Str
 from sigstore.oidc import IdentityToken
 
+_CLIENT_TIMEOUT: int = 30
+
 _logger = logging.getLogger(__name__)
 
 SIGNING_CERT_ENDPOINT = "/api/v2/signingCert"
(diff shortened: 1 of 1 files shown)


jku:
> This mirrors the fix applied in #67 for the Rekor client and #1 for other endpoints in the codebase.

?

jku:
> The FulcioSigningCert.post() and FulcioTrustBundle.get() methods were missing timeout parameters on their session.post() and session.get() calls, which could cause indefinite hangs if the Fulcio server is slow or unresponsive.

I am not opposed, this seems reasonable.

The comment about this being consistent with the rest of the code base seems incorrect: AFAIK the only timeouts currently in use are
* 5 secs for timestamp timeout (this is arguably too aggressive -- TSAs typically respond quickly but networking hickups still happen)
* TUF requests (done by python-tuf) have a default timeout of 30 seconds IIRC

All other requests (fulcio, rekor, rekor2, oidc) have no timeout at all. If we are setting timeouts, I think we should set them for all HTTP clients:
* 30 secs seems like an ok default value (timestamp could use that  global default too)
* We already define a common user agent in sigstore/_internal/__init__.py , that could be a decent place for a default timeout as well
* The one exception *might* be rekor v2 where the request processing can take a some seconds... A 30 sec timeout should still work but the rekorv2 value could be a little higher than the common default as well (like 30+10 secs)

PR #1763, opened 2026-05-04
uv-ify our development

woodruffw:
This switches us from `pip` (and `build`, etc.) to uv. The main advantage in our context is a universal lockfile, which means much faster resolutions + development iterations when dependencies change.

Some things I'll need to do with this PR:

- [x] Bootstrap uv in the CI as well
- [x] Switch our Dependabot usage to uv
- [ ] ~~Change our `requirements.txt` export to `uv export` (this will hopefully also be less painful to maintain over time)~~

Turns out we can't easily do `uv export --format=requirements.txt --no-dev` for this, since the lockfile is for the editable resolution. We could probably do something like the `requirements.in` approach but IMO not worth it for this iteration.

### DIFF
--- a/.github/workflows/requirements.yml
+++ b/.github/workflows/requirements.yml
@@ -43,4 +43,6 @@ jobs:
           python-version: ${{ matrix.python_version }}
 
       - name: Run test install
-        run: uv pip install --require-hashes --no-deps -r install/requirements.txt
+        run: |
+          uv venv
+          uv pip install --require-hashes --no-deps -r install/requirements.txt
--- a/.github/workflows/requirements.yml
+++ b/.github/workflows/requirements.yml
@@ -25,7 +25,7 @@ jobs:
       SIGSTORE_REF: ${{ inputs.ref }}
     strategy:
       matrix:
-        python_version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        python_version: [ "3.10", "3.11", "3.12", "3.13", "3.14" ]
 
     steps:
       - name: Populate reference from context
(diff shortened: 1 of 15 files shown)


woodruffw:
This is like 300% faster for most of our smaller CI steps, and about the same for the larger ones (where the runtime isn't dominated by `pip`).

~~I need to figure out the best way to deal with the conformance workflow, since the conformance script both executes _and_ imports sigstore.~~

.github/workflows/requirements.yml
```diff
@@ -38,12 +38,11 @@ jobs:
...
         with:
           python-version: ${{ matrix.python_version }}
-          allow-prereleases: true
-          cache: "pip"
 
       - name: Run test install
-        run: python -m pip install --require-hashes --no-deps -r install/requirements.txt
+        run: |
+          uv venv
+          uv pip install --require-hashes --no-deps -r install/requirements.txt
```
jku, line 48: is this going to work WRT "--prerelease-allow" that used to be an issue at least previously? 

Same question for another "uv pip install" elsewhere

jku (approved):
This still makes life quite annoying for folks in [PEP 668 externally-managed-environments](https://packaging.python.org/en/latest/specifications/externally-managed-environments/#externally-managed-environments) but I guess if it's just me then I just add yet another workaround.

.github/workflows/requirements.yml
woodruffw: It should, yeah, since we no longer have any prereleases in our dependency tree (we got rid of them when we dropped the transitive dep on betterproto).

PR #1768, opened 2026-05-08
oidc: allow setting redirect server port

### LINKED ISSUE #1029: Allow setting of redirect uri port
**Description**

I'm attempting to leverage `sigstore-python` library for an enterprise signing/verification tool. Working on a proof of concept resulted in this small example:
```python

issuer = Issuer("https://auth.company.com/")

identity = issuer.identity_token(client_id= "signaturevalidationtool", client_secret="super-secret")
```

Went through the process of setting up an app with our oauth manager to facilitate authentication, however the redirect URI cannot use wildcards with localhost ports. This security configuration choice leaves me stuck, since `sigstore-python` uses ephemeral ports.

Could we add add the ability to specify this port? Cosign already has a flag for [this](https://github.com/sigstore/cosign/blob/2bb2e88ed1f1267bd52b5ed608d65ce3a9413078/doc/cosign_sign-blob.md?plain=1#L53)

antonio-mazzini:
## Summary

Adds an optional `redirect_port` argument to `Issuer.identity_token`
so callers can bind the local OAuth redirect server to a fixed port
instead of always using an ephemeral one.

Some enterprise OIDC providers (and Microsoft Entra ID under certain
configurations) require a pre-registered redirect URI and do not allow
wildcards on `localhost` ports, which currently blocks `sigstore-python`
from being used in those environments. `cosign` already exposes a
`--oidc-redirect-url` flag that solves the same problem on the CLI side.

The default of `0` preserves the existing behaviour (OS-assigned port).

Closes #1029

## Test plan

- [x] `make lint` — ruff/mypy/bandit/interrogate green
- [x] `pytest test/unit` — 177 passed, no regressions
- [x] New unit test `test_identity_token_passes_redirect_port` verifies
      the parameter is forwarded to the OAuth flow

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+### Added
+
+* `Issuer.identity_token` accepts an optional `redirect_port` argument to bind
+  the local OAuth redirect server to a fixed port, for OIDC providers that
+  require a pre-registered redirect URI without `localhost` port wildcards
+  ([#1029](https://github.com/sigstore/sigstore-python/issues/1029))
+
 ### Fixed
 
 * Fixed ~60s hang after completing browser-based OIDC authentication.
--- a/sigstore/_internal/oidc/oauth.py
+++ b/sigstore/_internal/oidc/oauth.py
@@ -102,12 +102,18 @@ AUTH_SUCCESS_HTML = """
 
 
 class _OAuthFlow:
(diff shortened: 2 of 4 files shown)


jku (approved):
Seems reasonable to me. 

The default values in _OAuthRedirectServer and _OAuthFlow look unnecessary since we always provide a value but that seems like a nitpick.

woodruffw (approved):
LGTM as well 🙂

jku:
Thanks!

Issue #1771, opened 2026-05-13
dependabot failures with uv


jku:
These dependabot issues appeared when we started using the uv ecosystem in dependabot.yml

## issue 1

dependabot ends up running this when updating cryptography:

```
uv pip compile --build-isolation --output-file\=install/requirements.txt --generate-hashes -P cryptography==48.0.0 install/requirements.in
  × No solution found when resolving dependencies:
  ╰─▶ Because sigstore==4.2.0 depends on cryptography>=42,<47 and
      cryptography==48.0.0, we can conclude that sigstore==4.2.0 cannot be used.
      And because you require sigstore==4.2.0, we can conclude that your
      requirements are unsatisfiable.
```
this means dependabot silently fails . While this particular command only tries to update requirements.txt (meaning the failure is  "correct"), apparently this also blocks the pypproject.toml/uv.lock update -- or at least it's not appearing when expected?

## issue 2

pydantic update also silently fails: this looks like a dependabot bug (but did not appear with pip) :

```
updater | 2026/05/13 15:39:16 ERROR <job_1365984690> Error processing pydantic (Dependabot::SharedHelpers::HelperSubprocessFailed)
2026/05/13 15:39:16 ERROR <job_1365984690> Traceback (most recent call last):
  File "/opt/python/run.py", line 18, in <module>
    print(hasher.get_dependency_hash(*args["args"]))
          ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/opt/python/lib/hasher.py", line 13, in get_dependency_hash
    hashes = hashin.get_package_hashes(
        dependency_name,
    ...<2 lines>...
        index_url=index_url
    )
  File "/usr/local/.pyenv/versions/3.14.2/lib/python3.14/site-packages/hashin.py", line 654, in get_package_hashes
    data = get_package_data(package, index_url, verbose)
  File "/usr/local/.pyenv/versions/3.14.2/lib/python3.14/site-packages/hashin.py", line 582, in get_package_data
    content = json.loads(_download(url))
                         ~~~~~~~~~^^^^^
  File "/usr/local/.pyenv/versions/3.14.2/lib/python3.14/site-packages/hashin.py", line 60, in _download
    r = urlopen(url)
  File "/usr/local/.pyenv/versions/3.14.2/lib/python3.14/urllib/request.py", line 187, in urlopen
    return opener.open(url, data, timeout)
[...]

PR #1773, opened 2026-05-14
Bump the ceiling on cryptography version

jku:
This is done manually because dependabot currently fails to do this (#1771) but we really should have a release that is compatible with current cryptography.

* Allow cryptography 48.x
* Allow tuf 7.x
* Also update uv.lock (this brings in a handful of other updates)


jku:
I'm adding a similar ceiling bump for tuf in this same PR

woodruffw (approved):

PR #1774, opened 2026-05-14
Prepare 4.3.0 release

jku:
* rewrote changelog a bit to keep to two lines per item if possible
* Added the missing items

PR #1776, opened 2026-05-14
Encode DSSE as hashedrekord for Rekor v2

codysoyland:
## Summary

Implements [rekor-v2-spec §6.1.4](https://github.com/sigstore/architecture-docs/pull/63): for Rekor v2, a DSSE envelope is submitted and verified as a `hashedrekord/0.0.2` entry where:

- `digest = Hash(PAE(payloadType, payload))` (Hash is the externalized hash of the signing algorithm per the algorithm registry)
- `signature.content = envelope.signatures[0].sig`

Producer change in [`sigstore/_internal/rekor/client_v2.py`](sigstore/_internal/rekor/client_v2.py): build a `HashedRekordRequestV002` instead of a `DSSERequestV002`. Verifier change in [`sigstore/verify/verifier.py`](sigstore/verify/verifier.py): accept `hashedrekord/0.0.2` for DSSE bundles and reconstruct the expected entry body from the envelope's PAE digest and signature.

## Related

- Spec change: sigstore/architecture-docs#63
- Conformance tests (draft): sigstore/sigstore-conformance#371 — this branch passes the new `rekor2-dsse-*` fixtures.


sigstore/_internal/rekor/client_v2.py
```diff
@@ -137,13 +138,25 @@ def _build_dsse_request(
         cls, envelope: Envelope, certificate: Certificate
     ) -> EntryRequestBody:
         """
-        Construct a dsse request to submit to Rekor.
+        Construct a hashedrekord request for a DSSE envelope.
+
+        Rekor v2 only supports the hashedrekord entry type; DSSE envelopes are
+        uploaded as a hashedrekord whose digest is `Hash(PAE(payloadType,
```
loosebazooka: This comment `PAE(payloadType, payload)` is a bit weird, I think `Hash(PAE(dsse))` or `Hash(dsse.PAE)` makes more sense. Or the underlying pae function should just take a envelope as a parameter?

sigstore/_internal/rekor/client_v2.py
```diff
@@ -37,7 +38,7 @@
     RekorClientError,
     RekorLogSubmitter,
 )
-from sigstore.dsse import Envelope
+from sigstore.dsse import Envelope, _pae
```
loosebazooka: maybe `_pae` needs to be explicitly exported (via style change)? And if so, now the style feels less fitting as an external function?

sigstore/verify/verifier.py
```diff
@@ -431,18 +434,23 @@ def verify_dsse(
...
-        if entry._inner.kind_version.kind != "dsse":
-            raise VerificationError(
-                f"Expected entry type dsse, got {entry._inner.kind_version.kind}"
-            )
-        if entry._inner.kind_version.version == "0.0.2":
+        kind = entry._inner.kind_version.kind
+        version = entry._inner.kind_version.version
+        if kind == "hashedrekord" and version == "0.0.2":
+            _validate_hashedrekord_v002_dsse_entry_body(bundle)
+        elif kind == "dsse" and version == "0.0.2":
```
loosebazooka: I do think we should remove this completely (eventually) and not leave it dangling around.

loosebazooka (commented):
just took a quick look over. @jku or @woodruffw probably is ideal for this.

codysoyland:
The conformance test failures should disappear when we merge https://github.com/sigstore/sigstore-conformance/pull/374

jku (approved):
This make sense to me, including the algorithm registry.

Our testing seems adequate:
* the hard coded asset is verified
* A roundtrip sign-and-verify test for dsse on rekor2 also exists
 
I would have preferred to have a release before these changes but that did not seem to materialize so I think we can include this in #1774.

Thanks cody and appu

jku (changes requested):
Actually, could I get a sigstore-conformance action upgrade in this PR too -- I'd do it myself but then I can't review this.

I just tagged sigstore-conformance 0.0.28 at e2cc8e51870bb9af4e4ae3342e97cf5f7ea2cdd8 -- I expect that should pass for this PR.

jku (approved):
Thanks!

PR #1795, opened 2026-06-02
verify: match email identities case-insensitively

### LINKED ISSUE #459: Email address are treated as case sensitive when verifying signatures
**Description**

When using an email identity to verify a signature, the email is currently case sensitive. E.g., 

`--identity foo@bar.com` works but `--identity FoO@baR.coM` fails. 

Verification failed with error: Certificate's SANs do not match FoO@baR.coM; actual SANs: {'foo@bar.com'}

**Version**

Latest v1.0.1

arpitjain099:
#### Summary

`policy.Identity.verify` builds the set of certificate SANs and checks the expected identity with an exact string comparison (`self._identity in all_sans`). Email addresses are not case-sensitive in practice (and Fulcio normalizes them), so an identity that differs only in case is wrongly rejected:

```
--identity foo@bar.com   -> verifies
--identity FoO@baR.coM   -> VerificationError: Certificate's SANs do not match FoO@baR.coM; actual SANs: {'foo@bar.com'}
```

This was reported downstream in sigstore/model-transparency#459, where the root cause traces to this policy.

This PR matches email (`RFC822Name`) SANs case-insensitively using `casefold()`, while keeping URI and other-name SANs as exact matches (those are case-sensitive in general, so they are deliberately left untouched). The error message still reports the certificate's original SANs.

To test: `make test` (or `pytest test/unit/verify/test_policy.py`). The new `TestIdentity::test_succeeds_email_case_insensitive` fails on `main` and passes with this change; the rest of the verify suite is unaffected.

One thing for reviewers to confirm: this scopes case-insensitivity to email SANs only, which I believe is the right and safe behavior for Fulcio-issued certs. Happy to adjust if you would prefer a narrower (domain-only) or different treatment.

Resolves sigstore/model-transparency#459

#### Release Note

Email (RFC822Name) identities are now matched case-insensitively when verifying a certificate's SANs.

#### Documentation

No documentation change required; this is a bug fix to existing identity-verification behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,11 @@ All versions prior to 0.9.0 are untracked.
   The OIDC redirect server had incomplete HTTP responses and no connection
   management, causing a keep-alive deadlock with the browser.
 
-* `policy.Identity` now matches email (RFC822Name) SANs case-insensitively, so
-  an expected identity that differs only in case (for example `FoO@baR.coM`
-  versus a `foo@bar.com` SAN) verifies successfully. URI and other SAN types
-  are still matched exactly.
+* `policy.Identity` now matches the domain part of email (RFC822Name) SANs
+  case-insensitively, so an expected identity whose domain differs only in case
+  (for example `a@TnY.ToWn` versus an `a@tny.town` SAN) verifies successfully.
+  The local part stays case-sensitive (per RFC 5321 it can be case-significant),
+  and URI and other SAN types are still matched exactly.
   ([sigstore/model-transparency#459](https://github.com/sigstore/model-transparency/issues/459))
 
 ## [4.2.0]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ All versions prior to 0.9.0 are untracked.
   The OIDC redirect server had incomplete HTTP responses and no connection
   management, causing a keep-alive deadlock with the browser.
 
+* `policy.Identity` now matches email (RFC822Name) SANs case-insensitively, so
+  an expected identity that differs only in case (for example `FoO@baR.coM`
+  versus a `foo@bar.com` SAN) verifies successfully. URI and other SAN types
+  are still matched exactly.
+  ([sigstore/model-transparency#459](https://github.com/sigstore/model-transparency/issues/459))
+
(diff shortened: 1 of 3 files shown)


jku:
So it looks like
* pure case-sensitivity (current implementation) is not technically correct as domains are not sensitive
* the local part of the address is *in practice* insensitive for almost all addresses but this is an implementation detail: at least RFC5321 says
    >  The local-part of a mailbox MUST BE treated as case sensitive.
    >   Therefore, SMTP implementations MUST take care to preserve the case
    >   of mailbox local-parts.  In particular, for some hosts, the user
    >   "smith" is different from the user "Smith". 

In Sigstore context I think we must consider the local part sensitive: making a mistake in the other direction would be worse . I don't have a strong opinion on the domain: we could make it insensitive but I don't see it as important.

arpitjain099:
Good point, thanks @jku. RFC 5321 does say the local-part can be case-significant and that SMTP must preserve it, so matching the whole address case-insensitively is too aggressive and errs in the more dangerous direction for identity verification. I will narrow this to compare only the domain case-insensitively and keep the local-part exact, and update the tests and changelog to match.

woodruffw:
Hmm, I'm not sure we can make this case-insensitive at all? Here's what RFC 5321 says about the the local-part (i.e. user) of a mailbox address:

> The local-part of a mailbox MUST BE treated as case sensitive. Therefore, SMTP implementations MUST take care to preserve the case of mailbox local-parts.

https://datatracker.ietf.org/doc/html/rfc5321#section-2.4

In other words, in general, we can't assume that `smith@example.com` and `Smith@example.com` are actually the same identity.

arpitjain099:
You're right per RFC 5321, and good catch - the local part has to stay case-sensitive. The PR as it stands now only case-folds the domain/host part and leaves the local part untouched, so `Smith@example.com` and `smith@example.com` are still treated as distinct identities. The only thing that now matches is a domain-case difference like `a@TnY.ToWn` against an `a@tny.town` SAN (the host is case-insensitive per DNS); URI and other SAN types stay exact. Does domain-only folding seem acceptable to you, or would you rather keep matching fully exact and instead steer users to normalize the case of their expected identity? Happy to go either way.

PR #1796, opened 2026-06-03
sign: produce valid CSR for non-ASCII identities

### LINKED ISSUE #1507: sigstore-python generates invalid CSRs when identity claim isn't ASCII
TL;DR: On GitHub Actions, the "identity" we embed into the CSR is the `sub` claim from a GitHub Actions OIDC token. That `sub` claim contains the job's environment, which can contain non-ASCII characters (like UTF-8 emojis).

Consequently, Fulcio throws a 400 with a malformed CSR when we submit.

This has two aspects to it:

* I'm a little surprised pyca/cryptography actually allows us to build this CSR -- if Fulcio is right and the email is encoded as an IA5String, then we might be generating invalid DER here.
* Separately, we need to see if we can encode this as something other than an IA5String, or maybe change the CSR subject to something other than an "email" (since that's incorrect anyways). 

The latter might actually be really simple: per https://github.com/sigstore/fulcio/issues/863 Fulcio doesn't actually use this field at all, so we can probably just put a stub value into it and carry on as normal.

See https://github.com/pypa/gh-action-pypi-publish/issues/376.

arpitjain099:
#### Summary

Fixes #1507.

On GitHub Actions the identity we put into the CSR is the OIDC token's `sub`
claim, and that claim can contain non-ASCII characters (for example when a job
environment name is non-ASCII). We embedded that value verbatim as a
`NameOID.EMAIL_ADDRESS` attribute, which X.509 encodes as an IA5String, and
IA5String is ASCII-only.

pyca/cryptography does not reject this, so we happily built a CSR whose
IA5String held non-ASCII bytes. That is malformed DER, and Fulcio's stricter
ASN.1 parser rejected the request with an HTTP 400. (This also answers the
"I'm a little surprised pyca/cryptography actually allows us to build this CSR"
note in the issue: it does, and the resulting DER is invalid.)

Fulcio does not actually use the CSR subject (see sigstore/fulcio#863): the
certificate's identity comes from the OIDC token, not this field. So the
minimal, low-risk fix is to only include the `EMAIL_ADDRESS` attribute when the
identity is ASCII-encodable, and otherwise leave the subject empty. I chose this
over forcing a UTF8String encoding because that would change the on-wire CSR for
every signer (every email attribute would flip from IA5String to UTF8String) to
preserve a field nobody reads. With this change the common ASCII path is
byte-for-byte unchanged, and the non-ASCII path produces a valid CSR.

While here I pulled the CSR construction out of `_signing_cert` into a small
`Signer._build_csr` helper so it can be exercised directly.

How to test: the new unit tests in `test/unit/test_sign.py` build a CSR from a
non-ASCII identity and assert it no longer smuggles non-ASCII bytes into an
IA5String subject attribute. On `main` these tests fail (the EMAIL_ADDRESS
attribute contains the non-ASCII value); with this change they pass. The
existing ASCII behaviour is covered too.

```
python -m pytest test/unit/test_sign.py
```

#### Release Note

Fixed generation of invalid CSRs when the identity is not ASCII (for example a
GitHub Actions `sub` claim with a non-ASCII environment name), which Fulcio
rejected with HTTP 400. The unused email subject attribute is now omitted when
the identity is not ASCII.

#### Documentation

No documentation changes. This is an internal CSR-encoding fix with no public
API or CLI surface change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* Fixed generation of invalid CSRs when the identity is not ASCII (for
+  example a GitHub Actions `sub` claim containing a non-ASCII environment
+  name). The identity was embedded as an ASCII-only IA5String email
+  attribute, producing malformed DER that Fulcio rejected with HTTP 400.
+  The unused subject attribute is now omitted when the identity is not ASCII
+  ([#1507](https://github.com/sigstore/sigstore-python/issues/1507))
+
 ## [4.3.0]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All versions prior to 0.9.0 are untracked.
   example a GitHub Actions `sub` claim containing a non-ASCII environment
   name). The identity was embedded as an ASCII-only IA5String email
   attribute, producing malformed DER that Fulcio rejected with HTTP 400.
-  The unused subject attribute is now omitted when the identity is not ASCII
+  The CSR subject is unused by Fulcio, so it is now omitted entirely
   ([#1507](https://github.com/sigstore/sigstore-python/issues/1507))
 
 ## [4.3.0]
--- a/sigstore/sign.py
+++ b/sigstore/sign.py
@@ -54,7 +54,6 @@ from datetime import datetime, timezone
 import cryptography.x509 as x509
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import ec
-from cryptography.x509.oid import NameOID
 from sigstore_models.common.v1 import HashOutput, MessageSignature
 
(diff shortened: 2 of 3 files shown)


woodruffw (dismissed):
I think we should probably just omit the CSR subject entirely, since we know it isn't being validated. I think that'd be a bit simpler than conditionally including it (and needing to test that).

Alternatively, if the CSR requires _something_ non-empty for the subject, I think we should probably use a consistent sentinel value (something like `sentinel@fake.example.com`).

arpitjain099:
Good call, that's simpler. I'll omit the CSR subject entirely since Fulcio doesn't validate it, and drop the conditional. Updating the branch now.

jku (changes requested):
This makes sense to me -- the CSR looks pretty unsual now but if Fulcio is fine with this then I don't think that's an issue.

I did unfortunately create a conflict in CHANGELOG by making a release -- could you update that?

jku (approved):

PR #1798, opened 2026-06-03
workflows: Fix uv call

jku:


woodruffw (dismissed):

.github/workflows/pin-requirements.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           working-directory: install
 
       - name: Install dependencies
-        run: uv install pip-tools
+        run: uv pip install pip-tools
```
woodruffw:
```suggestion
        run: uv tool install pip-tools
```

woodruffw:
Oops, didn't mean to self-improve that suggestion.

(I switch it to `uv tool install` since `uv pip install` will fail in this case, without a venv. `uv tool install` is the right way to install an isolated tool.)

jku:
Oh I see: I guess then we can also use uvx directly to avoid the install step

I think we managed to get to a state where neither one can approve... I'll force-push a fix

jku:
> uv tool install is the right way to install an isolated tool

I am not sure I actually agree with this for CI: there is no way to use pinned versions with uv tool AFAICT. Longer term I think we should aim to have all our build/dev tools listed in uv.lock, and should use those versions -- at least then we'd get a benefit from the busy work of maintaining a lock file

jku:
actually  `uvx pip-compile` does not work because that's  not the package name. Will fix

.github/workflows/pin-requirements.yml
```diff
@@ -86,7 +83,10 @@ jobs:
           cd install
 
           echo "sigstore==${SIGSTORE_RELEASE_VERSION}" > requirements.in
-          pip-compile --allow-unsafe --generate-hashes --upgrade --output-file=requirements.txt requirements.in
+          uvx --from pip-tools pip-compile \
```
woodruffw: Pedantic of me, but I think perhaps we should `uv add --dev pip-tools` and then `uv run --dev pip-compile`, i.e. lock `pip-tools` itself as one of our development dependencies.

(Sorry, I only just thought of this.)

jku: sure, I'm always for pinning everything exactly

jku:
Okay, one more try:
* use `uv run --only-dev pip-compile` in pin-requirements.yml
* Add updated requirements.txt (I ran the "uv run --only-dev pip-compile" command locally)
* Also use retries in the workflow: otherwise this will still fail after release

woodruffw (approved):

Issue #1809, opened 2026-06-21
Allow cryptography 49.x version


daeho-ro:
**Description**

New version is released and now this is also needed to bump.

PR #1810, opened 2026-06-21
Allow cryptography 49.x

daeho-ro:
#### Summary

Bump the ceiling on the cryptography dependency to allow 49.x and update the lockfile (cryptography 49.0.0, pyOpenSSL 26.3.0).

- https://github.com/sigstore/sigstore-python/issues/1809

#### Release Note

Bump the cryptography dependency to allow 49.x


woodruffw:
#1811 is what we want here, to avoid future pain as well. But thanks @daeho-ro!

PR #1811, opened 2026-06-21
Remove upper bound on cryptography

### LINKED ISSUE #1809: Allow cryptography 49.x version

woodruffw:
This upper bound is pretty disruptive to integrators/downstream packagers. I _think_ we originally added it for strict semver reasons, but:

1. cryptography ~~does not use semver~~ uses the one true versioning scheme (of making regular semantically breaking releases);
2. Our use of their APIs is not high-risk for breakage;
3. cryptography _itself_ integrates sigstore-python's tests as an integration signal, meaning it won't break us unexpectedly.

See https://github.com/Homebrew/homebrew-core/pull/287851 for context.

Closes #1809

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
   "Topic :: Security :: Cryptography",
 ]
 dependencies = [
-  "cryptography >= 42, < 49",
+  "cryptography >= 42",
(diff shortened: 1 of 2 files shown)


jku (approved):
💯

PR #1826, opened 2026-07-08
Raise VerificationError on malformed checkpoint text

arpitjain099:
#### Summary

While reading through the checkpoint parser I noticed that `LogCheckpoint.from_text` calls `int(lines[1])` for the log size and `base64.b64decode(lines[2])` for the root hash without guarding either one. If a bundle's inclusion-proof checkpoint has a log-size line that isn't an integer, or a root-hash line that isn't valid base64, those calls raise `ValueError` / `binascii.Error`.

This is reachable during verification: `verify_checkpoint` runs `SignedCheckpoint.from_text` on the (attacker-controlled) checkpoint envelope, and the verifier only expects `VerificationError` there, so a malformed bundle surfaces an unexpected exception type instead of a clean verification failure. The rest of the parser already reports malformed input as `VerificationError`, so this just makes those two conversions consistent. It's a robustness fix, not a verification bypass: the malformed input still fails, just with the right error type.

To reproduce before the fix: `LogCheckpoint.from_text("rekor.example - 123\nNOTANINT\nYWJjZA==\n")` raises a bare `ValueError`. I added a unit test covering both the non-integer size and the bad-base64 hash, and confirmed it fails on the old code and passes with the guard. The rest of the offline unit suite still passes.

#### Release Note

Fixed: a malformed checkpoint in a bundle now raises `VerificationError` during verification instead of a raw `ValueError` / `binascii.Error`.

#### Documentation

No documentation changes needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* A malformed checkpoint in a bundle's inclusion proof now raises
+  `VerificationError` during verification, instead of leaking a raw
+  `ValueError` or `binascii.Error`
+  ([#1826](https://github.com/sigstore/sigstore-python/pull/1826))
+
 ## [4.4.0]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


jku (approved):
Thanks, lgtm

Issue #1729, opened 2026-04-01
Signer should handle ExpiredCertificate?


jku:
I was reviewing sign.py and noticed that
* we cache Fulcio certificate by default
* we don't try to renew the certificate when it expires, just raise ExpiredCertificate

It would probably be harmless and good UX to renew the cert in Signer (if OIDC token is still alive). This would obviously only apply to some processes where hundreds of artifacts are being signed.

I'll run some tests before actually proposing anything but I'm filing this already so I don't forget

jku:
This might not be **that** useful since 
* dex tokens are short lived (shorter than fulcio cert)
* CI OIDC tokens are probably likely short lived

woodruffw:
Yeah, I think that's why we didn't do it originally 🙂 -- in practice the OIDC cred itself expires when/before the cached signing cert does, so the signer API would need to go through the (potentially interactive) OIDC retrieval flow again.

jku:
I just checked and I believe the GCP workload token has a 60 minute expiry
* so theoretically the fulcio cert expires earlier (10 mins)
* that's still a lot of signing time even with rekor2 so fixing this is not that important

The minor improvement this would make is that users who do massive signing operations would only need to handle ExpiredIdentity: ExpiredCertificate would be handled internally.

davidbrochart:
I'm not sure it's related, but I get an [error](https://github.com/y-crdt/pycrdt/actions/runs/30074211367/job/89424251692) trying to publish to PyPI using https://github.com/pypa/gh-action-pypi-publish:
```python-traceback
  Traceback (most recent call last):
    File "/root/.local/lib/python3.13/site-packages/pypi_attestations/_impl.py", line 207, in sign
      bundle = signer.sign_dsse(stmt)
    File "/root/.local/lib/python3.13/site-packages/sigstore/sign.py", line 208, in sign_dsse
      cert = self._signing_cert()
    File "/root/.local/lib/python3.13/site-packages/sigstore/sign.py", line 125, in _signing_cert
      raise ExpiredIdentity
  sigstore.oidc.ExpiredIdentity
  During handling of the above exception, another exception occurred:
  Traceback (most recent call last):
    File "/app/attestations.py", line 152, in <module>
      main()
      ~~~~^^
    File "/app/attestations.py", line 148, in main
      attest_dist(dist_path, attestation_path, signer)
[...]

PR #1838, opened 2026-07-24
Return cached cert first when signing

facutuesca:
From the discussion in https://github.com/sigstore/sigstore-python/issues/1729#issuecomment-5069835340

When signing, use the (valid) cached cert even if the identity token is already expired.

Should fix https://github.com/pypa/gh-action-pypi-publish/issues/415 once we make a release and update `pypi-attestations` with it.

### DIFF
--- a/test/unit/test_sign.py
+++ b/test/unit/test_sign.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import hashlib
 import secrets
+from datetime import datetime, timedelta, timezone
 
 import cryptography.x509 as x509
 import pretend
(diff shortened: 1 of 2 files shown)


test/unit/test_sign.py
```diff
@@ -72,6 +73,18 @@ def test_build_csr_non_ascii_identity_produces_valid_csr():
...
+def test_signer_uses_valid_cached_certificate_with_expired_identity():
+    cert = pretend.stub(
+        not_valid_after_utc=datetime.now(timezone.utc) + timedelta(minutes=5)
+    )
+
+    signer = Signer.__new__(Signer)
+    signer._identity_token = pretend.stub(in_validity_period=lambda: False)
+    signer._Signer__cached_signing_certificate = cert
+
+    assert signer._signing_cert() is cert
```
woodruffw: Minor: any way we can do this with less stubbing? It'd be nice to exercise a real cached flow here.

woodruffw (dismissed):
Thanks @facutuesca!

This seems reasonable to me, only one nitpick about the test.

test/unit/test_sign.py
facutuesca: done, see the new test

woodruffw (approved):

woodruffw:
Thank you @facutuesca!

Issue #1018, opened 2024-05-14
Custom hash for intoto payload


laurentsimon:
The current implement hardcodes a list of hashes https://github.com/sigstore/sigstore-python/blob/main/sigstore/dsse.py#L35.
Intoto allows any hashes, so I think we need to support use cases where users use other hashes.

Would removing the list of hardcoded values be enough?

I'm happy to work on this.

/cc @mihaimaruseac

woodruffw:
> Would removing the list of hardcoded values be enough?

My read of the [DigestSet spec](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md) is that it contains a lot of things that don't belong in modern cryptography, like GOST, SHA1, and MD5. I understand their rationale for doing this in the context of a generic file-matching tool, but exposing anything (roughly) weaker than SHA-2/256 in the context of digital signatures is IMO risky.

So I don't want to remove the list outright, but I _would_ be fine with adding to it, so long as the amendments are strong general-purpose cryptographic digests 🙂. Do you have specific additions that you want to make?

laurentsimon:
> > Would removing the list of hardcoded values be enough?
> 
> My read of the [DigestSet spec](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md) is that it contains a lot of things that don't belong in modern cryptography, like GOST, SHA1, and MD5. I understand their rationale for doing this in the context of a generic file-matching tool, but exposing anything (roughly) weaker than SHA-2/256 in the context of digital signatures is IMO risky.

I definitely understand that.

> So I don't want to remove the list outright, but I _would_ be fine with adding to it, so long as the amendments are strong general-purpose cryptographic digests 🙂. Do you have specific additions that you want to make?

Our use case is as follows: for model signing, we speed up hash calculation by splitting large files into multiple "shards" of X bytes each. Then we compute the hash on each shard in parallel; and aggregate them via `sha256(h1 || h2 || ...)`. So our hash is parameterized by the shard size. We're currently calling `sha256pv1$<shard-size>`.

What would be a good way forward to support this "family" of hashes?

woodruffw:
> What would be a good way forward to support this "family" of hashes?

[...]

PR #1846, opened 2026-07-28
Preserve the validation error when parsing a malformed statement

DrVelvetFog:
`Statement(contents=...)` discards the pydantic `ValidationError`, so callers get `malformed in-toto statement` with no indication of what was malformed — a rejected digest algorithm, a missing field and a bad `_type` are indistinguishable, and `__cause__` is `None`.

`StatementBuilder.build()` in the same module already surfaces it:

```python
except ValidationError as e:
    raise Error(f"invalid statement: {e}")
```

This makes the parsing path match.

Found while parsing a statement whose subject used `gitCommit`. That rejection is deliberate per #1018 and this doesn't change it — only what the caller is told about it.

The test asserts the cause is preserved and that the message names the offending field; it fails without the change:

```
AssertionError: assert None is not None
 +  where None = Error('malformed in-toto statement').__cause__
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+### Fixed
+
+* Parsing a malformed in-toto statement now includes the underlying validation
+  error, instead of discarding it. `StatementBuilder.build()` already did this;
+  `Statement(contents=...)` did not, so a rejected digest algorithm, a missing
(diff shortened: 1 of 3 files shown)


jku (approved):
lgtm, thanks

jku:
ci approval UI is not showing up: I'll close and reopen, maybe that makes github reconsider...

jku:
Thank you, sorry for the review delay

PR #1856, opened 2026-08-01
test: run conformance against staging

### LINKED ISSUE #942: conformance: support the staging tests
The conformance test GitHub action now supports testing against Sigstore staging infrastructure. Testing against staging is beneficial because when infrastructures changes, it happens first in staging: finding out ASAP is good for clients.

Enabling the support requires two things:
* conformance client `test/integration/sigstore-python-conformance` needs to support a new argument `--staging`: this can be done by replacing the script with the script from: https://github.com/sigstore/sigstore-conformance/blob/main/sigstore-python-conformance
* A second call to the conformance action (with input `environment: staging`) should be added to .`github/workflows/conformance.yml` as described in [conformance README](https://github.com/sigstore/sigstore-conformance?tab=readme-ov-file#usage)

ECD5A:
#### Summary

Run the Sigstore conformance suite against both production and staging infrastructure.

The existing conformance wrapper already handles the action's `--staging` argument, but the workflow only exercised the default production environment. A two-entry matrix now gives production and staging independent jobs, keeps both environments visible in required-check output, and disables fail-fast so one failure does not hide the other environment's result.

Closes #942.

Validation:

- parsed `.github/workflows/conformance.yml` with PyYAML
- `git diff --check`
- `python -m py_compile test/integration/sigstore-python-conformance`

The first pushed workflow run will provide the end-to-end production and staging validation.

#### Release Note

None. Test infrastructure only; there is no user-facing API or CLI change.

#### Documentation

No documentation update is required. The workflow follows the `environment: staging` usage documented by `sigstore/sigstore-conformance`.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -13,6 +13,13 @@ permissions: {}
 
 jobs:
   conformance:
+    name: Conformance (${{ matrix.environment }})
+    strategy:
+      fail-fast: false
+      matrix:
+        environment:
+          - production
+          - staging
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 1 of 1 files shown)


jku:
This seems harmless but also is likely not as useful as it sounds: 
* the `environment: staging` action option does not actually do very much anymore: https://github.com/sigstore/sigstore-conformance/issues/279 -- the ability to set trusted_root.json and signing_config.json turned out to be much more useful for the conformance test suite and the option is there mostly as a smoke test
* we already have smoke tests for staging (see e.g. staging-tests.yml) that should make noise if anything breaks

ECD5A:
Thanks for the context. Agreed — given the existing staging smoke tests and the limited semantics of the environment: staging option, this matrix would duplicate CI without adding meaningful coverage. I’ll close this PR. The 	rusted_root.json / signing_config.json path sounds like the useful direction if a concrete conformance gap emerges.

PR #1866, opened 2026-08-11
Remove pyasn1 dep

woodruffw:
We can now use `cryptography.hazmat.asn1`, thanks to the hard work of @facutuesca and @DarkaMaul 🙂 

This should be a non-breaking change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,10 +26,9 @@ classifiers = [
   "Topic :: Security :: Cryptography",
 ]
 dependencies = [
-  "cryptography >= 42",
+  "cryptography >= 47",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -134,3 +133,6 @@ ignore = [
   "UP011",
   "UP015",
 ]
+
+[tool.uv]
+exclude-newer = "P7D"
```
woodruffw, line 138: This is consistent with our Dependabot-side cooldown. I figure it's nice to have as part of `uv sync`, etc., too.

facutuesca (approved):
nice!

PR #1867, opened 2026-08-11
Remove pyOpenSSL entirely

woodruffw:
Removes `pyOpenSSL` in favor of cryptography's first-class X.509 validator.

This is stacked on #1866.

### DIFF
--- a/sigstore/verify/verifier.py
+++ b/sigstore/verify/verifier.py
@@ -27,7 +27,12 @@ import rekor_types
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import ec
-from cryptography.x509 import Certificate, ExtendedKeyUsage, KeyUsage
+from cryptography.x509 import (
+    Certificate,
+    ExtendedKeyUsage,
+    KeyUsage,
+    UnsupportedGeneralNameType,
+)
 from cryptography.x509.oid import ExtendedKeyUsageOID
 from cryptography.x509.verification import (
     Criticality,
@@ -270,7 +275,7 @@ class Verifier:
         try:
             # The verified chain includes the end-entity certificate, which callers omit.
             return verifier.verify(certificate, []).chain[1:]
-        except X509VerificationError as e:
+        except (X509VerificationError, UnsupportedGeneralNameType) as e:
             raise CertValidationError(
                 f"failed to build timestamp certificate chain: {e}"
             )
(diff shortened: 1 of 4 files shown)


sigstore/verify/verifier.py
```diff
@@ -244,35 +243,34 @@ def _establish_time(self, bundle: Bundle) -> list[TimestampVerificationResult]:
...
+        )
 
         try:
-            # get_verified_chain returns the full chain including the end-entity certificate
-            # and chain should contain only CA certificates
-            return store_ctx.get_verified_chain()[1:]
-        except X509StoreContextError as e:
+            # The verified chain includes the end-entity certificate, which callers omit.
+            return verifier.verify(certificate, []).chain[1:]
+        except X509VerificationError as e:
```
facutuesca: `ClientVerifier.verify()` can also raise `UnsupportedGeneralNameType` ([doc](https://cryptography.io/en/latest/x509/verification/#cryptography.x509.verification.ClientVerifier.verify)) (unless something before is validating it?)

sigstore/verify/verifier.py
```diff
@@ -244,35 +243,34 @@ def _establish_time(self, bundle: Bundle) -> list[TimestampVerificationResult]:
...
-        # down to a self-signed root.
-        store.set_flags(X509StoreFlags.X509_STRICT)
-        for parent_cert_ossl in self._fulcio_certificate_chain:
-            store.add_cert(parent_cert_ossl)
-
-        store.set_time(timestamp_result.time)
-
-        store_ctx = X509StoreContext(store, certificate)
+        # Client verifiers normally require the client-auth EKU. Fulcio certificates
+        # instead use code-signing, which is checked separately below; overriding
```
facutuesca, line 259: should we check that the certs have the code signing EKU?

facutuesca (commented):
LGTM, just a couple of comment/questions

sigstore/verify/verifier.py
woodruffw: I _think_ that happens below, in `_verify_common_signing_cert`.

sigstore/verify/verifier.py
woodruffw: Ah yeah, I guess we should catch that explicitly too. Thanks!

woodruffw: (I made this another naive failure mode, since we don't really expect this to happen in practice.)

facutuesca (approved):
LGTM!

PR #1870, opened 2026-08-12
verify: raise MetadataError instead of IndexError for trusted roots with no tlogs (#1822)

### LINKED ISSUE #1822: Verifier.__init__ raises unhandled IndexError when the trusted root has no transparency logs
If a `trusted_root` with no tlog entries is provided, the verfier crashes with a standard python error.

See:
https://github.com/sigstore/sigstore-python/blob/2c74d8f38701bcaee4613091efc5c5ea463b87af/sigstore/verify/verifier.py#L89-L91

Since the verifier documents its failures as VerificationError, a caller catching only VerificationError gets an uncaught crash. It's not security relevant both because the crash is blocking, and because it assume a likely malformed trusted_root (or is it supported to do a private sigstore deploy with no logs?).

The other verifiers (js/rs/go) seem to throw different failed verification errors for the same test case.

agu2347:
## Bug

`Verifier.__init__` raises an unhandled `IndexError` when given a `trusted_root` with no transparency logs, instead of a documented `sigstore.errors` exception.

Fixes #1822

## Context

Per @woodruffw's comment on the issue: this is intentional (a Sigstore instance is expected to have at least one tlog), but "we could turn that into a more structured error." `Verifier` documents its failures as `VerificationError` (or subclasses), so a caller catching only that gets an unhandled crash.

## Fix

`TrustedRoot` already has an established pattern for this exact situation -- `rekor_keyring()`, `ct_keyring()`, and `get_fulcio_certs()` all raise `MetadataError` when required data is missing from the trusted root (e.g. `"Did not find any Rekor keys in trusted root"`, `"Fulcio certificates not found in trusted root"`). This PR applies the same pattern to `Verifier.__init__`: check `trusted_root._inner.tlogs` before indexing into it, and raise `MetadataError("No transparency logs found in trusted root")` instead of letting the `IndexError` propagate.

## Testing

Added two unit tests to `test/unit/verify/test_verifier.py`, using `pretend.stub` for `trusted_root` to avoid any network dependency (matching the file's existing style for non-network tests):

- `test_verifier_init_no_transparency_logs` -- a `trusted_root` with an empty `tlogs` list raises `MetadataError`.
- `test_verifier_init_with_transparency_logs` -- a `trusted_root` with a `tlogs` entry still constructs successfully and picks its `base_url`, proving the new guard doesn't affect the happy path.

Verified: `pytest test/unit --skip-staging -m "not production"` passes 177/177 (175 existing + 2 new), 28 skipped (network-marked). Reverting just `sigstore/verify/verifier.py` reproduces the exact reported `IndexError` in the new negative test while the positive test still passes, confirming this is a genuine regression guard. `ruff format --check`, `ruff check`, and `mypy sigstore/verify/verifier.py` (the project's actual `make lint` gate) are all clean.

### DIFF
--- a/sigstore/verify/verifier.py
+++ b/sigstore/verify/verifier.py
@@ -59,7 +59,7 @@ from sigstore._internal.sct import (
 from sigstore._internal.timestamp import TimestampSource, TimestampVerificationResult
 from sigstore._internal.trust import KeyringPurpose
 from sigstore._utils import base64_encode_pem_cert, sha256_digest
-from sigstore.errors import CertValidationError, VerificationError
+from sigstore.errors import CertValidationError, MetadataError, VerificationError
 from sigstore.hashes import Hashed
 from sigstore.models import Bundle, ClientTrustConfig, TrustedRoot
 from sigstore.verify.policy import VerificationPolicy
@@ -92,6 +92,8 @@ class Verifier:
 
         # this is an ugly hack needed for verifying "detached" materials
         # In reality we should be choosing the rekor instance based on the logid
+        if not trusted_root._inner.tlogs:
+            raise MetadataError("No transparency logs found in trusted root")
         url = trusted_root._inner.tlogs[0].base_url
         self._rekor = RekorClient(url)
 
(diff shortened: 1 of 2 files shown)


jku:
There seem to be two PRs for this. I think I prefer the other one as this one does not quite fix the issue it documents:

> Verifier documents its failures as VerificationError (or subclasses), so a caller catching only that gets an unhandled crash.

With MetadataError this problem remains  (it's also not really correct: trusted_root.json is not TUF metadata)

PR #1872, opened 2026-08-13
Warn (and optionally fail) on a stale trust root with --offline

### LINKED ISSUE #1175: `--offline` should warn when the trust root is unreasonably old
`--offline` disables all possible online operations, including routine TUF trust root updates. This is desirable for offline environments where trust is pre-established or established asynchronously, but can result in users having _very_ stale trust roots without realizing it (meaning they may be missing revocations or important rotations that allow them to verify new signatures):

Fixing this has two parts:

1. Determine an "acceptable" oldness policy. Considering trust roots "stale" after 24h might be reasonable, but might also be too noisy for some offline users.
2. Emit a warning when performing `sigstore verify` commands in `--offline` mode with a trust root that's older than the window chosen in (1).

From (2), there's maybe a (3): pick a "red line" after which the warning becomes a hard error, forcing the user to update. Maybe 24h for a warning and 1 week for the red line, although again this could use user/integrator feedback.

CCing @mgorny @sethmlarson for thoughts!

mangrisano:
Closes #1175.

## Summary

`--offline` disables TUF trust root refresh, so a user can end up verifying against a very stale trust root without realizing it (missing key rotations or revocations). This adds a staleness check in offline mode.

## Behavior

When verifying with `--offline`, `TrustUpdater` inspects the cached TUF `timestamp.json` metadata and compares its `expires` against the current time:

- expired for longer than the **warn** threshold (default 24h) → logs a warning;
- expired for longer than the **error** threshold (default 7d) → raises `TUFError`, aborting verification.

If there is no cached `timestamp.json` (e.g. a fresh install running only off the embedded trust root, never refreshed), the check is skipped (debug log only).

### Why "past expiry"?

The TUF `timestamp` role's `expires` is the only signed temporal anchor available offline (there is no signed "issued at"), and it's exactly TUF's mechanism for bounding freshness. So staleness is measured as `now - timestamp.expires`: the thresholds mean "expired for longer than X", i.e. a grace period on top of the repository's own declared expiry.

## Configuration

Both thresholds are configurable, and either can be disabled with `off` (or `0`):

- CLI: `--offline-staleness-warn` / `--offline-staleness-error` (durations like
  `24h`, `7d`, `2w`)
- env: `SIGSTORE_OFFLINE_STALENESS_WARN` / `SIGSTORE_OFFLINE_STALENESS_ERROR`

They thread through `ClientTrustConfig.from_tuf/production/staging` down to `TrustUpdater`, so library users get the same defaults (and can customize via `ClientTrustConfig`).

## Notes / open questions

- Default thresholds (24h / 7d) follow the suggestion in #1175 — happy to tune.
- I left `Verifier.production/staging` untouched: the defaults already flow through them, so per-threshold customization there felt like speculative API surface. Glad to add it if you'd prefer.

## Tests

Added unit tests in `test/unit/internal/test_trust.py`: warn, hard error, fresh (no warning), missing timestamp (quiet), and error-level disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ All versions prior to 0.9.0 are untracked.
 
 ## [Unreleased]
 
+### Added
+
+* `sigstore verify` now warns (and can fail) when used with `--offline` and a
+  stale cached trust root. The thresholds are configurable via
+  `--offline-staleness-warn` / `--offline-staleness-error` (or the
+  `SIGSTORE_OFFLINE_STALENESS_WARN` / `SIGSTORE_OFFLINE_STALENESS_ERROR`
+  environment variables); pass `off` to disable either level.
+  ([#1175](https://github.com/sigstore/sigstore-python/issues/1175))
+
 ## [4.5.0]
 
 ### Fixed
--- a/docs/advanced/offline.md
+++ b/docs/advanced/offline.md
@@ -32,6 +32,33 @@ $ sigstore verify identity foo.txt \
     --cert-oidc-issuer 'https://github.com/login/oauth'
 ```
 
+## Detecting a stale trust root
+
+Because `--offline` skips the TUF update, the cached trust root can silently
+become stale. To surface this, `sigstore-python` inspects the cached TUF
+timestamp metadata and:
+
+* emits a **warning** once the trust root has been expired for longer than
+  24 hours;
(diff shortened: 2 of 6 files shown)


woodruffw:
Please don't engage with this repo in this way. If you have a design, discuss it with us first, do not send an automated PR.

PR #1875, opened 2026-08-16
ci: run conformance tests against staging

### LINKED ISSUE #942: conformance: support the staging tests

HaikalE:
Closes #942.

The current conformance client already supports the `--staging` argument. This completes the remaining workflow side of #942 by adding a second `sigstore-conformance` invocation with `environment: staging`, matching the usage documented by `sigstore-conformance`.

The staging run uses the same existing `xfail` configuration as the production run.

Validation:
- `git diff --check`
- workflow-only change; no application code changed

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -34,6 +34,13 @@ jobs:
             test_verify*managed-key-happy-path]
             test_verify*managed-key-and-trusted-root]" # see issues 1442, 1244
 
+      - uses: sigstore/sigstore-conformance@21533cde107c734ebc153c3e3a24d75fc9811a36 # v0.0.29
+        with:
+          entrypoint: ${{ github.workspace }}/test/integration/sigstore-python-conformance
+          environment: staging
+          xfail: "test_verify*intoto-with-custom-trust-root]
+            test_verify*managed-key-happy-path]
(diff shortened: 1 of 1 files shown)


jku:
I'm going to close this and the issue:

Change seems harmless but also is likely not as useful as it sounds: 
* the `environment: staging` action option does not actually do very much anymore: https://github.com/sigstore/sigstore-conformance/issues/279 -- the ability to set trusted_root.json and signing_config.json turned out to be much more useful for the conformance test suite and the option is there mostly as a smoke test
* we already have smoke tests for staging (see e.g. staging-tests.yml) that should make noise if anything breaks

PR #1880, opened 2026-08-21
Raise VerificationError when the trusted root has no tlogs

### LINKED ISSUE #1822: Verifier.__init__ raises unhandled IndexError when the trusted root has no transparency logs

theadsingh:
`Verifier.__init__` reads `trusted_root._inner.tlogs[0].base_url` to pick a Rekor
instance. If the trusted root carries no transparency log instances, that indexes an
empty list and the caller gets a bare `IndexError`, which is not part of the documented
surface:

```
url = trusted_root._inner.tlogs[0].base_url
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: list index out of range
```

This raises `VerificationError` instead, so a caller catching the documented exception
type sees it. Same shape as the checkpoint fix in #1826.

Worth noting the empty-tlog case is only reachable with the certificate authorities
still populated: `get_fulcio_certs()` runs one line earlier and already raises
`MetadataError` for a trusted root that is empty across the board. My first attempt at a
reproducer blanked everything and never reached the tlog lookup at all, so the test
keeps the CAs intact and asserts that before exercising the constructor.

I have not changed `TrustedRoot._verify()`, which felt like the other candidate spot.
Rejecting tlog-less roots at construction would also affect paths that never touch
Rekor, and I could not convince myself that was safe, so this stays scoped to the
verifier. Happy to move it if you would rather the validation lived there.

Closes #1822


jku (dismissed):
Normally I would say we should just validate the trusted root earlier... but in this case there is a counter argument: GitHub actually runs their instance *without a log*. Now I don't think sigstore-python knows what to do with an instance like that currently but it could in future... so I suppose your solution is preferable.

jku:
Would you mind resolving the changelog conflict? 

(I can also do it but then we need another maintainer to approve)

theadsingh:
Rebased on main and kept both entries in the Unreleased/Fixed list. Offline suite is green (177 passed).

jku (approved):

PR #1883, opened 2026-08-23
tests: verify current issuer certificates

### LINKED ISSUE #922: tests: Add test that verifies using current issuer certificates
During #910 development the test suite was passing but a simply running this failed with InvalidSCTKeyError:

```
sigstore sign README.md
sigstore verify identity --cert-identity $ID --cert-oidc-issuer $ISSUER README.md
```

The reasons for this seem to be:
* SCT verification code made assumptions about the order of certificates in the issuer "chain"
* At some point in time "chain" actually was in that assumed order
* all of the current test data was generated during that time
* current "issuer chain" looks different

So the problem is that right now we have no test that actually asserts that our SCT verification works on current sigstore trust root. We should  add
* A new asset `test/unit/assets/new_bundle.txt` (like the other txts, just a text file)
* A new good signature bundle in `test/unit/assets/new_bundle.txt.sigstore` (the bundle is the result of running `sigstore sign new_bundle.txt`
* a test that asserts that the new bundle verifies correctly (copying `test_verifier_offline()` probably works fine)

Note that if you create this new bundle, it will contain the email address that you used to authenticate with.

Aryadeepta:
#### Summary

Adds regression coverage for #922 using a bundle signed with the current
production Fulcio issuer certificates.

The existing test assets were generated against older issuer certificate
chains and did not ensure that SCT verification continued to work with the
current production trust root.

This adds a fresh production signing bundle and verifies it with
`Verifier.production(offline=True)`.

Closes #922

Tested with:

- `make lint`
- `make test`

#### Release Note

None. This is a test-only change.

#### Documentation

None.


jku:
Hi, thank you for the test. I'm going to close this as the situation has improved since the the issue was filed 2 years ago: at least the `test_fix_bundle_*` tests indirectly test this as part of "make test" (I'm not surprised if there are other tests too). Also, pyopenssl was completely removed since then so the original issue should now be impossible

Apologies for leaving the issue open.

PR #1892, opened 2026-09-08
build: Switch to uv_build

jku:
Switch from flit to uv_build as build backend to remove an additional dependency. This should also make the build actually reproducable (since flit was not strictly pinned)
* stop using dynamic version (not supported by uv): add a build time check and tweak version test instead to ensure versions match. I did not use `importlib.metadata.version("sigstore")` to avoid the import and call on every invocation
* Pin the uv_build version with the `version-file: uv.lock` trick for build reproducibility: otherwise uv-setup is not very strict about the uv version and will kind of use whatever is available
* Results are basically identical (trivial changes in build metadata)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,12 @@ doc: $(VENV)/pyvenv.cfg
 
 .PHONY: package
 package: $(VENV)/pyvenv.cfg
+	@SIGSTORE_VERSION=$$(. $(VENV_BIN)/activate && python -c "import sigstore; print(sigstore.__version__)") && \
+		PROJECT_VERSION=$$(uv version --short --frozen 2>/dev/null) && \
+		if [ "$$SIGSTORE_VERSION" != "$$PROJECT_VERSION" ]; then \
+			echo "Error: sigstore.__version__ ($$SIGSTORE_VERSION) does not match pyproject.toml version ($$PROJECT_VERSION)"; \
+			exit 1; \
+		fi
 	uv build
 
 .PHONY: release
--- a/Makefile
+++ b/Makefile
@@ -97,12 +97,6 @@ doc: $(VENV)/pyvenv.cfg
 
 .PHONY: package
 package: $(VENV)/pyvenv.cfg
-	@SIGSTORE_VERSION=$$(. $(VENV_BIN)/activate && python -c "import sigstore; print(sigstore.__version__)") && \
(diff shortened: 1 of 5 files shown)


jku:
> * Pin the uv_build version with the `version-file: uv.lock` trick for build reproducibility: otherwise uv-setup is not very strict about the uv version and will kind of use whatever is available

this is likely not that important in practice, I can remove it if there's complaints. I just think build dependency versions should be pinned as strictly as runtime ones.

pyproject.toml
```diff
@@ -71,6 +72,9 @@ dev = [
   "pip-tools>=7.6.0",
 ]
 
+# pin uv for setup-uv action
+uv = ["uv"]
```
woodruffw, line 76: I had never actually thought of using uv to pin itself like this before 🙂

Makefile
```diff
@@ -97,6 +97,12 @@ doc: $(VENV)/pyvenv.cfg
 
 .PHONY: package
 package: $(VENV)/pyvenv.cfg
+	@SIGSTORE_VERSION=$$(. $(VENV_BIN)/activate && python -c "import sigstore; print(sigstore.__version__)") && \
+		PROJECT_VERSION=$$(uv version --short --frozen 2>/dev/null) && \
+		if [ "$$SIGSTORE_VERSION" != "$$PROJECT_VERSION" ]; then \
+			echo "Error: sigstore.__version__ ($$SIGSTORE_VERSION) does not match pyproject.toml version ($$PROJECT_VERSION)"; \
+			exit 1; \
+		fi
```
woodruffw: My 0.02c would be for removing `__version__` now, technically it'd be breaking but IIUC modern Python prefers the `importlib.metadata` APIs for runtime version retrieval anyways.

woodruffw (dismissed):

Makefile
jku: I see.

I actually have a review comment of my own here -- this ends up doing `uv sync --dev` during package build, I think this unwanted.

So I think I'll remove this build time test -- there is still the unit test to make sure the version matches. I am not against removing `__version__` but that does feel breaking (I think there's a real chance someone is using it) so I'm not doing it here.

Resolving this, feel free to reopen if you feel more strongly about the version removal

jku:
Added a commit to remove the  build time version check: installing things during package build creates unnecessary supply chain risk. I still left `__version__` in the code for now for compatibility

woodruffw (approved):

PR #1900, opened 2026-09-13
verify: support RSA PKCS1v15 certificate keys

noah-ing:
#### Summary

Related to #1439. This implements the RSA PKCS#1 v1.5/SHA-256 portion, not the entire algorithm-support request.

Both artifact and DSSE verification currently invoke the signing key through an ECDSA-only operation. A valid RSA-signed bundle can pass certificate and transparency verification, then fail with `TypeError: RSAPublicKey.verify() missing 1 required positional argument: 'algorithm'`.

This change dispatches signature verification by public-key type and adds RSA PKCS#1 v1.5/SHA-256 for 2048-, 3072-, and 4096-bit keys. The existing ECDSA behavior and `Hashed` algorithm validation are preserved.

There is a separate certificate boundary to handle: the issuer's certificate-signature algorithm does not determine the subject key's artifact-signature algorithm. `_get_key_details()` now derives the RSA profile from the subject key instead. An unrestricted RSA subject can have an ECDSA- or PSS-signed certificate; a PSS-restricted subject SPKI must not be treated as an unrestricted PKCS#1 key. The verifier checks the certificate's subject SPKI and rejects that restricted profile before artifact or DSSE signature verification.

No signing defaults, dependencies, or public API signatures change. RSA-PSS artifact signatures and Ed25519/Ed25519ph verification are outside this patch. Verification never retries with another padding scheme or hash.

##### Regression coverage

- Real RSA signatures across the supported key sizes, with wrong-padding, wrong-hash, unsupported-size, substitution, and mislabeled-prehash negatives.
- Real certificates separating the subject key from ECDSA, RSA PKCS#1, and RSA-PSS issuer signatures.
- Artifact signature/body-consistency tests for Rekor v1/v2, and DSSE signature/body-consistency tests for Rekor v2. These component tests explicitly isolate common certificate/log-trust verification.
- Eighteen complete synthetic Rekor v1 bundle tests exercising certificate chain, SCT, signed entry timestamp, inclusion proof/checkpoint, identity policy, signature, and log-body consistency without mocked verification. They include correctly signed inconsistent log entries and tampered evidence.
- Five frozen-fixture tests for PSS-only subject SPKIs. They first establish that the underlying signatures are valid, then require the public verifier to reject the unsupported subject profile.

The complete-bundle test file produces six failures against parent `4963a123`, at the RSA/ECDSA verification mismatch, and passes on this branch. The synthetic fixtures do not claim production Fulcio enrollment or public Rekor publication. Full RSA v2 trust-path integration is not claimed by the component tests.

##### Validation

On commit `f340c513`:

- Full host and clean-checkout Linux arm64 Docker suites: **285 passed, 30 skipped, 2 xpassed** each.
- Focused RSA/DSSE/key-details suite with the minimum supported `cryptography==47.0.0`: **106 passed, 1 skipped**.
- Pinned upstream conformance `v0.0.29`, verification-only: **128 passed, 6 skipped, 6 xfailed**, using the existing expected-failure configuration.
- Ruff formatting/lint, mypy, Bandit, reference-documentation check, wheel/sdist builds, isolated installed-wheel RSA verification, and a Gitleaks scan of the committed diff passed.

No new skip or xfail was introduced. The full suite's skips and XPASS results come from existing tests. The complete host/integration and conformance runs used network access; a separate network-disabled Docker unit run passed with **257 passed, 37 skipped, 1 xpassed**.

From the checkout:

```sh
uv sync --locked --dev
make lint
make test TEST_ARGS='-q --tb=short'
make package
```

The focused end-to-end and subject-profile regressions can also be run directly:

```sh
uv run --locked pytest \
  test/unit/verify/test_verifier_rsa_bundle.py \
  test/unit/verify/test_verifier_rsa_spki.py
```

#### Release Note

Added RSA PKCS#1 v1.5/SHA-256 signing-certificate verification for artifacts and DSSE envelopes with 2048-, 3072-, and 4096-bit keys. RSA key details were derived from the subject key independently of the issuer's certificate-signature algorithm. RSA-PSS artifact/DSSE signatures and PSS-restricted subject keys remained unsupported; signing defaults were unchanged.

Included in `CHANGELOG.md`.

#### Documentation

The accepted profile and limitations are recorded in `CHANGELOG.md`. `test/assets/rsa-pss-spki/README.md` documents the public-only synthetic fixtures, the issuer/subject distinction, and replay versus regeneration requirements. Test-module docstrings distinguish isolated component coverage from complete synthetic trust-path verification.

There is no separate documentation-site PR in this change.


woodruffw:
This is a 1200 line diff with an LLM comment with no clearly motivated use case. Please don't do this.

